PR #114, opened 2016-03-18
Email hotfix

PSmolic:
@thomasplevy this is quick patch for email problem but we should revisit this thing soon. And I don't have environment set to test this anywhere. Any chance you can upload it to sendbox, I deleted email there so bug is reproducible there?


PSmolic:
I will reopen it on develop.

PR #143, opened 2016-05-20
String translations

PSmolic:
### Task:
https://trello.com/c/wYL9nSuE/442-missing-translations


includes/class.llms.frontend.assets.php
```diff
@@ -61,7 +61,10 @@ public function enqueue_scripts() {
 		wp_enqueue_script( 'collapse', plugins_url( 'assets/js/vendor/collapse.js', LLMS_PLUGIN_FILE ) );
 		wp_enqueue_script( 'transition', plugins_url( 'assets/js/vendor/transition.js', LLMS_PLUGIN_FILE ) );
 
-		wp_enqueue_script( 'llms', plugins_url( '/assets/js/llms' . LLMS_Frontend_Assets::$min . '.js', LLMS_PLUGIN_FILE ), array( 'jquery' ), '', true );
+		wp_register_script( 'llms', plugins_url( '/assets/js/llms' . LLMS_Frontend_Assets::$min . '.js', LLMS_PLUGIN_FILE ), array( 'jquery' ), '', true );
+		// TODO: move this translatable string to some better place.
+		wp_localize_script( 'llms', 'quiz_translations', array( 'enter_answer' => __( 'You must enter an answer to continue.' ) ) );
```
PSmolic: This string should be moved to better place but since it is the first one of this kind I am not sure where to place it.

templates/quiz/results.php
```diff
@@ -68,9 +68,9 @@
 						<h5 class="llms-content-block">
 							<?php
 							if ( $is_passing_score ) {
-								_e( apply_filters( 'lifterlms_quiz_passed','Passed' ), 'lifterlms' );
```
PSmolic: I think that this wasn't working because way it was placed in filter.. This should work but I don't know how to test it.

PR #270, opened 2016-12-20
Fix for compatibility with OptimizePress live editor.

tpkemme:
Adds a function in class.llms.placenice.php that loads the necessary frontend assets to make lifterlms work with the OptimizePress live editor.


thomasplevy:
@tpkemme looks good, will merge in after CI checks have passed

PR #283, opened 2017-01-10
Apply llms_page_restricted filter to all content restriction cases

matthalliday:
This will let developers modify existing content restrictions (i.e.
allowing admin users to view course content outside of the course time
period).


thomasplevy:
@matthalliday yea good call, building and testing a release this AM will try to pull this into it if time allows, otherwise will hit 3.5.6

thank you!

matthalliday:
Awesome! 👍

PR #285, opened 2017-01-14
Adds unique class to each student dashboard LI

hallshouse:
This change uses jQuery to add a unique class to each List Item in the student dashboard so each one can be styled individually.

Examples:
/* Remove "Redeem a Voucher" list item */
.llms-sd-item4 {
    display: none;
}

/* Remove "Order History" list item */
.llms-sd-item5 {
    display: none;
}


thomasplevy:
@hallshouse this can be accomplished already:

```
.lms-sd-nav .llms-sd-item:nth-child(4) { 
     display: none;
}
```

Also, if it is absolutely necessary to add these classes, it could (maybe should) be done via PHP rather than an inline js.

hallshouse:
Thanks Thomas:
A better solution indeed... Didn't know you could do :nth-child(#)
The final CSS solution:
.llms-sd-item:nth-child(4) {
    display: none;
}
need to remove the .lms-sd-nav from the path.

Best Regards

PR #361, opened 2017-07-24
Code Climate compliance

hallshouse:


hallshouse:
Thomas:
I will not be fixing the 3 remaining issues.
Even if I manually change them directly in GitHub, they never get resoled.

thomasplevy:
@hallshouse I think this may help but I can fix a few line endings if that's the only issue: https://help.github.com/articles/dealing-with-line-endings/#platform-windows

thomasplevy:
@hallshouse could you explain why you've disabled the default date filters and replaced them with what -- as far as I can tell -- are more or less the default date filters, just rewritten? I'm trying to figure out why you've decided to do this and I'm assuming there's a good reason for it but I just can't figure out what it is and I don't feel great about merging code into production that I don't fully understand.

Thanks,

hallshouse:
The reason I had disabled the date filters, then later rewrote them was to place the date options to the right of the course, lesson and quiz filters.

PR https://github.com/gocodebox/lifterlms/pull/366 has the latest code.
FYI: Even in vagrant, I still could not fix that pesky \r\n issue.

PR #400, opened 2018-01-03
Add link to error message to continue user journey.

andreasblumberg:
The additional link provides the ability to continue without beeing lost while he already subscribed to a plan, although he doesnt know or remembers. This way the user gets the ability to continue to the course or even go back and chose a different course as he now realised he picked the wrong one. A redirect in order to push the user towards the course would be also possible but does not provide the chance to let the user decide what he really wants. @thomasplevy Looking forward to your thoughts about this request.

### DIFF
--- a/includes/controllers/class.llms.controller.orders.php
+++ b/includes/controllers/class.llms.controller.orders.php
@@ -249,7 +249,7 @@ class LLMS_Controller_Orders {
 		} // make sure the user isn't already enrolled in the course or membership
 		elseif ( llms_is_user_enrolled( $person_id, $product->get( 'id' ) ) ) {
 
-			return llms_add_notice( __( 'You already have access to this product!', 'lifterlms' ), 'error' );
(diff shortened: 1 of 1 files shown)


thomasplevy:
(Line 250)[https://github.com/andreasblumberg/lifterlms/blob/ce0ce90883afcf76476fc06ee9631b22ad4f75de/includes/controllers/class.llms.controller.orders.php#L250] runs the check where I think this update makes the most sense

The lack of ID on the order -- the message this PR updates -- would occur if someone *manually* updated the order-key posted in a hidden field and the order itself couldn't be located during a failed payment retry or if for some reason the creation of a new pending order failed (a database error, or something, for example). In both of these scenarios I do think the next step is to try again. 

If you're seeing this particular error pop *because* the student already has access to the course I think we have an issue somewhere above that's preventing the enrollment check on 250 from working properly. Do you have a test case you can setup / provide me with that guarantees this particular error occurs when the student already has access to the product they're trying to process?

andreasblumberg:
Thanks for your feedback. my fault, I was within the wrong line. 250 is the one i should have edited with the message and link instead. that for i will fix it with the next commit.

thomasplevy:
Great! Seems like a good improvement as you've noted just want to make sure it's in the right place!

PR #406, opened 2018-01-16
fix for is_search called incorrectly

Mte90:
I get a lot of:
```Notice: is_search was called incorrectly. Conditional query tags do not work before the query is run. Before then, they always return false. Please see Debugging in WordPress for more information. (This message was added in version 3.1.0.)```
Seems that happen on different url and after a little bit of debugging I found that happen on lifter.
In a day I can have 3mb of logs stuff, with this fix the problem doesn't happen anymore.

### DIFF
--- a/includes/class.llms.query.php
+++ b/includes/class.llms.query.php
@@ -154,10 +154,10 @@ class LLMS_Query {
 	 * @version  3.6.0
 	 */
 	public function pre_get_posts( $query ) {
-
+		global $wp_query;
 		$modify_tax_query = false;
 
-		if ( is_search() ) {
+		if ( isset( $wp_query ) && is_search() ) {
 			$modify_tax_query = true;
 		}
 


Mte90:
any hints for this?

thomasplevy:
@Mte90 I hadn't merged this in yet as I wasn't entirely certain this was the "correct" solution as I'm not certain why pre_get_posts would ever be firing before the query is set... The `$query` is the WP_Query that's passed in... However, after looking closer I think that this `is_seach()` check is possibly causing some other issues I haven't been able to track down because it is setup to affect queries other than the main query (which is the only query this should be affecting...)

I've setup a few more tests and run some code against this alternate solution.

Can you do me a favor and run that to see if it fixes your log issues in addition to fixing these other issues I think may be related: https://github.com/gocodebox/lifterlms/commit/587df79c1ce9da12dcd35b26132984b484ea5f77

Thanks for your patience. I've had a lot on my plate the last few months and I apologize for taking so long to look at this PR.

Mte90:
I patched the code and I don't see any errors but looking at the new code the effects of my version or your are the same.
Only that your is better :-D

thomasplevy:
Not trying to play who's better, I saw some other issues that my patch fixed as well but wasn't entirely sure how to recreate the issues you're identifying so I thought mine fixed that but wasn't certain.

Thanks again, I'll have this merged into the next update more than likely in the next few hours

Best,

Issue #417, opened 2018-01-19
Blank Markup


sujaypawar:
In price box, it seems there is blank markup. It can be handled differently.
Here is the screenshot - http://screenshots.sharkz.in/sujay/180119163907.png

thomasplevy:
Yep, good call. I'll clean that up in a future release.

thomasplevy:
potential PR in #422

PR #422, opened 2018-01-23
Removed empty html markup.

dcmayankavi:
Issue: https://goo.gl/2T7ubm


thomasplevy:
@dcmayankavi i realize empty markup is pending as #417 but the whitespace created by the empty featured element is allowing access plans to align visually when one is featured and the others are not. I'm sure there's a better solution than what's currently in core but moving the if statement as this PR does will create a new bug!

test_some_engagements_-_lifterlms_demo_site

thomasplevy:
now that i'm thinking about this the trial empty markup may be serving the exact same purpose too... i need to do some more testing on it to confirm

dcmayankavi:
@thomasplevy, Instead of adding empty markup you can add a class for `non-featured` element which will add top margin, or you can add the top margin as per other's height using JS. I'm not sure but I think instead of using empty markup you can use other solutions as for that.

thomasplevy:
those sound like good solutions, want to update the PR or should I close this?

thomasplevy:
Closing and will track on #417

PR #428, opened 2018-02-02
Progress data attribute added.

dineshchouhan:
We are implementing a progress bar style like below screenshot, we required a small addition to your current markup, you can check in PR difference. It'll not affect your existing style.

Screenshot: https://goo.gl/K4qDoL


thomasplevy:
Will be in 3.16.7 released more than likely in the next few days.

Thanks @dineshchouhan

PR #459, opened 2018-02-23
Filter Added to get_audio & get_video methods of Lesson.

dineshchouhan:


thomasplevy:
@dineshchouhan thanks for this. I updated a bit to ensure the empty return also gets filtered, see #460

Issue #393, opened 2017-11-29
student->get_progress() is inefficent


thomasplevy:
should cache progress on usermeta
completions should trigger async generator to recalc progress

Mte90:
I made a draft for that https://github.com/gocodebox/lifterlms/pull/466
With this code do -70 query with the demo data, I need only help to reset that value when the user update the status.

thomasplevy:
see #474

PR #466, opened 2018-03-01
Working on 393

Mte90:
As explained in #393 the method `get_progress` with the demo data made like 70 query everytime for this value.
70 query!
This commit cache and store this value for the specific user and based on the type in a usermeta.
Now the page do less query!!
The issue actually for me is to cleanup this value when the user do a lesson etc. I am not an expert of the code but I guess that @thomasplevy will have some idea how to implement that part.

### DIFF
--- a/includes/models/model.llms.student.php
+++ b/includes/models/model.llms.student.php
@@ -980,7 +980,7 @@ class LLMS_Student extends LLMS_Abstract_User_Data {
 		$total = 0;
 		$completed = 0;
 		
-		$saved_complete = get_user_meta( $this->get_id(), '_status_' . $type . '_' . $object_id );
+		$saved_complete = get_user_meta( $this->get_id(), '_status_' . $type . '_' . $object_id, true );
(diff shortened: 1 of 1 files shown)


thomasplevy:
I think there's a better solution than caching on the usermeta table -- the completed status is already available on the lifterlms_user_postmeta table (see how `LLMS_Student->insert_completion_postmeta()` works)

This will mean that additional code doesn't need to be written in order to clear cached data when the progress changes

However... in general this solution will only solve performance issues when checking if the object is completed -- we really want to solve the issues in `LLMS_Student->get_progress()`

Mte90:
Probably yes a better solution there is to avoid saving in the usermeta, but looking at the amount of query is better to save that value somewhere. If there is a specific table is not a problem.
But instead do that check with all this query everytime check if this value already exist is the most simple way. It's better to avoid transients for this one.

Mte90:
I confirm locally that this branch save 70 query to the db.

thomasplevy:
See #476

PR #467, opened 2018-03-01
Optimization experiments

Mte90:
The idea is to look with query monitor in a page and see the same query repeat more than once.
I tried with caching in a var inside the class and the code seems work but the call keep to stay so I am wondering if this class is called without existing as global object to cache this values.
If this approach that for every query there is a cache will be better for the system, but I have no so much knowledge of the plugin to understand where is the issue.
In any case improve the performance looking at the code seems very easy.
I hope that this code like for #466 can help to speed and implement this improvements.

### DIFF
--- a/includes/models/model.llms.student.php
+++ b/includes/models/model.llms.student.php
@@ -10,6 +10,10 @@ if ( ! defined( 'ABSPATH' ) ) { exit; }
  * @version 3.16.0
  */
 class LLMS_Student extends LLMS_Abstract_User_Data {
+	
+	private $enrollment_date = array();
+	private $enrollment_status = '';
+	private $is_complete = array();
 
 	/**
 	 * Retrieve an instance of the LLMS_Instructor model for the current user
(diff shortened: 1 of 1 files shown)


thomasplevy:
I think a better way to achieve an optimized class would be to utilize [WP_Object_Cache](https://codex.wordpress.org/Class_Reference/WP_Object_Cache)

This PR won't address multiple queries to the same things made through by different instances of the the class but utilizing the object cache would.

It's possible to do something along these lines if we optimized a single global instance of the `LLMS_Student` but in many circumstances multiple students are instantiated on the same page load (especially on the admin panel or when logged in as an instructor/admin) so I think it'd be better to save frequently used values to the object cache

I'm working on moving queries to the postmeta table over to the `LLMS_LLMS_Query_User_Postmeta` class -- if all direct $wpdb queries are added to this class then we could implement really simple object caching over there. Alternatively getters in the student class could implement their own cache getter/setters methods

I'm not saying this solution wouldn't work but I think there's a better solution than this... Setting things up with this pattern means there's a ton of other private class vairables that could be potentially added to the class and it feels like it'll start to get really messy really fast...

I know this isn't perfect today but I don't want to implement solutions that I think will be hard to maintain moving forward and I see a few issues with this possible solution...

Mte90:
I agree with you, if you need to experiment this things I am here :-)

Issue #453, opened 2018-02-20
Make "Edit Quiz" tooltip say "Add Quiz" if a quiz doesn't exist yet


thomasplevy:
https://www.evernote.com/l/AngaKX6CKoRI5rxkoniOWRBMr2K9_nLih-o

PR #482, opened 2018-03-27
Version 3.17.0 - assignments prep

thomasplevy:


thomasplevy:
Among many other things fixes #453

PR #483, opened 2018-03-28
Add RTL support

actual-saurabh:
Summary of Changes:

1. Add automated RTL version generation using gulp.
1. Refactor stylesheet processing in gulp to arrange them based on the task, instead of on the functional type of stylesheet to accomodate RTL generation
1. Recognise and use `SCRIPT_DEBUG` when enqueuing stylesheets for internal and vendor stylesheets.


actual-saurabh:
The issues that code climate generates are unavoidable as of now. The code is similar because it needs to be similar:

screen shot 2018-03-28 at 2 03 57 pm

PR #515, opened 2018-04-28
address issue 511

### LINKED ISSUE #511: Code improvement on BuddyPress detection
immagine

As I can see in every page there is this query, but there isn't Buddypress in my installation.

immagine

AS I can see on https://github.com/gocodebox/lifterlms/blob/d824132daa0dd49347940a3b7c484c1e1f88960f/includes/class.llms.integrations.php#L52 
the class are defined maybe they are executed on every page.

There is the request https://github.com/gocodebox/lifterlms/blob/master/includes/abstracts/llms.abstract.integration.php#L125 maybe we can add a check in the classes with detect the plugin like for https://github.com/gocodebox/lifterlms/blob/0eb956847a57893aeca29794148c111462c4c565/includes/class.llms.widgets.php#L34 inside every class.
In the meantime we can disable with an hook and save a query in every page, in all the pages of the site.

thomasplevy:
fixes #511

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 LifterLMS Changelog
 ===================
 
++ Reduce unnecessary DB queries for integrations by checking for dependencies and then calling querying the options table to see if the integration has been enabled.
++ Fix error occurring when all 3rd party integrations are disabled by filter, credit to [@Mte90](https://github.com/Mte90)!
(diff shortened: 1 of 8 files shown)


Mte90:
You are right with this code is not required to use the filter with my patch :-D
The query is not executed.

thomasplevy:
I'll merge your patch in too, don't want to have errors for any reason

Mte90:
yes better avoid errors :-D

PR #554, opened 2018-05-22
Coupon expiration fix

### LINKED ISSUE #552: Expiry occurs 24 hours earlier than UX suggests
https://github.com/gocodebox/lifterlms/blob/0b12fe452d85b3a85714f568c4961c565f2f1b3b/includes/models/model.llms.coupon.php#L219 checks if the current time is lesser than the saved expiry. However, the comparison is made by converting the expiry date (for example, 05/22/2018) into seconds which would turn out to be the timestamp for 00:00 Hrs on that day.

So, anyone trying to use the coupon **on** the date gets an error message that is slightly confusing because the date hasn't passed yet:

screen shot 2018-05-22 at 6 48 12 pm

The UX also suggests that the coupon should still work on the expiry date:

screen shot 2018-05-22 at 6 46 26 pm

Adding a day (24 hours) to expiry before this check would fix the issue.

thomasplevy:
Resolves #552

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ LifterLMS Changelog
 v3.18.? - 2018-05-??
 --------------------
 
++ Fix coupon issue causing coupons to expire at the beginning of the day on the expiration date instead of at the end of the day
 + Improved integrations settings screen to allow each integration to have it's own settings tab (page) with only its own settings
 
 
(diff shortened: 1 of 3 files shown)


thomasplevy:
@actual-saurabh check this for me. I abstracted the expiration date into a new function and added some new tests.

The main change here is that adding a full 24 hours not only broke the unit test but also caused what would be a new issues (although probably unlikely to ever be encountered):

Set expiration date to 2018-05-22, expected expiration *should* be 2018-05-22 23:59:59
Adding 24 hours sets the expiration at 2018-05-23 00:00:00
So the fix should  be add 23 hours 59 minutes and 59 seconds (or `DAY_IN_SECDONDS - 1`)
This gets us to our expected expiration date and fixes the bug

I also cleanud up some various code formatting issues from way back when in this PR

Issue #558, opened 2018-05-25
Style Cascade Order is incorrect in certificate export


actual-saurabh:
This part of the certificate export removes all link tags and converts them into style tags that are appended at the end. 

https://github.com/gocodebox/lifterlms/blob/d04671439f66f15668ff71cbd706dcfc35d79673/includes/class.llms.certificates.php#L178-L213

It doesn't accomodate the particular order in which stylesheets added through the link tag and through inline style tags are loaded for the actual results. So, a lot of custom CSS becomes ineffective and the export looks different from how the certificate appears on the website.

Instead if the link tag is replaced exactly in the same position using http://php.net/manual/en/domnode.replacechild.php, the order would be maintained.

thomasplevy:
Can you provide a specific example of a link losing it's initial order during the loop? The getElementsByTagName seems to be loading them in their initial order and then during the loop they're all appended in the same order so I don't see the order changing here (I did a quick comparison).

I wouldn't care but see the my note on #559 which has me questioning this issue now

actual-saurabh:
I missed the removal of non `rel="stylesheet"` link tags.

So, the line that removes all the link tags should still exist (I'll update the PR accordingly)

What is causing an issue is that the order of inline `style` tags. If inline css is added after a stylesheet is loaded:

```
<link rel="stylesheet" href="some_stylesheet_url.css">
<style>
/* contains definitions that override styles in `some_stylesheet_url.css`
</style>
<link rel="stylesheet" href="another_stylesheet_url.css">
```

After processing, this becomes *equivalent* to:
```
<style>
/* contains definitions that override styles in `some_stylesheet_url.css`
</style>
<link rel="stylesheet" href="some_stylesheet_url.css">
<link rel="stylesheet" href="another_stylesheet_url.css">
```

because all the link tags are replaced and appended at the end, breaking the order between inline styles and link tags.

thomasplevy:
I figured out what I was seeing...

After some debugging I found that on my test site there would be 27 `<links>` to be replaced or removed but only the first 21 links would actually be updated resulting in missing styles. Mysterious.

I found this comment in the php docs: I found the following on PHP.net: http://php.net/manual/en/class.domnodelist.php#83178

The original code did not modify the dom with relation to the list of elements found in the `$links` variable but by using replace child links would be removed during iteration. This caused some links to get missed completely (and therefore they weren't getting inlined). After your update to remove other links it also meant they would just completely dissappear.

[...]

PR #559, opened 2018-05-25
Fix certificate style cascade order, #558

actual-saurabh:


thomasplevy:
@actual-saurabh I'm seeing some strange issues with this... At a glance it all looks good but I noticed you removed the part that removes all the links after converting the CSS links to inline styles.

I wanted to double check and I think we still want to remove all non CSS link's. This includes prefetches, xml feed links, manifest links, and more... There's not much harm in having these but I'd like to clean up the DOM as much as possible during this export.

I went to add this back in and I'm noticing that certain links are not properly converted when using the replaceChild. I'm having a bugger of a time figuring out why that's happening but consistently the certificates stylesheet from the LifterLMS core is not converted to an inline style. All the other links aside can you check if this is occurring on your site too?

certificate-w00t-pn7mlyyjyw1a_html

Also I'm curious why running a foreach and appending is changing the cascade order. I don't see that happening after doing a little comparison of the initial order compared to the exported order. Are you seeing a specific items that's losing it's initial order during the export?

actual-saurabh:
See https://github.com/gocodebox/lifterlms/issues/558#issuecomment-393030405

PR #589, opened 2018-06-15
Feature/bulk enroll

### LINKED ISSUE #562: Bulk Enrollment for Courses & Memberships
- [x] Using `manage_users_extra_tablenav` action here: https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/class-wp-users-list-table.php#L300, it'd be easy to add a `select2` box for memberships and courses and have a simple UI for bulk enrollment. The choice would be limited to only one course or membership at a time.

- [x] Set a custom administrative *action* called `'enroll'` (https://github.com/WordPress/WordPress/blob/master/wp-admin/includes/class-wp-list-table.php#L478-L499 & https://github.com/WordPress/WordPress/blob/master/wp-admin/users.php#L99

- [x] Filter the `default` case of the switch statement here: https://github.com/WordPress/WordPress/blob/master/wp-admin/users.php#L444 to get the array of `user_ids` and the rest of the `$_POST` data and then loop through each user and run `llms_enroll_student`https://github.com/gocodebox/lifterlms/blob/master/includes/functions/llms.functions.person.php#L156-L171 to complete enrollment

actual-saurabh:
Closes #562


thomasplevy:
See #600

PR #602, opened 2018-06-29
Fix/account forms

actual-saurabh:
Fixed usage for llms_verify_nonce(), and added the same to the account update form


thomasplevy:
@actual-saurabh something weird happened as a result of my hasty merge & revert... your fork is showing folder structure changes from #602 but this PR does not pull those additional updates and changes in...

I'm trying to figure out how to get all the changes from 602 & these changes into master but it's proving difficult...

Sorry to be difficult here but short of rewriting all the code via copy and paste I can't figure out how to get this merged in without losing a bunch of your work...

PR #606, opened 2018-07-03
Fix forward slash handling especially for LaTex

### LINKED ISSUE #596: Backslashes in post content disappear on cloning
During json encoding, each backslash (`\`) converts into `\\` which during json decode here disappears:  https://github.com/gocodebox/lifterlms/blob/master/includes/class.llms.generator.php#L82. A detailed explanation can be found here: https://stackoverflow.com/a/32057601/1589999

So, we'd have to replace the code ain Line 82 above with the following to preserve backslashes:
```
    $raw = json_decode( str_replace( '\\', '\\\\', $raw ), true );
```

actual-saurabh:
Fixes #596


thomasplevy:
@actual-saurabh going to hold on this until you add the builder-related updates we spoke about this AM

actual-saurabh:
I've made those changes as per https://github.com/gocodebox/lifterlms/issues/596#issuecomment-403037022 in https://github.com/gocodebox/lifterlms/pull/606/commits/f50103f2750e1cc8b70f796fc0e83b028ddf63a1 and this is ready for review/ merging.

thomasplevy:
@actual-saurabh after merging this in I saw 2 new issues:

1) course exports are not importing directly. I spent about 20 minutes debugging and I can't tell if it's related to this or something pre-existing. I'm exhausted, I can't figure it out and it's not an emergency so I'm terrified of pushing a new issue before I take a long weekend.

2) With slashes the data in the builder never registers as having  been properly saved. I'm sure this has something to do with the data comparisons on the JS side in the builder's sync controller. Again, just too tired to debug.

Please have a look and if you can't recreate these I'll look at it next week when I get back from vacation

actual-saurabh:
@thomasplevy I'll try and debug this today and/or tomorrow.

actual-saurabh:
I didn't get time to look into this yet. I'll give this a go again later today and tomorrow.

PR #655, opened 2018-10-01
Fix phpcs errors

marciojc:
Fix 7 phpcs errors.


thomasplevy:
These are fixes automated via build tools. Thanks for your contribution and I'd like to encourage you to pickup a new issue from our open issues instead of manually updating PHPCS issues.

Thanks!

Issue #644, opened 2018-09-28
Course titles with quotes & apostrophes are converted to HTML entities in email subject lines


thomasplevy:
### 1. Reproduction Steps

+ Create a course with a title containing quotes or apostrophes, eg
+ The Official 'Quickstart' "Course" for LifterLMS
+ Purchase the course
+ Email subject line (default) will be: `Purchase Receipt for The Official &#8216;Quickstart&#8217; &#8220;Course&#8221; for LifterLMS`

### 2. Expected Behavior

Email subject line should be `The Official 'Quickstart' "Course" for LifterLMS`

PR #658, opened 2018-10-02
decode html entities for email titles

philwp:
fix for #644 using `wp_specialchars_decode`


thomasplevy:
thanks @philwp, I'll take a closer look later in the week but this one should be pretty solid

PR #678, opened 2018-10-06
Fix backslash escaping in builder

### LINKED ISSUE #677: Builder removes backslashes from inputs
### 1. Reproduction Steps

Add a `\`(backslash) in any element. Save. Exit Builder. Come back to builder.

### 2. Expected Behavior

Backslash will be saved and will stay as it is.

### 3. Actual Behavior

Backslashes disappear

### 4. Error Messages / Logs

NA

### 5. System Report

NA

### 6. Browser, Device, and Operating System

NA

actual-saurabh:
## Description

Fixed json_decode in builder to escape and save backslashes correctly. Fixes #677 but not #596.

## How has this been tested?

Added the following string to Quiz Description, question and options:

`C:\\Program Files\Test\`

Before the fix, it would turn into `C:\Program FilesTest` once the builder is exited and returned to or the quiz is previewed. After the fix the string stays as it is.

I also tested cloning and export/import and with both of them, this fix doesn't work because the code doesn't come into play at all. However, unlike a previous attempt, it doesn't break cloning or export/import. (The string continues to turn into `C:\Program FilesTest` on cloning and export/import. )Fixing that is more complicated that what's done here but at least users will be able to start using LaTex and backslashes in course and course unit content and part of the problem will get resolved.

## Types of changes
Bug fix for #677 

## Checklist:

### DIFF
--- a/includes/admin/class.llms.admin.builder.php
+++ b/includes/admin/class.llms.admin.builder.php
@@ -359,7 +359,8 @@ if ( ! empty( $active_post_lock ) ) {
 		}
 
 		// only mess with our data
-		$data = json_decode( $data['llms_builder'], true );
+		$data = json_decode( str_replace( '\\', '\\\\', $data['llms_builder'] ), true );
+		
 
 		// setup our return
 		$ret = array(
--- a/includes/admin/class.llms.admin.builder.php
+++ b/includes/admin/class.llms.admin.builder.php
@@ -360,7 +360,6 @@ if ( ! empty( $active_post_lock ) ) {
 
 		// only mess with our data
 		$data = json_decode( str_replace( '\\', '\\\\', $data['llms_builder'] ), true );
-		
 
 		// setup our return
 		$ret = array(


thomasplevy:
Builder does not recognize updates as successful saves due to escaped data being returned in the save response. In @actual-saurabh's example the escaped `C:\\\\Program Files\\Test\\` will be returned and compared against the unescaped `C:\\Program Files\Test\` and will not be recognized as successfully saved. Tracking on the model will not restart and the save attempt will be made again on the next save attempt unnecessarily.

Changes at at https://github.com/gocodebox/lifterlms/compare/actual-saurabh-fix/backslash additionally unescape the response before returning data from the save API. This ensures the the unescaped data is compared to determine if the sync was successful.

I think with this change we're all good. I think.

actual-saurabh:
@thomasplevy That's a good catch. It works!

thomasplevy:
Thanks @actual-saurabh will merge into 3.24.1

Issue #525, opened 2018-05-01
Searching orders on admin panel should search partial email address as well as customer names


thomasplevy:
Currently only exact email addresses are matched

thomasplevy:
Resolved via #681

PR #681, opened 2018-10-07
Introduces a more robust solution for searching customer orders based on various user specific fields.

yojance:
## Description
See #525 

Introduces a more robust solution for searching customer orders based on various user specific fields. Main functionality was taken and slightly modified from WooCommerce's own customer's search feature. 

It searches the `wp_users` table for:
- user_login
- user_url
- user_email
- user_nicename
- display_name

It searches the `wp_usermeta` table for:
- first_name
- last_name

## How has this been tested?
Local database with a large number of orders with some different and some similar user data and then manually performed search queries and ensured the results were coming back as you'd expect.

## Checklist:

### DIFF
--- a/includes/admin/post-types/post-tables/class.llms.admin.post.table.orders.php
+++ b/includes/admin/post-types/post-tables/class.llms.admin.post.table.orders.php
@@ -245,9 +245,10 @@ class LLMS_Admin_Post_Table_Orders {
 
 
 	/**
-	 * Modify the search query for varios post types before retriving posts
-	 * @param  obj    $query  WP_Query obj
-	 * @return obj
+	 * Modify the search query for various post types before retrieving posts
+	 *
+	 * @param  $query WP_Query
+	 * @return WP_Query
 	 *
 	 * @since  2.5.0  moved from a non-classed function
 	 * @version  3.4.8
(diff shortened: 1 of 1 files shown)


actual-saurabh:
The build process for PHP 7.1 has timed out, otherwise the build is successful.

thomasplevy:
@yojance I've just reviewed all this and I found a small issue with the limit applied to the user queries. It looks to me like these limits prevent the full results from being displayed since the main query that the user metadata is being added to is already being limited. While removing this limit does result in a larger / slower query I don't see a better way to do it and still ensure the full results can be found.

Example: I register all of my test users with a `@lifterlms.com` test domain. If I wanted to see an orders list of all users with this domain I'd add this to the search results. I should see 15+ pages of results but with the limit in this PR I see *two* results. If I remove the limits I get the 15+ pages expected.

I've removed it and will merge it in with this small modification but I was curious if you have any thoughts on why I shouldn't remove it.

Thanks again

yojance:
@thomasplevy 

No particular reason at all. The search feature was slow on my local machine so I thought having a filter to set the number of results shown would be helpful. 

Glad to see this made it into core now 👍

Issue #694, opened 2018-10-16
"View as Visitor" allows you to take a quiz


thomasplevy:
### 1. Reproduction Steps

+ Add a quiz to a FREE lesson
+ Use "View as Visitor" from admin bar

### 2. Expected Behavior

+ The quiz should NOT be accessible because quizzes are not accessible to logged out users (even on free lessons)

### 3. Actual Behavior

The quiz is accessible when viewing as a visitor

### 4. Solution

View as functionality should be modified to show NO quiz button to reflect the actual user experience of a visitor / logged out user. Currently it's showing the quiz button because there's no logic built out to hide it

LiamBailey:
I would like to take this if possible?

thomasplevy:
All your @LiamBailey!

If you have any questions please let me know or join us in slack on the `#developers` channel if you want some realtime problem solving: https://lifterlms.com/slack

LiamBailey:
I have made a pull request: https://github.com/gocodebox/lifterlms/pull/705 - please email me on info[at]webbyscots.com if you have a demo site for testing. If not I will set up my own and test. Thanks.

thomasplevy:
@LiamBailey we don't host public testing sites. If you don't already have a way to test WordPress and WordPress plugins locally you can spool up a free temp site using http://poopy.life/ & then install your modified version of LifterLMS for testing

LiamBailey:
Yes sorry for the delay. I will get a test site up and get this tested. Thanks.

PR #705, opened 2018-10-27
Make quiz start only for logged in users

LiamBailey:
## Description
I have attempted to solve issue 694 by only showing the button to logged in users, and correcting the ajax to look for wp_error.


thomasplevy:
@LiamBailey have you tested this using the "preview management" "View as Visitor" feature from the admin bar exactly as the reproduction steps in #694 

I haven't pulled it down and ran a test but based on my understanding of how everything here works I do not believe that this will actually fix the identified issue.

Can you confirm your testing methods before I proceed with a more thorough test on the actual code?

Thanks

PR #722, opened 2018-12-12
Rename instances of `preprare_query()` to `prepare_query()`

tnorthcutt:
## Description
I noticed there were several instances of methods called `preprared_query()` (note the misspelling of the word "prepared". This misspelling seems to have been consistent so the code executed properly, but it's probably asking for future trouble to leave those misspellings in place.

### DIFF
--- a/includes/abstracts/abstract.llms.database.query.php
+++ b/includes/abstracts/abstract.llms.database.query.php
@@ -223,7 +223,7 @@ abstract class LLMS_Database_Query {
 	 * @since    3.8.0
 	 * @version  3.8.0
 	 */
-	abstract protected function preprare_query();
+	abstract protected function prepare_query();
 
 	/**
 	 * Execute a query
@@ -235,7 +235,7 @@ abstract class LLMS_Database_Query {
 
 		global $wpdb;
 
-		$this->sql = $this->preprare_query();
+		$this->sql = $this->prepare_query();
 		if ( ! $this->get( 'suppress_filters' ) ) {
 			$this->sql = apply_filters( $this->get_filter( 'prepare_query' ), $this->sql, $this );
 		}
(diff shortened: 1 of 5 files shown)


tnorthcutt:
@thomasplevy good plan. I started to implement those changes, but realized we'll run into an issue with the abstract method [here](https://github.com/gocodebox/lifterlms/pull/722/files#diff-7a8011b9ed472ade2e4199580bfeddddL226), since we can't declare an abstract `preprare_query()` and also define the implementation (because it's abstract).

Ideas?

thomasplevy:
well the backwards compatible `preprare_query()` doesn't have to be an abstract, the new version `prepare_query()` should be the abstract method... however, doing that will mean that I will actually have to run through all the extending classes and update them anyway because the class will throw errors if the abstract method isn't implemented so you're right that won't actually work...

I do not have a better idea at this moment. I don't have the physical capacity right now to ship an update to every add-on that currently utilizes this and even if I did what happens when someone updates the core and not the add-on or vice versa... I can do version dependency bumps on all of them to the next version that this is included in to prevent crashes but I really need to take some time to really work through all the add-ons and make the updates then.

For now we'll have to leave this pending until I can find the time to work through all the add-ons and figure out exactly how much code is affected by this.

Thanks for your patience

tnorthcutt:
>>> however, doing that will mean that I will actually have to run through all the extending classes and update them anyway because the class will throw errors if the abstract method isn't implemented so you're right that won't actually work...

Yep, makes sense! No worries on my end; sorry it turned out not to be a super simple fix!

thomasplevy:
Closing and tracking this as a bug at #829 

Sorry @tnorthcutt I can't figure out how to fix this properly

PR #734, opened 2019-01-10
update contributor name

eri-trabiccolo:
I've changed how my profile name appears in the contributors list so that it doesn't mess up the grid appearance anymore.


thomasplevy:
Thanks @eri-trabiccolo!

PR #741, opened 2019-01-22
Correct default account endpoint slugs

tnorthcutt:
## Description
Just noticed these default endpoint slugs were a little wonky; looks like the voucher and orders default endpoints didn't get added quite right in https://github.com/gocodebox/lifterlms/commit/31014137b72969620a92656a210725c2a2c2c05b


thomasplevy:
Nice catch @tnorthcutt, I'll merge this into 3.27.0. Thanks!

PR #742, opened 2019-01-23
Fix html element typo

tnorthcutt:


thomasplevy:
@tnorthcutt you're correct this is a typo but the `<h1>` breaks the layout without a bit of CSS applied.

If you'd like to make a few additional changes:

+ Remove the `h2` styles `/assets/css/admin-add-ons.scss` on lines 6-9 
+ Add a class to the `<h1>` in question `wp-heading-inline` which the WP core uses on screens like the plugins/themes screen for headlines.

This will fix the typo, remove unused styles, and ensure the layout doesn't break as a result.

As an aside, the broken layout will only be seen when using the LifterLMS Helper plugin which adds an element inline with the header to manage license keys for premium add-ons

tnorthcutt:
Interesting! I'll leave those changes to you guys 👍

thomasplevy:
See #743

PR #751, opened 2019-02-04
Remove unused code

tnorthcutt:
Probably undesirable to have old commented out code here, especially as it has resulted in un commented-out PHP being there. Not sure if that'll ever get executed, but I'm betting there's no reason for it to be there.


thomasplevy:
Yep, you're right. This is a WIP that I must've accidentally merged.

PR #753, opened 2019-02-05
Added class LLMS_Log_Handler

verygoodplugins:
## Description
For your consideration: demo of improved logging system, based on WooCommerce's and WP Fusion's logging systems.

- Added directory `/includes/logging/`
- Added class `LLMS_Log_Handler` in `/includes/logging/class.llms.log.handler.php`
- Added class `LLMS_Log_Table_List` in `/includes/logging/class.llms.log.table.list.php`

Logs can be written to using `LLMS()->logger->handle( $level, $user, $message, $context );`

Available levels are

- `error`
- `warning`
- `notice`
- `info`

`$user` expects a WordPress user ID. Pass `0` for system or unidentified user.

`$message` is a string.

`$context` (optional) is an array which can contain additional debugging information.

## How has this been tested?
It's basically an isolated system so it shouldn't affect any other parts of the code. I've confirmed that viewing, filtering, writing to, and flushing the logs all work as intended.

## Screenshots
Screenshot

## Types of changes
New feature. Introduces a database-based logging system where log entries can be filtered by user, severity, and source.

## Checklist:


thomasplevy:
@verygoodplugins,

Thanks so much for this contribution, however, after reviewing it I can't drop it in just yet. I have some things that I need to see happen before I'm comfortable with a merge. If you'd like to continue working on this PR you can help address the items below otherwise we'll tackle it internally as time allows:

- [ ]  Despite the fact that this follows the assumed naming convention of our files, we're working to transition filenames to follow this standard: https://github.com/gocodebox/lifterlms/blob/master/docs/coding-standards.md#file-names
- [ ] The "Logs" menu item should move to a tab under LifterLMS -> Status
- [ ] Add a constant which can be used to specify if text or db logging is used
- [ ] When DB logging is used `llms_log()` will log to this log handler. Otherwise, will utilize the existing text file logging
- [ ] If DB logging is enabled, show this log screen otherwise show the existing text logging viewer
- [ ] New installs should use DB logging by default, existing installs (sites which upgrade) will use text by default and can opt-in to DB logging. This is important to ensure we don't "lose" the ability to access text logs on existing sites unless we explicitly say we want to.

PR #786, opened 2019-03-14
Index auto-enrollment display table data by course ID instead of titl…

### LINKED ISSUE #785: [Membership] Auto-enrollment table doesn't display courses with identical titles
In https://github.com/gocodebox/lifterlms/blob/master/includes/admin/post-types/meta-boxes/class.llms.meta.box.membership.php#L46, the table data is indexed by title. If there are multiple courses with the same title, only one ends up displaying.

Replacing title with `$course_id` fixes this completely.

actual-saurabh:
…e, fix #785

## Checklist:

### DIFF
--- a/includes/admin/post-types/meta-boxes/class.llms.meta.box.membership.php
+++ b/includes/admin/post-types/meta-boxes/class.llms.meta.box.membership.php
@@ -41,6 +41,8 @@ class LLMS_Meta_Box_Membership extends LLMS_Admin_Metabox {
 
 			$course = new LLMS_Course( $course_id );
 
+			$title = $course->get( 'title' );
+
 			$data[ $course_id ] = array(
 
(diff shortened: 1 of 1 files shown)


thomasplevy:
Existing code is "lazy" sorting memberships alphabetically with `ksort` ([link](https://github.com/gocodebox/lifterlms/blob/a4ec8455a643d7be461f6adbecb6dcafdc6f7b5e/includes/admin/post-types/meta-boxes/class.llms.meta.box.membership.php))

With this PR we begin sorting by ID, which is maybe okay, but I want to consider the potential regression before I merge this in.

I've actually received feedback in support that the order is *confusing* because it does not allow users to choose the order in which courses are enrolled.

I'm wondering if the `ksort` should be removed and if we should add some jQuery to make the table sortable and then allow users to choose the order of the courses in this list manually.

Maybe I'm overthinking it but this does change some existing behavior in addition to fixing the bug.

PR #793, opened 2019-04-02
Add a filter for the section footer's "more" URL and text.

pondermatic:
## Description
Added the ability to filter the link URL and text in the dashboard section's footer.

## How has this been tested?
I have not successfully configured my WordPress development environment to run phpunit.
However, with this new hook, I am able to add a filter that changes the the "more" text.

## Types of changes

## Checklist:

### DIFF
--- a/templates/myaccount/dashboard-section.php
+++ b/templates/myaccount/dashboard-section.php
@@ -5,6 +5,7 @@
  * @version  3.19.5
  */
 defined( 'ABSPATH' ) || exit;
+$more = apply_filters( 'llms_' . $action . '_more', $more );
 ?>
 
 <section class="llms-sd-section <?php echo $slug; ?>">
@@ -21,7 +22,6 @@ defined( 'ABSPATH' ) || exit;
 
 	<?php if ( $more ) : ?>
 		<footer class="llms-sd-section-footer">
-			<?php $more = apply_filters( 'llms_' . $action . '_more', $more ); ?>
 			<a class="llms-button-secondary" href="<?php echo esc_url( $more['url'] ); ?>"><?php echo $more['text']; ?></a>
 		</footer>
 	<?php endif; ?>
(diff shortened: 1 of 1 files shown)


thomasplevy:
@pondermatic

Looks great but I'm going to nitpick a little... I know I don't have any specific coding standards that rule against what I'm about to say but I'm going to say this anyway:

I really dislike having variables defined inline in templates like this (and I KNOW there's areas in LifterLMS where this is done. In most of these existing pieces of code the templates SHOULD be refactored to avoid this).

Additionally, I don't like executing a single line of PHP in the midst of a template (again, there's code that does this, but I don't like it and I want it to be avoided).

In most cases I'd prefer the filter / code to be run in the function that calls the template or, when that's not possible, at the top of the template before the first closing `?>` tag.

Since this template is used by multiple functions it would mean adding this filter several times (which would be not DRY and annoying).

Simply moving the filter up to the top of the template would fix the problem I'm seeing and, additionally, would allow outputting footer information even if the core didn't already have some. Making this extendability addition even a bit more extendable.

If you make that change I'll get it merged in to the next release.

Thanks!

pondermatic:
@thomasplevy,

> I really dislike having variables defined inline in templates like this

> Additionally, I don't like executing a single line of PHP in the midst of a template

I agree and was only following the same pattern as how the title is filtered.

I will move the filter hook for $more to the top of the template. Should I move the filter hook for $title to the top as well or might that create a change in behavior?

BTW, thank you for the quick and thoughtful response. I like your reasoning.

thomasplevy:
@pondermatic,

> I agree and was only following the same pattern as how the title is filtered.

Life is full of inconsistencies... So, while I don't mind having a line like `<?php echo ... ?>` in the midst of a template I loathe having a line like `<?php $var = ... ?>`

If we weren't filtering an array it could be done on a single line without having to redefine the variable but it is an array...

Thanks for the update.

I'm working on improving our coding standards documentation and adding some documentation around templating is probably a good next step. See our existing here (with recent updates, but this stuff takes time and I have a lot of things that are arguably a bit more important than coding standard docs on my plate): https://github.com/gocodebox/lifterlms/tree/master/docs

PR #807, opened 2019-04-11
Make sure backslashes in post content are kept on cloning/importing

### LINKED ISSUE #596: Backslashes in post content disappear on cloning

eri-trabiccolo:
fix #596

## Description
Fix the cases when backslashes in content disappeared on cloning/importing

## How has this been tested?
I tested it by cloning, exporting and importing, a course (and a single lesson) with a backslash in its content.

## Types of changes

## Checklist:

### DIFF
--- a/includes/abstracts/abstract.llms.post.model.php
+++ b/includes/abstracts/abstract.llms.post.model.php
@@ -321,7 +321,7 @@ abstract class LLMS_Post_Model implements JsonSerializable {
 	 * @version  3.14.6
 	 */
 	private function create( $title = '' ) {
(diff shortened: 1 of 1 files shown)


thomasplevy:
@actual-saurabh can you review this to see if it fixes #596 properly

PR #809, opened 2019-04-12
Define missing properties

pondermatic:
## Description
Added missing property definitions.

## How has this been tested?
PHPUnit and PHPCodeSniffer.

## Types of changes
Added DocBlocks with description, `@var`, `@since` and `@deprecated` (where applicable) to new (and some existing) property definitions.
Most property definitions are now sorted alphabetically.

Properties that were not defined or defined with the `var` keyword are now defined with the `protected`
 keyword if the property is only used within that class's hierarchy. This may break other WordPress plugins that expect the properties to be `public`.

## Checklist:

### DIFF
--- a/includes/Users/User.php
+++ b/includes/Users/User.php
@@ -4,7 +4,29 @@ namespace LLMS\Users;
 
 class User
 {
+	/**
+	 * @var string
+	 * @since 1.3.0
+	 */
+	protected $first_name;
+
+	/**
+	 * @var int
+	 * @since 1.3.0
+	 */
 	private $id;
+
+	/**
+	 * @var string
+	 * @since 1.3.0
+	 */
+	protected $last_name;
+
+	/**
+	 * @var array
+	 * @since 1.3.0
+	 */
+	protected $quiz_data;
 
 	/**
 	 * Constructor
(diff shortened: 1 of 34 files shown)


thomasplevy:
@pondermatic thanks!!!

thomasplevy:
@pondermatic I looked through this too quickly and didn't realize that some properties which were undefined (therefore public) are now private.

Since I cannot guarantee one way or the other that this won't cause issues I can't in good conscience merge it in.

Could you resubmit with the instances of undeclared or mis-declared properties being declared as public just to be safe. We can responsibly change them later if we have to.

Thanks!

pondermatic:
I will change the newly defined (and redefined from `var`) to public. I to am worried that it make affect other plugins. I thoroughly checked all of the LifterLMS code.

pondermatic:
BTW, when I defined properties that hold an array, I default the value to empty array instead of null.

thomasplevy:
> I will change the newly defined (and redefined from var) to public. I to am worried that it may affect other plugins. I thoroughly checked all of the LifterLMS code.

Thanks!

> BTW, when I defined properties that hold an array, I default the value to empty array instead of null. Let me know if you think that this is OK.

I think this is the right way to go, thanks

PR #815, opened 2019-04-12
Define missing properties

pondermatic:
## Description
Added missing property definitions.

## How has this been tested?
PHPUnit and PHPCodeSniffer.

## Types of changes
Added DocBlocks with description, `@var`, `@since` and `@deprecated` (where applicable) to new (and some existing) property definitions.
Sorted property definitions alphabetically.

Properties that were not defined or defined with the `var` keyword are now defined with the `public`
visibility keyword (instead of `protected`). This preserves compatibility with other WordPress plugins that expect the properties to be `public`.

## Checklist:

### DIFF
--- a/includes/Users/User.php
+++ b/includes/Users/User.php
@@ -4,7 +4,29 @@ namespace LLMS\Users;
 
 class User
 {
+	/**
+	 * @var string
+	 * @since 1.3.0
+	 */
+	public $first_name;
+
+	/**
+	 * @var int
+	 * @since 1.3.0
+	 */
 	private $id;
+
+	/**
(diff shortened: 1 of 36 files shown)


pondermatic:
I can not find a reason why `ci/circleci: test-php-71` failed running `LLMS_Test_Functions_Access_Plans::test_llms_insert_access_plan_update`. That test does not appear to use any code that I changed. The tests pass on my machine with PHP 7.1.28 and 7.2.11.
PR #809 passed the unit tests and the only difference is that it defined `protected` properties and this one uses `public`.

PR #823, opened 2019-04-16
Fix typo in 'php_time_limit' key name.

pondermatic:
## Description
Corrected typo in the `php_time_limit` key name.

## How has this been tested?
PHPUnit and PHP_CodeSniffer

## Types of changes
The server data key names are displayed on the [System Report](/wp-admin/admin.php?page=llms-status&tab=report) page. Unless someone is screen scraping that page, we shouldn't be worried about breaking expected behavior.
However, the server data is also used in the `LLMS_Tracker` class and I'm not sure how. The incorrect and corrected versions of the `php_time_limit` key are not found anywhere else in LifterLMS.

## Checklist:

### DIFF
--- a/includes/class.llms.data.php
+++ b/includes/class.llms.data.php
@@ -408,7 +408,7 @@ class LLMS_Data {
 			$data['php_max_input_vars'] = ini_get( 'max_input_vars' );
 			$data['php_memory_limit'] = ini_get( 'memory_limit' );
 			$data['php_post_max_size'] = ini_get( 'post_max_size' );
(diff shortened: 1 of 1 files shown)


thomasplevy:
Before a merge I need to check our telemetry data API as well as workflows relevant to customer support requests which use data generated by this function (in the system report) and ensure no issues will appear in these areas as a result of from fixing this typo here.

thomasplevy:
Telemetry API suffers from this typo and requires an update to accommodate accepting either the typo or correct version of the key before this can be deployed.

PR #843, opened 2019-04-30
Allow complete enrollments deletion from the database

eri-trabiccolo:
## Description
Added the logic to completely delete enrollments from the database.
Additionally  all icons on the course/membership student management list table now implements the llms tooltip data attribute api to add popover tooltips exposing what the button does.

## How has this been tested?
I've tested the new feature making sure that:

* Enrollment post metas were correctly deleted when:
  * the Enrollment status was set to `Deleted` on Order edit screen: `Deleted` is the new fictitious status exposed as an enrollment status on the order;
  * the Order post type related to an enrollment was completely deleted;
  * the Enrollment was manually deleted via the new trash can icon action button on the course/membership student management list.

* The Order status was set to `Cancelled` when:
  * the Enrollment "status" was set to `Deleted `on Order edit screen.

## Types of changes

## Checklist:

## Considerations:
* At the moment you can unenroll a student from the Order screen and then be able to delete it from the course/membership student management list:
Schermata 2019-04-30 alle 15 31 39
Doing that the Order status won't be switched to `Cancelled`. Is what we want?!
* Instead of exposing `Deleted` as an enrollment status on the order, since it's not an actual enrollment status (when deleted an enrollment just doesn't exist anymore :D) wouldn't be better to just add a new button in the enrollment metabox next to the "Update Enrollment Status" button:
Schermata 2019-04-30 alle 15 38 29


thomasplevy:
Thanks @eri-trabiccolo, probably not going to get to this today but I'll take a look tomorrow

thomasplevy:
I'm not going to get a review in on this today. Expect that tomorrow afternoon with a merge next week.

includes/models/model.llms.student.php
```diff
@@ -1435,6 +1462,73 @@ public function unenroll( $product_id, $trigger = 'any', $new_status = 'expired'
...
+				$delete = true;
+
+			}
+		}
+
+		// delete if we can
+		if ( $delete ) {
+
+			// Unenroll the student if enrolled
+			$this->unenroll( $product_id, $trigger, 'expired' );
```
thomasplevy: I don't think we need to unenroll before deleting. At this point they should already be unenrolled based on the interface, however, if someone wants to programmatically delete without running those checks that's just fine by me, I think.

I didn't run this but I think if this line is removed it just cuts down on the number of queries that need to be executed.

If I'm reading this all properly we should be able to simply delete the records without unenrolling first.

includes/controllers/class.llms.controller.orders.php
```diff
@@ -304,6 +310,24 @@ public function error_order( $order ) {
...
+	 * Will delete any enrollment records linked to the LLMS_Order with the ID of the deleted post
+	 *
+	 * @since [version]
+	 *
+	 * @param int $post_id WordPress post ID
+	 * @return void
+	 */
+	public function on_delete_order( $post_id ) {
+
+		$order = new LLMS_Order( $post_id );
```
thomasplevy: it'll be safer to use `llms_get_post()` here.

I'd write it like this:
```
$order = llms_get_post( $post_id );
if ( $order && is_a( 'LLMS_Order', $order ) ) {
...
```

assets/js/llms-metabox-students.js
```diff
@@ -117,13 +123,27 @@
 			this.update_student_enrollment( $el.attr( 'data-id' ), 'remove' );
 		};
 
+		/**
+		 * Delete a Student's enrollment.
+		 *
+		 * @since 3.0.0
```
thomasplevy: remove duplicate `@since` in favor of `@since [version]` below

includes/models/model.llms.student.php
eri-trabiccolo: I thought the same at the beginning, also to have this method doing one and one thing only.
But then I thought that in the rare case someone wanted to use this public method it should trigger the unenroll as well.

But yes I'm going to remove this line and amend the documentation here:
https://github.com/gocodebox/lifterlms/pull/843/files/4c7bbab3ed0717eeb3fc0fba2d01a67ea72eb097#diff-337dc0e7959977273f0d8630e90dd68fR403
replacing it with something that says the opposite, like "this method doesn't unenroll the student, you might want to run llms_unenroll_student before"

includes/controllers/class.llms.controller.orders.php
eri-trabiccolo: fine, thanks.

fixed in e75d820

assets/js/llms-metabox-students.js
eri-trabiccolo: right
fixed in 792ee4f

includes/models/model.llms.student.php
eri-trabiccolo: I've changed idea about the documentation "replacement", I don't think it's needed.
Fixed in 3e0ced7

eri-trabiccolo:
Will work on the needed tests tomorrow :D

eri-trabiccolo:
Hi @thomasplevy 
implementing the tests I realized that basically this condition:
https://github.com/gocodebox/lifterlms/pull/843/files#diff-25a25e787bb5217d6b8361f9d8b08b0aR832
will never match when the requested "status" is `delete`, because `llms_delete_student_enrollment` will basically never return `false`.

I implemented the `llms_bulk_delete_user_postmeta`:
https://github.com/gocodebox/lifterlms/pull/843/files#diff-5d3048699bc7c0a2db351fc11862b606R40

(wrongly) assuming that the `llms_delete_user_postmeta`:
https://github.com/gocodebox/lifterlms/blob/3.32.0-beta.2/includes/functions/llms.functions.user.postmeta.php#L21

would return false also if no postmeta exists (nothing to delete), but actually it will only if something wrong occurs during the actual deleting of a db row for an existing postmeta.

This means that also the following action can be triggered when actually nothing has been removed as nothing to remove was there:
`do_action( 'llms_user_enrollment_deleted', $this->get_id(), $product_id );`
https://github.com/gocodebox/lifterlms/pull/843/files#diff-c80bdb2a329ed5e6acc4754d78b7e7f8R1517

What do you think about this?
Do you think we need to "improve" either `llms_delete_user_postmeta` and/or `llms_bulk_delete_user_postmeta` to inform us about what was deleted and what was simply not existent?!

Otherwise before saying an enrollment has been deleted we should check it existed... with a db query (:/).

Let me know if we can live with the current state or not, when you can/want.

includes/models/model.llms.student.php
```diff
@@ -1435,6 +1462,70 @@ public function unenroll( $product_id, $trigger = 'any', $new_status = 'expired'
...
+	 * Delete a student enrollment.
+	 *
+	 * @since [version]
+	 *
+	 * @param  int     $product_id  WP Post ID of the course or membership.
+	 * @param  string  $trigger     Optiona. Only delete the student's enrollment if the original enrollment trigger matches the submitted value
+	 *                              "any" will remove regardless of enrollment trigger. Default "any".
+	 * @return boolean Whether or not the enrollment records have been succesfully removed.
+	 *
+	 * @see  llms_delete_student_enrollment() calls this function without having to instantiate the LLMS_Student class first.
```
thomasplevy: Despite the pre-existing `@see` tags in other areas of this class this tag should appear below the changelog tags (`@since`) see doc: https://github.com/gocodebox/lifterlms/blob/master/docs/documentation-standards.md#functions-and-class-methods

thomasplevy:
@eri-trabiccolo looks really solid on the tests. I'm going to do a thorough review of this tomorrow.

Thanks!

includes/models/model.llms.student.php
eri-trabiccolo: right, bad copy&paste :D 
fixed in 9bd326b

eri-trabiccolo:
> You're correct. `llms_delete_user_postmeta()` should have an additional condition to return false when the existing postmeta doesn't exist at all.

implemented in 5dae869
let me know if you think it's wron/not enough, also because that's part of the public api...

eri-trabiccolo:
I've added some functional tests in dc2c4bb
where I only check the tested methods return what expected and trigger the expected actions: e.g. i don't check whether or not a user is enrolled after calling $student->is_enrolled, as these kind of tests are already covered in `tests/unit-tests/user/class-llms-test-student.php`, but if you think a redundancy here makes sense, just let me know :D

PR #845, opened 2019-05-01
Pass $check_availability as parameter for the acces plan checkout url…

### LINKED ISSUE #844: Access plan checkout url filter should pass `get_checkout_url` input parameter
I needed to filter an access plan checkout url using the filter hook defined at the following line:
https://github.com/gocodebox/lifterlms/blob/master/includes/models/model.llms.access.plan.php#L318
and I've found odd not to have the `$check_availability` parameter available (which is the - optional - input parameter for the `get_checkout_url` method).

Would like to have the `$check_availability` parameter passed to the filter callback.

nothing

**Possible solution**
I would add a new filter called 'llms_access_plan_get_checkout_url':
`apply_filters( 'llms_access_plan_get_checkout_url', $ret, $check_availability, $this );`
and deprecate the 'llms_plan_get_checkout_url' filter.
That's because a simple addition of the `$check_availability` as third parameter of the current filter, while addressing backward compatibility, won't respect the "standard" I can see for all the filters used there, which wants the access plan instance (`$this`) passed as latest parameter.
A small PR is on the way.

eri-trabiccolo:
… filter

fix #844

## Description
Add a new filter for the access plan checkout url, so that the `$check_availability` parameter can be passed but keeping backward compatibility.

## How has this been tested?
I've hooked into the new and old filter checking everything was fine.

## Types of changes

## Checklist:

### DIFF
--- a/includes/models/model.llms.access.plan.php
+++ b/includes/models/model.llms.access.plan.php
@@ -6,7 +6,7 @@
  * @package  LifterLMS/Models
  *
  * @since    3.0.0
- * @version  3.30.1
+ * @version  [version]
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
Done, despite the test failed on CircleCI which I think is a false positive...

PR #847, opened 2019-05-03
Add Memberships reporting tab

eri-trabiccolo:
## Description
Add Memberships reporting tab.

## How has this been tested?
I've played with the new reporting tab and subtabs, checking they behave as expected without throwing any error/notice.

## Types of changes
New feature

## Checklist:

## Consideration:
Even after the introduction of the `LLMS_Abstract_Post_Data` class you can still see a bunch of code in "common" between the `LLMS_Membership_Data`  and `LLMS_Course_Data`. Maybe an additional more generic class can be introduced to avoid code repetition.

Also, the new Membership table|tabs classes defined in the reporting/tables|tabs/*membership* files, are mostly copies (or partial copies with very slightly differences) of the Course ones...

### DIFF
--- a/includes/admin/reporting/class.llms.admin.reporting.php
+++ b/includes/admin/reporting/class.llms.admin.reporting.php
@@ -13,6 +13,8 @@ defined( 'ABSPATH' ) || exit;
  *
  * @since 3.2.0
  * @since [version] Fix redundant `if` statement in the `output_widget` method.
+ * @since [version] Added Memberships tab.
+ * @since [version] The `output_event()` method now outputs the student's avatar whent in 'membership' context.
  */
 class LLMS_Admin_Reporting {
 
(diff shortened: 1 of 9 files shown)


eri-trabiccolo:
I surely have some issue with the automatic phpcs checks as when I run `composer run phpcs` I don't get all those errors codeclimate keeps fon inding... :(

thomasplevy:
> I surely have some issue with the automatic phpcs checks as when I run composer run phpcs I don't get all those errors codeclimate keeps fon inding... :(

I haven't updated our PHPCS to run the full ruleset. Mostly because there are too many broken rules for me to manage. Over time we're improving little by little.

thomasplevy:
most of the issues are not CS rules though, their found by phpmd (php mess doctor). As a rule I don't explicitly deny PRs based on these but I like them to still show as they help us be aware of better practices and strive to have less messy code based on these things. Sometimes I find it damn near impossible to reduce cognitive complexity of complicated functions though... /shrug

eri-trabiccolo:
Ah yeah I see... I don't understand how to reduce the cognitive complexity sometimes either, which I guess is a new cognitive complexity issue 🤣

PR #856, opened 2019-05-14
add a "Display Name" field to student's profile information

### LINKED ISSUE #691: Add a "Display Name" field to student's profile information (privacy concern when commenting)

Currently LifterLMS does not supply a `display_name` to `wp_insert_user()` when registering new users.

If account settings require (or allow) a name during registration the user's full name will be used and displayed publicly when commenting and the user will have no way to customize their name.

Add a "Display Name" text field to the "Edit Account" area of the student dashboard so student's have the ability to choose their public display name.

LifterLMS Social Learning already implements this solution. If we add this to the core we should remove it from Social Learning to ensure there's no duplicate fields when using the add-on

eri-trabiccolo:
fix #691

## Description
add a "Display Name" field to student's profile information in the student's dashboard.

## How has this been tested?
Tested altering and saving the new Display Name field in the student's dashboard.

## Types of changes
New feature

## Checklist:

### DIFF
--- a/includes/class.llms.person.handler.php
+++ b/includes/class.llms.person.handler.php
@@ -1,9 +1,9 @@
 <?php
 /**
- * User Handling for login and registration (mostly)
+ * User Handling for login and registration (mostly).
  *
  * @since 3.0.0
- * @version 3.35.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -13,6 +13,8 @@ defined( 'ABSPATH' ) || exit;
  *
  * @since 3.0.0
  * @since 3.35.0 Sanitize field data when filling field with user-submitted data.
+ * @since [version] Added a "Display Name" text field to the "Edit Account" area of the student dashboard.
+ *               Also added `llms_pre_save_display_name` filter.
  */
 class LLMS_Person_Handler {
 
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
This actually doesn't mean to be the final version, but more a field on which we can discuss.
The display_field is also added by a LLMS add-on that is going to be updated as well...
What's the lllms way to avoid "conflicts" with not up to date add-on. e.g:
when an user updates LLMS core but doesn't update the add-on.

eri-trabiccolo:
so here I will use the same approach suggested in https://github.com/gocodebox/lifterlms-social-learning/issues/30#issuecomment-492703123
a check on the LifterLMS Social Learning version, will be fine.

eri-trabiccolo:
Is there a more beautiful way to check an integration version than the one below :/ ?

```
		if ( LLMS()->integrations()->get_integration( 'social_learning' ) &&
				function_exists( 'LLMS_Social_Learning' ) &&
				property_exists( LLMS_Social_Learning(), 'version' ) &&
				version_compare( LLMS_Social_Learning()->version, '1.2.1', '>' )
		) { 
```

thomasplevy:
We don't need to run a "compat" version check on social learning within the core. The issue would be we don't want duplicate fields displaying if a user upgrades the core but not the add-on, right?

It'll be a small number of impacted users in comparison to the number of users that aren't using Social Learning who don't need an additional check

eri-trabiccolo:
That makes sense 😃

thomasplevy:
Overall looks solid, will merge tomorrow (probably)

eri-trabiccolo:
This PR needs to be totally re-worked when the new way to create the forms will be in place.

thomasplevy:
I know I'm sorry @eri-trabiccolo

thomasplevy:
I'm going to close it because it does need to be fixed with a block in the editor. Sorry to waste your time on this :-/

PR #857, opened 2019-05-15
Post filter on Reporting students table: filter post results based on…

### LINKED ISSUE #850: Post filter on Reporting students table doesn't filter post results based on instructor access
### Reproduction Steps

+ Navigate to Reporting -> Students (as an instructor)
+ Use the Course/Membership filter to filter results

### Expected Behavior

+ Should only be able to select courses/membership the instructor has access to

### Actual Behavior

+ All courses/memberships on the site will be listed / can be selected
+ The `LLMS_Instructor->get_students()` method prevents students from other courses/memberships from being displayed but the other posts should still not be listed in the dropdown.

---

I don't think the method `select2_query_posts()` should blanket filter posts by instructor by default. 

Instructors can always view the names of other courses/memberships on the site so by exposing the list for searching there's no inherit issue. However, in this particular case it causes a confusing UX.

I think the method should optionally filter by the instructor's information. Perhaps by passing additional information via data attributes. Maybe...

eri-trabiccolo:
… instructor access

fix #850

## Description
Added logic to limit the posts query performed by `LLMS_AJAX_Handler::select2_query_posts()` to only those posts accessible by a certain instructor.

## How has this been tested?
Tested on Reporting students table making sure that in the post filter select2 I could see:
1) only course/memberships for which I was set as instructor, **when accessing as instructor**
2) all the available course/memberships, **when accessing as "administrator" or LMS Manager**

And via unit tests.

## Types of changes
Bug fix

## Checklist:

### DIFF
--- a/assets/js/llms-admin.js
+++ b/assets/js/llms-admin.js
@@ -1,12 +1,13 @@
 /**
  * LifterLMS Admin Panel Javascript
  *
- * @since    ??
- * @since    3.32.0 `llmsPostsSelect2` function allows posts fecthing based on post statuses.
- * @version  3.32.0
+ * @since ??
+ * @since 3.32.0 `llmsPostsSelect2` function allows posts fecthing based on post statuses.
+ * @since [version] `llmsPostsSelect2` function allows posts (llms posts) fetching filtered by their instructor id.
+ * @version [version]
  *
- * @param    obj   $  traditional jQuery reference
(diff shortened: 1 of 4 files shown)


thomasplevy:
@eri-trabiccolo I don't know why I've ignored this for so long. It's perfect. Thank you <3

PR #868, opened 2019-05-20
Admin account order view: pass order instance as parameter to all hooks

### LINKED ISSUE #402: View Order hooks doesn't have the order id
Looking on https://github.com/gocodebox/lifterlms/blob/master/templates/myaccount/view-order.php#L29 seems that these filters doesn't give to the develop what is the order id to use.
The only way to get that value is to write: `$wp->query_vars['orders']`.

To simplify the development will be helpful to pass this as parameter in all the hook in that file.

Ps: I developed a tiny plugin https://github.com/CodeAtCode/lifterlms-attach-pdf that show a file uploaded on the backend of the order (we will use for invoice made manually). The code where I am using the filter https://github.com/CodeAtCode/lifterlms-attach-pdf/blob/master/public/LifterLMS_Attach_PDF.php#L57

eri-trabiccolo:
fix #402

## Description
Added a further param to the various `do_action`, `apply_filters` calls so to pass the current order instance to the hook callbacks in templates/myaccount/view-order.php

## How has this been tested?
Just verified no errors were triggered accessing that view plus automated tests passed.

## Types of changes
new feature?

## Checklist:

### DIFF
--- a/templates/myaccount/view-order.php
+++ b/templates/myaccount/view-order.php
@@ -1,8 +1,10 @@
 <?php
 /**
- * View an Order
- * @since    3.0.0
- * @version  3.19.0
+ * View an Order.
+ *
+ * @since 3.0.0
+ * @since [version] Pass the current order object instance as param for all the actions and filters.
+ * @version [version]
  */
 defined( 'ABSPATH' ) || exit;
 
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
I've added a non related commit to get rid of this redundant check. I think it's safe, nothing should (be able to) `nullify` the $order instance after the first check, right?!

templates/myaccount/view-order.php
```diff
@@ -209,7 +207,7 @@
 			'transactions' => $transactions,
 		) ); ?>
 
-		<?php do_action( 'lifterlms_after_view_order_table' ); ?>
+		<?php do_action( 'lifterlms_after_view_order_table', $order ); ?>
 
 	<?php endif; ?>
```
thomasplevy: tsk tsk tsk :-p

eri-trabiccolo: yeah sorry :/

Issue #870, opened 2019-05-23
[Capabilities] Export and Clone links should check `edit_course` capability


actual-saurabh:
Unlike the link to Builder which checks the `'edit_course'`capability: https://github.com/gocodebox/lifterlms/blob/f93a34750bfc8b2390239129a112fb5f3686b43d/includes/admin/post-types/post-tables/class.llms.admin.post.table.courses.php#L35, the export and clone links check the `'edit_post'` capability: https://github.com/gocodebox/lifterlms/blob/cdbee2b4b3bfabf3339bd6870f3d3af8e9227da0/includes/admin/post-types/class.llms.post.tables.php#L38 and https://github.com/gocodebox/lifterlms/blob/cdbee2b4b3bfabf3339bd6870f3d3af8e9227da0/includes/admin/post-types/class.llms.post.tables.php#L47, which leads to unexpected behaviour when fine tuning user capabilities using custom code or plugins.

PR #877, opened 2019-06-01
Fix/export clone capability

actual-saurabh:
## Description

 * Fixed capability as per #870 to hide export & clone links on posts table for unauthorised users.
 * Fixed superglobal access by refactoring to `llms_filter_input()`.
 * Add additional capability check so that unauthorised direct export/clone requests are prevented.

## How has this been tested?

- [x] User with right capability is able to export/clone as usual
- [x] User without such a capability on a course doesn't see the links on post table
- [x] If such a user attempts to directly request the action, they get an error.

## Types of changes

 * Bug fix for #870 

## Checklist:

### DIFF
--- a/includes/admin/post-types/class.llms.post.tables.php
+++ b/includes/admin/post-types/class.llms.post.tables.php
@@ -68,7 +68,7 @@ class LLMS_Admin_Post_Tables {
 		$action = llms_filter_input( INPUT_GET, 'action' );
 
 		// bail early if request doesn't concern us.
-		if ( empty( $action) ){
+		if ( empty( $action ) ){
 			return;
 		}
 
--- a/includes/admin/post-types/class.llms.post.tables.php
+++ b/includes/admin/post-types/class.llms.post.tables.php
@@ -67,43 +67,43 @@ class LLMS_Admin_Post_Tables {
 
 		$action = llms_filter_input( INPUT_GET, 'action' );
 
-		// bail early if request doesn't concern us
+		// bail early if request doesn't concern us.
 		if( empty( $action) ){
 			return;
 		}
 
-		// bail early if it isn't a clone/ export request
+		// bail early if it isn't a clone/ export request.
 		if ( 'llms-clone-post' !== $action && 'llms-export-post' !== $action ) {
 			return;
 		}
 
 		$post_id = llms_filter_input( INPUT_GET, 'post' );
 
-		// bail if there's no post ID
+		// bail if there's no post ID.
 		if ( empty( $post_id ) ) {
 			wp_die( __( 'Missing post ID.', 'lifterlms' ) );
 		}
 
 		$post = get_post( $post_id );
 
-		// bail if post ID is invalid
+		// bail if post ID is invalid.
 		if ( ! $post ) {
 			wp_die( __( 'Invalid post ID.', 'lifterlms' ) );
 		}
 
-		// bail if the action isn't supported on post type
+		// bail if the action isn't supported on post type.
 		if ( ! post_type_supports( $post->post_type, $action ) ) {
 			wp_die( __( 'Action cannot be executed on the current post.', 'lifterlms' ) );
 		}
 
-		// bail if user doesn't have permissions
+		// bail if user doesn't have permissions.
 		if( ! current_user_can( 'edit_course', $post->ID ) ) {
 			wp_die( __( 'You are not authorized to perform this action on the current post.', 'lifterlms' ) );
 		}
 
 		$post = llms_get_post( $post );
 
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
hi @actual-saurabh I was on this yesterday but I actually was not really able to reproduce the issue in the first place. Even if technically understand it :D 
Any hints on how to reproduce it?

Anyways I've implemented a slightly different code which takes into account 
1) the current post type (e.g. the clone link is available for lesson post types too)
2) the fact that the llms post type caps are filterable.

Can we merge our approaches into your PR, what do you think?
https://github.com/eri-trabiccolo/lifterlms/commit/de92515032e13a0068802090ea16fedc0d63619c#diff-d94f1fc908f1be0f1689867e15e7fc39R47

actual-saurabh:
@eri-trabiccolo 

- [ ] The filterability is something that I missed. That makes a lot of sense.
- [ ] In fact, a similar approach should be used in https://github.com/gocodebox/lifterlms/blob/f93a34750bfc8b2390239129a112fb5f3686b43d/includes/admin/post-types/post-tables/class.llms.admin.post.table.courses.php#L35
- [ ] Additionally, this should also be used in https://github.com/gocodebox/lifterlms/pull/877/commits/f3aa79aed3bad7d40266ff36212a25bf4342eac9#diff-d94f1fc908f1be0f1689867e15e7fc39R100

Thanks!

eri-trabiccolo:
I see, but my "problem" is that I still don't see what we're trying to fix here :), so I've investigated furthermore into this...

I mean, `current_user_can( 'edit_course', $post->ID );`, when `$post` is a `course` cpt, and the cpt meta cap `edit_post` is mapped to `edit_course` (as it is by default) isn't the same as `current_user_can( 'edit_post', $post->ID );`?

`edit_post` is a meta cap,  so when I call  `current_user_can( 'edit_post', $post->ID );` on a cpt, it has to map it to a primitive cap, see (note that [map_meta_cap is true](https://github.com/gocodebox/lifterlms/blob/master/includes/class.llms.post-types.php#L273)):
https://github.com/WordPress/WordPress/blob/5.2.1/wp-includes/capabilities.php#L152
hence (according to the code above linked) it will break down to:
`$post_type->cap->edit_published_posts;`
`$post_type->cap->edit_posts;`
`$post_type->cap->edit_others_posts;`
or
`$post_type->cap->edit_private_posts;`

When, instead we call, `current_user_can( 'edit_course', $post->ID );` on the course cpt, we are in this case:
https://github.com/WordPress/WordPress/blob/5.2.1/wp-includes/capabilities.php#L576
(`edit_course` is a `course` post_type meta cap https://github.com/gocodebox/lifterlms/blob/master/includes/class.llms.post-types.php#L175 )
hence map_meta_cap will call itself passing `edit_post` as cap to check against (since `$post_type_meta_caps['edit_course'] == 'edit_post'`) and we're back to square one.

So either we call `current_user_can( 'edit_post', $post->ID );` or  `current_user_can( 'edit_course', $post->ID );` on a course cpt map_meta_cap will break down to, e.g. if we're the author to the course, to:
`$post_type->cap->edit_published_posts;` => `'edit_published_courses'`

I'm sure I'm missing something...

actual-saurabh:
@eri-trabiccolo as mentioned in https://github.com/gocodebox/lifterlms/issues/870#issue-447771170, this leads to unexpected behaviour when using custom code or a capability manager plugin to fine tune user permissions and users who don't have the `edit_course` capability but have the `edit_post` one are still able to view the export and clone links and perform those actions.

eri-trabiccolo:
Ok, but `edit_course`, the way it is implemented in lifterlms, is a meta cap as `edit_post` is: users don't have such a cap, this cap is mapped to other caps a user can have.
Both of them do not exist as cap and break down to
```
$post_type->cap->edit_published_posts
$post_type->cap->edit_posts
$post_type->cap->edit_others_posts
```
or
```
$post_type->cap->edit_private_posts
```
hence to:
```
edit_published_courses
edit_courses
edit_others_courses
```
or
```
edit_private_courses
```
(for a `course` post type, given the standard llms configuration).

As mentioned in my previous comment (https://github.com/gocodebox/lifterlms/pull/877#issuecomment-498048567) checking `edit_course` user permission on a course cpt will end up checking `edit_post` on it so in this case, my code, which has the only advantage of not hard-coding the `edit_course`/`edit_lesson` meta cap, doesn't really add anything special.

So I was wondering whether we're introducing code to keep llms safe from 3rd party plugin misconfiguration and/or incomplete/wrong custom code rather than actually fixing something.
That's all.

Thanks for your time.

PR #882, opened 2019-06-11
Fix data types in DocBlocks.

pondermatic:
## Description
Fixed data types in DocBlocks.


includes/notifications/class.llms.notifications.php
```diff
@@ -323,6 +324,7 @@ public function load_processor( $type, $path = null ) {
 	 * Load a single view
 	 * @param    string     $trigger  trigger id (eg: lesson_complete)
 	 * @param    string     $path     full path to the view file, allows third parties to load external views
+	 * @param    string     $prefix
```
thomasplevy: Needs a description:

Classname prefix. Defaults to "LLMS". Can be used by 3rd parties to adjust the prefix in accordance with the projects standards.

PR #885, opened 2019-06-19
Add support for modules

actual-saurabh:
## Description

This adds autoloading support for modules inside `includes/modules/`.

Modules have an expected naming convention without which they won't work:

- The name of the directory is the name of the module. That is, for a folder `includes/modules/certificate-builder`, the name of the module is `'certificate-builder'`.
- The main class for  a module is expected inside the module name folder as `class-llms-{modulename}.php`. That is, in the same example as above, the main class that will be included is `class-llms-certificate-builder.php`.
- Each module will have an associated constant that acts as a toggle for the module and the name will be `LLMS_{ALLCAPSMODULENAME}`. Setting this constant explicitly to false inside `wp-config.php` will disable the module. For instance, to disable the Certificate Builder module, add the following:

```php
define( 'LLMS_CERTIFICATE_BUILDER', false );
```
- Modules (both internal and external, for example from plugins and addons) can use their own naming conventions and load by hooking into a `lifterlms_modules` filter. This will filter an associative array of available modules where each module is in turn an array:

```php
$modules['module_name'] = array(
    'name' => 'module_name',
    'file_path' => 'path/to/the/main-file-to-be-included.php',
    'constant' => 'TOGGLE_CONSTANT_MODULE_NAME'
);
```

## Checklist:

### DIFF
--- a/includes/modules/class-llms-modules.php
+++ b/includes/modules/class-llms-modules.php
@@ -1,0 +1,108 @@
+<?php
+
+class LLMS_Modules {
+
+	public $loaded = array();
+
+	protected static $_instance = null;
+
+	public static function instance() {
+		if ( is_null( self::$_instance ) ) {
+			self::$_instance = new self(); }
+		return self::$_instance;
+	}
+
+	/**
+	 * Constructor
+	 *
+	 * @since    [version]
+	 * @version  [version]
+	 */
+	private function __construct() {
+		$this->load();
+	}
+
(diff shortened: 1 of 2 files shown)


thomasplevy:
@actual-saurabh This looks generally good but I don't want to merge it in until we actually start using this so I'm going to close this PR and we can merge this in later when the cert builder stuff is done and ready for a merge.

PR #894, opened 2019-07-03
Allow getting LLMS_Post_Model raw properties

### LINKED ISSUE #893: Getting LLMS_Post_Model raw properties
Implementing the Courses resource rest-api we often need to access to the model's "raw" properties.

Unless I'm not missing something, sometimes, this is not really straightforward using the model api. For instance, how can I get the Course's raw content?
`$course->get('content')` will give me the [llms_content for that course](https://github.com/gocodebox/lifterlms/blob/3.33.2/includes/abstracts/abstract.llms.post.model.php#L156).
That might be fine, I can get the course's `WP_Post` and access the $post->content;

But what if I need the course's `capacity_message`? That prop is "scrubbed" (given its type `text`, this means that its tags are stripped) we basically have to duplicate some model's code and skip the scrub.

What do you think if we add some code to the LLLMS_Post_Model's in order to get the "raw" properties?

Also, is it fine to open this kind of issues here or should I open them on the lifterlms repo? :D

eri-trabiccolo:
fixes #893

## Description
Adds a parameter to the `LLMS_Post_Model->get` method in order to retrieve "raw" properties (aka unfiltered).

## How has this been tested?
I've tested this didn't introduce any regression navigating the front courses.

## Types of changes
Possibly a breaking change?!

## Checklist:

### DIFF
--- a/includes/abstracts/abstract.llms.post.model.php
+++ b/includes/abstracts/abstract.llms.post.model.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Abstracts
  *
  * @since 3.0.0
- * @version 3.31.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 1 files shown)


includes/abstracts/abstract.llms.post.model.php
```diff
@@ -402,13 +339,108 @@ public function export() {
...
+	 * @param string  $key The property key.
+	 * @param boolean $raw Optional. Whether or not we need to get the raw value. Default false.
 	 * @return mixed
+	 */
+	private function ___get( $key, $raw = false ) {
+
+		// force numeric id and prevent filtering on the id.
+		if ( 'id' === $key ) {
+
+			return $raw ? $this->$key : absint( $this->$key );
```
thomasplevy: I think that the ID should always be forced to an `int` I don't see any reason why we'd want a potentially non int id. Do you?

eri-trabiccolo: nope, just for consistency :D 
udating...

PR #896, opened 2019-07-05
Fix missing return value

pondermatic:
## Description
Methods that normally return a value should not return void.

## How has this been tested?
`composer phpcs`
`composer tests-run`

## Types of changes
Changed return type of methods that normally return a value but could have returned void. Instead of returning void, they now return false.

## Checklist:


includes/abstracts/abstract.llms.post.model.php
```diff
@@ -330,15 +330,16 @@ private function create( $title = '' ) {
...
+	 * @since    [version] Returns false if post is not cloneable.
+	 * @version  [version]
 	 */
 	public function clone_post() {
 
-		// if post type doesnt support cloning don't proceed
+		// if post type doesn't support cloning, don't proceed
 		if ( ! $this->is_cloneable() ) {
-			return;
+			return false;
```
thomasplevy: Is there some reason eluding me that we need to explicitly define `false` instead of just changing the documentation to reflect the way the code is actually working currently?

includes/models/model.llms.course.php
```diff
@@ -874,7 +874,9 @@ public function get_end_date() {
 
 	/**
 	 * Find the next uncompleted lesson in the course syllabus
-	 * @return int [Next uncompleted lesson]
+	 * @return int|false next uncompleted lesson id or false if all lessons are complete
+	 * @since [version] Returns false if all lessons are complete.
+	 * @version [version]
 	 */
 	public function get_next_uncompleted_lesson() {
```
thomasplevy: This (and a handful of other functions at the bottom of this class) should have been marked as deprecated but were "deprecated" before I became aware of docblocking as a way of life so I "deprecated" them by never adding inline docks and moving them to the bottom of the file. This was done back in 2016 and I've forgotten these functions exist.

includes/abstracts/abstract.llms.post.model.php
pondermatic: I consider it bad programming practice for a function that normally returns a value to also possibly return void. I read about this years ago, but can't seem to find any web pages about this topic right now. PHPStorm warns that a missing return value (in functions that can also return a value) may be a programmer mistake.

After more thought, I think I'll just change it to return null to keep this method's behavior the same and make PHPStorm happy. If someone wants to test for a uncloneable condition in the future, we could change it to return false, return WP_Error or throw an exception.

PR #900, opened 2019-07-12
LLMS_Post_Model: add a method to bulk update an llms_post with an arr…

eri-trabiccolo:
…ay of properties

## How has this been tested?
Tested with automated tests.

## Types of changes
New "internal" feature that should be backward compatible.

## Checklist:

### DIFF
--- a/includes/abstracts/abstract.llms.post.model.php
+++ b/includes/abstracts/abstract.llms.post.model.php
@@ -41,7 +41,7 @@ defined( 'ABSPATH' ) || exit;
  * @since 3.31.0 Treat `post_excerpt` fields as HTML instead of plain text.
  * @since [version] Add parameter to the `get()` method in order to get raw properties.
  * @since [version] Add `comment_status`, `ping_status`, `date_gmt`, `modified_gmt`, `menu_order` as gettable\settable post properties.
- * @since [version] Add `update_model()` method that will allow to update an object at once given an array of properties.
+ * @since [version] Add `set_bulk()` method that will allow to update an object at once given an array of properties.
  */
 abstract class LLMS_Post_Model implements JsonSerializable {
 
@@ -911,7 +911,7 @@ abstract class LLMS_Post_Model implements JsonSerializable {
 	 *
 	 * @since 3.0.0
 	 * @since 3.30.3 Use `wp_slash()` when setting properties.
-	 * @since [version] Turned to be only a wrapper for the update_model() method.
+	 * @since [version] Turned to be only a wrapper for the set_bulk() method.
 	 *
 	 * @param string|array $key_or_array Key of the property or a an associative array of key/val pairs.
 	 * @param mixed        $val          Optional. Value to set the property with. Default empty string.
@@ -928,23 +928,29 @@ abstract class LLMS_Post_Model implements JsonSerializable {
 		} else {
 			$model_array = $key_or_array;
 		}
-		return $this->update_model( $model_array );
+		return $this->set_bulk( $model_array );
 
 	}
 
 
 	/**
-	 * Updater
+	 * Bulk setter.
 	 *
 	 * @since [version]
 	 *
 	 * @param array $model_array Associative array of key/val pairs.
-	 * @return boolean true on success, false on error or if there was nothing to update.
+	 * @param array $wp_error    Optional. Whether or not return a WP_Error. Default false.
+	 * @return boolean|WP_Error True on success. If the param $wp_error is set to false this will be false on error or if there was nothing to update.
+	 *                          Otherwise this will be a WP_Error object collecting all the errors encountered along the way.
 	 */
-	public function update_model( $model_array ) {
+	public function set_bulk( $model_array, $wp_error = false ) {
 
 		if ( empty( $model_array ) ) {
-			return false;
+			if ( ! $wp_error ) {
+				return false;
+			} else {
+				return new WP_Error( 'empty_data', __( 'Empty data', 'lifterlms' ) );
+			}
 		}
 
 		$llms_post = array(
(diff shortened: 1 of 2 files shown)


includes/abstracts/abstract.llms.post.model.php
```diff
@@ -910,70 +911,130 @@ protected function scrub_field( $val, $type ) {
...
+		$post_properties = array_keys( $this->get_post_properties() );
+		$unsettable_properties = $this->get_unsettable_properties();
 
-				case 'title':
-					$val = apply_filters( 'title_save_pre', $val );
-				break;
+		foreach ( $model_array as $key => $val ) {
 
-			}
+			$val = $this->scrub( $key, $val );
```
thomasplevy: I'd like to see the content of this loop be broken out into a protected method solely responsible for filtering a single key when it's being updated.

That way the content of the loop can be reduced to something like:

```php
$val = $this->apply_update_filters( $this->scrub( $key, $val ), $key );
```

I think this should reduce the overall complexity and logic of this method a bit

Maybe I'm overcomplicating it though? Thoughts?

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -910,70 +911,130 @@ protected function scrub_field( $val, $type ) {
...
-			switch ( $key ) {
+	/**
+	 * Updater
+	 *
+	 * @since [version]
+	 *
+	 * @param array $model_array Associative array of key/val pairs.
+	 * @return boolean true on success, false on error or if there was nothing to update.
+	 */
+	public function update_model( $model_array ) {
```
thomasplevy: I think the method should be called `set_bulk()`

I also think that we should add an additional parameter `$wp_error` so we can grab **more information** in an error as an option.

This parameter should default to `false` and when `false` it will just return a simple boolean when an error / no update is required (exactly how the existing `set()` method works. Also allowing for backwards compat.

If `$wp_error' is `true` then we can create a `WP_Error` object and collect as many errors are encountered and then return that error object (instead of false).

The REST API will likely wish to use this to provide more robust information during an create/update request but existing methods which are likely just checking for a bool on set can safely continue working as is.

Do you think that solves the error problem you're asking about?

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -910,70 +911,130 @@ protected function scrub_field( $val, $type ) {
...
+				continue;
 			}
-		} // we have a problem...
-		else {
 
+			$llms_post[ $type ][ $llms_post_key ] = apply_filters( 'llms_set_' . $this->model_post_type . '_' . $key, $val, $this );
+
+		}// End foreach().
+
+		if ( empty( $llms_post['post'] ) && empty( $llms_post['meta'] ) ) {
```
thomasplevy: With `$wp_error` param addition:

```php
if ( $wp_error ) {
 return new WP_Error( ... );
} else {
return false;
}
```

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -910,70 +911,130 @@ protected function scrub_field( $val, $type ) {
...
 			return false;
+		}
+
+		$has_error = false;
```
thomasplevy: this can become a WP_Error object, collect the errors and later check if there are any errors to and return either the object OR `false` if `wp_error` is `false`

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -910,70 +911,130 @@ protected function scrub_field( $val, $type ) {
...
+				$u = update_post_meta( $this->id, $this->meta_prefix . $key, $val );
+				if ( is_numeric( $u ) || true === $u ) {
+					$has_error = $has_error || false;
+				} else {
+					$has_error = true;
+				}
+			}
 		}
 
+		return ! $has_error;
```
```php
if ( $wp_error_obj->has_error() ) {
   return $wp_error ? $wp_error : false;
}
return true

File: includes/abstracts/abstract.llms.post.model.php (hunk 2 above)
Comment (eri-trabiccolo) [reply includes/abstracts/abstract.llms.post.model.php]: Yes totally, thanks!

File: includes/abstracts/abstract.llms.post.model.php (hunk 1 above)
Comment (eri-trabiccolo) [reply includes/abstracts/abstract.llms.post.model.php]: Mmm sorry I don't get this :) 
...Or maybe yes because I though about that too, but basically I made a method that wrapped this:
https://github.com/gocodebox/lifterlms/blob/565950b04461d9aaca4f66852ea03de2ad8c2c10/includes/abstracts/abstract.llms.post.model.php#L960-L993
without the `continue` and ending with:
`return apply_filters( 'llms_set_' . $this->model_post_type . '_' . $key, $val, $this );`
in place of
`$llms_post[ $type ][ $llms_post_key ] = apply_filters( 'llms_set_' . $this->model_post_type . '_' . $key, $val, $this );`

But then I realized that in the `foreach` I still needed to perform the same checks:
```php
if ( in_array( $key, $post_properties ) ) {
...
} elseif ( ! in_array( $key, $unsettable_properties ) ) {
...
} else {
	continue;
}
```
because I needed to separate the `post_properties` from the `meta_properties`.
So I basically gave up because of that repetition.
But I can see how creating that method would lighten the reading/understanding of the code.
So if I got what you mean and you're fine with the "checks repetition" (or if you have a good way to avoid it that I'm totally missing) let me know and I'll do it!

includes/abstracts/abstract.llms.post.model.php
eri-trabiccolo: in 565950b04461d9aaca4f66852ea03de2ad8c2c10

includes/abstracts/abstract.llms.post.model.php
eri-trabiccolo: in 565950b

includes/abstracts/abstract.llms.post.model.php
eri-trabiccolo: in 565950b

includes/abstracts/abstract.llms.post.model.php
eri-trabiccolo: in 565950b

includes/abstracts/abstract.llms.post.model.php
eri-trabiccolo: `apply_update_filters` implemented in ea1221f3fa2ea8fc1f6ae3027772283cb73fb89a

What do you think?! Still a little bit too cryptic don't you think?

thomasplevy: @eri-trabiccolo man I'm sorry... Now that i'm seeing it this way you're right it was better the other way as there's too much redundancy now. Can you rollback to your previous version fo this function (without breaking out the `apply_update_filter()` method) and I'll have another look and merge it.

Sorry!

eri-trabiccolo: :D No problem at all. 
Latest commit reverted and force pushed.

eri-trabiccolo:
I know that slipping https://github.com/gocodebox/lifterlms/pull/900/commits/db501c36e588314048458a8f8da06de6fe82e576 into this PR is not really orthodox, so let me know if you want me to rollback.

PR #902, opened 2019-07-16
Add ability to filter the table CSS class names.

pondermatic:
## Description
Added the ability to change the array of table CSS class names. I needed a specific table to have left justified text using the `.llms-table.text-left td` selector.
Added `llms_table_get_table_classes` WordPress filter.

## How has this been tested?
* Tested `/wp-admin/admin.php?page=llms-reporting&tab=memberships` with and without this change to make sure the page did not change.
* Tested my plugin with a class that extends `LLMS_Admin_Table` and these methods:
```php
    protected function register_hooks() {
        add_filter( 'llms_table_get_table_classes', [ $this, 'add_table_classes' ], 10, 2 );
    }

    public function add_table_classes( $classes, $table_id ) {
        if ( $table_id !== $this->id ) {
            return $classes;
        }

        $classes[] = 'text-left';

        return $classes;
    }
```
* `composer tests-run`
* `composer phpcs`

## Types of changes

## Checklist:

### DIFF
--- a/includes/abstracts/abstract.llms.admin.table.php
+++ b/includes/abstracts/abstract.llms.admin.table.php
@@ -400,6 +400,28 @@ abstract class LLMS_Admin_Table extends LLMS_Abstract_Exportable_Admin_Table {
 	}
 
 	/**
+	 * Returns an array of CSS class names to use on this table.
+	 * @return array
+	 * @since   [version]
+	 * @version [version]
+	 */
+	protected function get_table_classes() {
+		$classes = [
+			'llms-table',
+			'llms-gb-table',
+			'llms-gb-table-' . $this->id,
+		];
+
+		/**
+		 * Filters the CSS classes to use on the table.
+		 * @param array $classes CSS class names
+		 * @param array $table_id id property of this table object
+		 * @since [version]
+		 */
+		return apply_filters( 'llms_table_get_table_classes', $classes, $this->id );
+	}
+
+	/**
 	 * Get HTML for the filters displayed in the head of the table
 	 * @return   string
 	 * @since    3.4.0
@@ -434,11 +456,7 @@ abstract class LLMS_Admin_Table extends LLMS_Abstract_Exportable_Admin_Table {
 	 */
 	public function get_table_html() {
 
-		$classes = array(
-			'llms-table',
-			'llms-gb-table',
-			'llms-gb-table-' . $this->id,
-		);
+		$classes = $this->get_table_classes();
 
 		if ( $this->is_zebra ) {
 			$classes[] = 'zebra';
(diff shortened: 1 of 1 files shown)


includes/abstracts/abstract.llms.admin.table.php
```diff
@@ -434,11 +456,7 @@ public function get_table_filters_html() {
...
 	public function get_table_html() {
 
-		$classes = array(
-			'llms-table',
-			'llms-gb-table',
-			'llms-gb-table-' . $this->id,
-		);
+		$classes = $this->get_table_classes();
 
 		if ( $this->is_zebra ) {
```
thomasplevy: @pondermatic I think these conditionally-added classes should probably get moved into the new `get_table_classes()` method so that they can be modified by the new filter in that method. Do you agree?

includes/abstracts/abstract.llms.admin.table.php
```diff
@@ -399,6 +399,28 @@ public function get_search() {
 		return esc_attr( trim( $this->search ) );
 	}
 
+	/**
+	 * Returns an array of CSS class names to use on this table.
+	 * @return array
+	 * @since   [version]
```
thomasplevy: Gonna nitpick our documentation here. I KNOW I KNOW that our existing codebase doesn't follow the new standards. I do not have interest or capacity to rewrite the docblock for every existing function in the codebase but I want all new code to follow our new standards and we're updating existing methods as we go.

The documented standards remove the `@version` tag for methods placing them only in the file header

In it's place is an additional `@since` tag recording the changes to the methods.

Finally, the `@since` tags are moved above the method's `@param` and `@return` tags.

See docs: https://github.com/gocodebox/lifterlms/blob/master/docs/documentation-standards.md#changelogs and example https://github.com/gocodebox/lifterlms/blob/master/docs/documentation-standards.md#functions-and-class-methods

I know this is hypocritical and nitpicky but I want new code to follow the standards. Thanks for understanding <3

thomasplevy:
Also failing tests should work now. Had some issues with debian recently changing to mariadb instead of mysql so all our builds started failing.

includes/abstracts/abstract.llms.admin.table.php
pondermatic: I understand completely and I was afraid when I submitted this that I forgot to check on somehting. :)

includes/abstracts/abstract.llms.admin.table.php
pondermatic: I don't understand. The classes **were** moved from `get_table_html()` to `get_table_classes()`.

thomasplevy: Pointer here was unclear, sorry. I'm referring to the classes that haven't been moved on lines 461 - 467 of the file after your new method is added.

```php
if ( $this->is_zebra ) {
   $classes[] = 'zebra`; 
}
etc..
```

pondermatic: I'm glad you caught that. Can't believe I missed it!

thomasplevy:
@pondermatic beautiful thank you <3

PR #904, opened 2019-07-18
Rename Gulp tasks

pondermatic:
## Description
Renamed gulp task files for Windows compatibility.


thomasplevy:
I think references to the files in the index need to be updated as well

For example:

https://github.com/gocodebox/lifterlms/blob/54a653cc61a152a3f6e0b7f766c177fadf48db71/gulpfile.js/index.js#L9

pondermatic:
I have updated all references to the three renamed files and can now successfully run all gulp tasks without error messages.

PR #906, opened 2019-07-23
LLMS_Post_Model->set_bulk() can accept post properties prefixed by "p…

eri-trabiccolo:
…ost_"

## Description
As of now we could create an LLMS_Post_Model with an array of the type 
```php
array(
    'post_title'   => 'A title',
    'post_content' => 'A content'
)
```
while to update the same model we should provide an array of the type
```php
array(
    'title'   => 'An updated title',
    'content' => 'An updated content'
)
```
This change will unify the way we feed the LLMS_Post_Model either when creating or updating it.

## How has this been tested?
existing unit tests only

## Types of changes
New feature

## Checklist:

### DIFF
--- a/includes/abstracts/abstract.llms.post.model.php
+++ b/includes/abstracts/abstract.llms.post.model.php
@@ -958,10 +958,17 @@ abstract class LLMS_Post_Model implements JsonSerializable {
 
 		foreach ( $model_array as $key => $val ) {
 
+			$is_post_property = in_array( $key, $post_properties );
+
+			// sanitize the post properties keys by removing the "post_" prefix.
+			if ( $is_post_property && "post_" === substr( $key, 0, 5 ) ) {
+				$key = substr( $key, 5 );
+			}
+
 			$val = $this->scrub( $key, $val );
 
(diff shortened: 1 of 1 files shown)


includes/abstracts/abstract.llms.post.model.php
```diff
@@ -958,10 +958,17 @@ public function set_bulk( $model_array, $wp_error = false ) {
 
 		foreach ( $model_array as $key => $val ) {
 
+			$is_post_property = in_array( $key, $post_properties );
+
+			// sanitize the post properties keys by removing the "post_" prefix.
+			if ( $is_post_property && "post_" === substr( $key, 0, 5 ) ) {
```
thomasplevy: single quotes instead of double (check coding standards via `composer run-script check-cs`

Fix and resubmit please

eri-trabiccolo: Yes sorry, I ran the check on the lifterlms-rest dir.

eri-trabiccolo: Fixed in 0a2b7d525cf0e4ce67a1fc74a6a8013309c519fb

PR #929, opened 2019-09-16
Return the author's display name as the default instructor's name

nicolas-jaussaud:
## Description
The get_instructors() method of the LLMS_Post_Instructors() is not returning the same result when there is no instructors set and it fallback to the post author.

When instructors are set, the method return for each instructor an array with the following indexes: label, visibility, id and name.

However, when there is no instructors set, the array of the instructor only contain 3 indexes: label, visibility and id. 

The objective of this merge is to return the name as well for this case.

## Checklist:
- [ ] My code passes all existing automated tests 
The code needs a modification in the automated tests (see in the commit)

### DIFF
--- a/includes/models/model.llms.post.instructors.php
+++ b/includes/models/model.llms.post.instructors.php
@@ -93,10 +93,12 @@ class LLMS_Post_Instructors {
 
 		// if empty, respond with the course author in an array
 		if ( ! $instructors ) {
+			$author = get_userdata( $this->post->get( 'author' ) );
 			$instructors = array(
 				wp_parse_args(
 					array(
-						'id' => $this->post->get( 'author' ),
+						'id' 	=> $this->post->get( 'author' ),
+						'name' 	=> $author ? $author->display_name : '',
 					),
 					llms_get_instructors_defaults()
 				),
--- a/includes/models/model.llms.post.instructors.php
+++ b/includes/models/model.llms.post.instructors.php
@@ -93,12 +93,12 @@ class LLMS_Post_Instructors {
 
 		// if empty, respond with the course author in an array
 		if ( ! $instructors ) {
-			$author = get_userdata( $this->post->get( 'author' ) );
+			$author      = get_userdata( $this->post->get( 'author' ) );
 			$instructors = array(
 				wp_parse_args(
 					array(
-						'id' 	=> $this->post->get( 'author' ),
-						'name' 	=> $author ? $author->display_name : '',
+						'id'   => $this->post->get( 'author' ),
+						'name' => $author ? $author->display_name : '',
 					),
 					llms_get_instructors_defaults()
 				),
--- a/tests/unit-tests/class-llms-test-post-instructors.php
+++ b/tests/unit-tests/class-llms-test-post-instructors.php
@@ -90,6 +90,8 @@ class LLMS_Test_Post_Instructors extends LLMS_UnitTestCase {
 			$post->set_instructors();
 			$expect = $defaults;
 			$expect['id'] = $user_ids[1];
+			$author = get_userdata( $user_ids[1] );
+			$expect['name'] = $author->display_name;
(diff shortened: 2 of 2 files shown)


thomasplevy:
@nicolas-jaussaud welcome to LifterLMS and thanks for the PR.

Could you help me out with the following before we merge:

A) Update your PR to fix the two coding standards issues reported by circle ci
B) Reply with a set of recreation steps so I can see the issue this fixes first hand. It seems like this is possibly a use case with some custom code (but I'm not sure at a glance) so if that's the case could you give me a quick snippet that I can run in a custom plugin or something that would help me see what's happening and how this fixes it.

Thanks <3

thomasplevy:
@actual-saurabh I believe you were mentioning something about this yesterday, correct? If so could you give me some reproduction steps here in writing so I can test this properly before merging.

@nicolas-jaussaud please have a look at the coding standards, it's a quick fix: `composer run-script fix-cs` Thanks

nicolas-jaussaud:
Sorry about the coding standard issue! I hope it will be fixed now.

It's indeed an issue with custom code, here is an example function (the commented foreach loop corresponds to the fix I'm currently using):

```php
function course_instuctors( $course_id ) {

  $course = new \LLMS_Course( $course_id );
  $instructors = $course->get_instructors( true );

  /**
   * When no instructor has been set, LifterLMS retrun the post author as the
   * default instructor, in that particular case, there is no name set.
   */
  // foreach( $instructors as $instructor ) {
  //   if( !isset($instructor['name']) ) {
  //     $author = get_userdata( $instructor['id'] );
  //     $instructor['name'] = $author->display_name;
  //   }
  // }

  return !empty($instructors) ? $instructors : array();
}

```

The issue happens when the $course_id correspond to a course where no instructor has been set by the user. In that case, the function returns the WordPress post author of the course as the instructor.

In that particular case, the get_instructors() method will return an array like this one:
```
array(1) { 
  [0]=> array(3) { 
    ["label"]=> string(6) "Author" 
    ["visibility"]=> string(7) "visible" 
    ["id"]=> int(1) 
  } 
} 
```

Instead of an array like this one (in all the other case):
```
array(1) { 
  [0]=> array(4) { 
    ["label"]=> string(6) "Author" 
    ["visibility"]=> string(7) "visible" 
    ["id"]=> int(1) 
    ["name"]=> string(5) "Admin" 
  } 
} 
```

thomasplevy:
@nicolas-jaussaud thanks for the information.

@actual-saurabh mentioned something like this to me the other day and I'm going to talk to him about the issue.

I'm wondering if it'd be better for us to back this up to see why the instructor isn't being properly set instead of patching it when it's not set.

I have a hunch that this might be happening at the moment of course creation (no instructor is stored). If that's the case I'd rather update it by actually storing the instructor at that moment instead of having a fallback for when it's not stored.

We may merge this in anyway to avoid running a database migration after deployment but I want to make sure we track down the bug first.

Thanks,

actual-saurabh:
I've described the related issue with #937. I feel that if we're expecting all instructor information to come from the metadata in one place, we should do so everywhere.

This basically means that all instructor information must be procured from the database and not return the author information by default.

Conversely, it could make more sense to include the author information as a fallback in case something fails when saving instructor information. In which case, we should accept this PR and change the Mine filter to use the author info as fallback to close #937

thomasplevy:
Related commit in the blocks plugin: https://github.com/gocodebox/lifterlms-blocks/commit/6bbc14105e7f4b0cbac853e87b280f126e464af3

With that PR we should be able to prevent posts from ending in a state that has the missing information that's fixed via this PR.

I'm hesitating to merge this now as I think we'd be better off writing a migration to fix the bugged posts instead of patching them at runtime.

Any thoughts?

thomasplevy:
@nicolas-jaussaud this is still pending but I accidentally closed this when renaming the master branch. If you could kindly resubmit against the "dev" branch we'll get this merged in.

Thanks

PR #936, opened 2019-09-18
Add SendWP Connector to Settings

thomasplevy:
## Description

SendWP is a transactional email subscription service for WordPress.

This PR adds a "Connect to SendWP" button on the LifterLMS Engagements Settings screens.

While SendWP is not a feature or LifterLMS product it will solve problems for users on servers with email deliverability issues. There are plenty of other transactional email services available but SendWP is very simple and we've entered into a partnership with SendWP. 

## How has this been tested?

+ New unit tests
+ Manual testing

## Screenshots

N/A

## Types of changes

New feature, no breaking changes.

## Checklist:

### DIFF
--- a/includes/admin/class-llms-sendwp.php
+++ b/includes/admin/class-llms-sendwp.php
@@ -167,6 +167,7 @@ class LLMS_SendWP {
 			break;
 		}
 
+		$install = null;
 		if ( ! $is_sendwp_installed ) {
 
 			include_once ABSPATH . 'wp-admin/includes/plugin-install.php';
@@ -218,7 +219,6 @@ class LLMS_SendWP {
 		foreach ( $settings as $i => $setting ) {
 			if ( 'email_options' === $setting['id'] && 'sectionend' === $setting['type'] ) {
 				return $i + 1;
-				break;
 			}
 		}
 		return $i;
--- a/includes/admin/class-llms-sendwp.php
+++ b/includes/admin/class-llms-sendwp.php
@@ -55,14 +55,14 @@ class LLMS_SendWP {
 	public function add_settings( $settings ) {
 
 		// Short circuit if missing unauthorized.
-		if ( ! current_user_can( 'install_plugins' )  ) {
+		if ( ! current_user_can( 'install_plugins' ) ) {
 			return $settings;
 		}
 
 		$new_settings = array(
 			array(
-				'id'    => 'sendwp_connect_start',
-				'type'  => 'sectionstart',
+				'id'   => 'sendwp_connect_start',
+				'type' => 'sectionstart',
 			),
 			array(
 				'id'    => 'sendwp_title',
@@ -71,7 +71,8 @@ class LLMS_SendWP {
 				'desc'  => sprintf(
 					// Translators: %1$s = Opening anchor tag; %2$s = Closing anchor tag.
 					__( '%1$sSendWP%2$s makes WordPress email delivery as simple as a few clicks so you can relax, knowing your important emails are being delivered on time.', 'lifterlms' ),
-					'<a href="https://lifterlikes.com/sendwp" target="_blank" rel="noopener noreferrer">', '</a>'
+					'<a href="https://lifterlikes.com/sendwp" target="_blank" rel="noopener noreferrer">',
+					'</a>'
 				),
 			),
 			array(
@@ -159,7 +160,7 @@ class LLMS_SendWP {
 				continue;
 			}
 			$is_sendwp_installed = true;
-			$activate = activate_plugin( $path );
+			$activate            = activate_plugin( $path );
 			if ( is_wp_error( $activate ) ) {
 				return $activate;
 			}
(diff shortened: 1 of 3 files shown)


includes/admin/class-llms-sendwp.php
```diff
@@ -0,0 +1,399 @@
...
+
+	/**
+	 * Install / Activate SendWP plugin.
+	 *
+	 * @since [version]
+	 *
+	 * @return WP_Error|true
+	 */
+	private function install() {
+
```
eri-trabiccolo: we could check on if `is_plugin_active( 'sendwp/sendwp.php' );` (requires: `include_once( ABSPATH . 'wp-admin/includes/plugin.php' );`) and skip all the rest in case the plugin is already active (but user not connected).

includes/admin/class-llms-sendwp.php
```diff
@@ -0,0 +1,399 @@
...
+			$activate = activate_plugin( $upgrader->plugin_info() );
+			if ( is_wp_error( $activate ) ) {
+				return $activate;
+			}
+		}
+
+		// Final check to see if SendWP is available.
+		if ( ! function_exists( 'sendwp_get_server_url' ) ) {
+			return new WP_Error( 'llms_sendwp_not_found', __( 'SendWP Plugin not found. Please try again.', 'lifterlms' ), $install );
```
eri-trabiccolo: if in the future `sendwp_get_server_url()` doesn't exist because sendwp has changed its API, but still the plugin is installed and active, the `$install` variable we pass as WP_Error data will be not defined: so for this edge case we should have a fallback if `empty($install)` or if `!isset($install)`, right?!

includes/admin/class-llms-sendwp.php
```diff
@@ -0,0 +1,399 @@
...
+			 * @since [version]
+			 *
+			 * @param {string} register_url Registration URL.
+			 * @param {string} client_name Client name.
+			 * @param {string} client_secret Client secret.
+			 * @param {string} client_redirect Client redirect URL.
+			 * @param {int} partner_id SendWP partner ID.
+			 * @return {void}
+			 */
+			function llms_sendwp_register_client( register_url, client_name, client_secret, client_redirect, partner_id ) {
```
eri-trabiccolo: never considered this as the easiest way to post+goto :D great!

includes/admin/class-llms-sendwp.php
```diff
@@ -0,0 +1,399 @@
...
+	 * @since [version]
+	 *
+	 * @param array $settings Default engagement settings.
+	 * @return int
+	 */
+	private function get_splice_index( $settings ) {
+		foreach ( $settings as $i => $setting ) {
+			if ( 'email_options' === $setting['id'] && 'sectionend' === $setting['type'] ) {
+				return $i + 1;
+				break;
```
eri-trabiccolo: if we `return` we don't need a `break`, right?

includes/admin/class-llms-sendwp.php
thomasplevy: SendWP devs from SaturdayDrive came up with that. Can't take credit :-(

includes/admin/class-llms-sendwp.php
thomasplevy: Yes we could but activating an already active via `activate_plugin()` responds with `null` the same as if the plugin was successfully activated from a deactivated state. So I think the current version does more or less the same thing without having to add another path to the function.

includes/admin/class-llms-sendwp.php
thomasplevy: Good catch, thanks

thomasplevy: Fixing this by setting `$install` as `null` before the condition so in this edgecase we'll just have `null` passed as the data to the WP_Error.

includes/admin/class-llms-sendwp.php
thomasplevy: no, no we don't.

includes/admin/class-llms-sendwp.php
eri-trabiccolo: Ok I see what you mean, my thought was to skip the 
`foreach ( get_plugins() as $path => $details )`
with `get_plugins` that might be expensive, although its result is cached I'm not sure whether it's already called during a typical admin request. But anyways that's fine :D

thomasplevy: Oh I see what you mean... Well, I don't feel like rewriting and retesting... whatever...

I did see if anyone has benchmarked `in_array()` vs `foreach()` and found this: https://gist.github.com/Maks3w/3097685

`is_plugin_active()` uses `in_array()` under the hood. According to this person `in_array()` would be *slightly* faster when doing loose comparisons in the loop. So I think performance wise we're arguing over nothing (but that's really the best kind of argument, right?

eri-trabiccolo: yeah we're arguing over nothing.
I was more talking about `get_plugins()` VS `is_plugin_active()` not about `in_array()` VS `foreach()`. But yes nevermind, if I find `get_plugins()` being slow I'll push a PR, but more likely I will not as this is not critical at all :D

PR #941, opened 2019-09-23
Make sure to actually process recurring charges only for existing ord…

eri-trabiccolo:
…ers and users.

## Description
Added some logic to avoid processing recurring charges for deleted orders, or for existing orders whose user has been deleted.
When the requested code for #940 will be implemented, the code of this PR would still make sense for edge cases like  orders/users manually removed from the db.

## How has this been tested?
existing and new unit tests. **haven't ran a manual test!)**

## Types of changes
Improvement/possible breaking change (but unit tests say it's fine)

## Checklist:
- [x] My code has been tested.
- [x] My code passes all existing automated tests.
- [x] My code follows the LifterLMS Coding Standards.

PR #947, opened 2019-09-24
Recurring charges improvements code refactor

thomasplevy:
Update to #941


thomasplevy:
@eri-trabiccolo I reorganized the code in #941 because the nested `if` statements were stressing me out. I don't know if this is actually any better but it feels a little more readable and as far as I can tell should function the same.

Can you have a quick look at it and double check I haven't broken anything here.

Thanks

PR #977, opened 2019-10-28
Don't remove `pre_get_posts` callback from within the callback itself.

### LINKED ISSUE #976: Conflict with CartFlows (which requires WooCommerce): visiting a CartFlows checkout page results in 404.
### Reproduction Steps

+ Install and activate WooCommerce and create a product;
+ Install and activate CartFlows (https://wordpress.org/plugins/cartflows/);
+ Create a flow;
+ Setup CartFlows Step Checkout (Woo) so to bind it to a WooCommerce product;
+ Visit the just created checkout page.

### Expected Behavior

I should see the checkout page I've just created.

### Actual Behavior

I'm brought to the 404 page.

### Error Messages / Logs

n/a

### System Information
<details>

```
--- THEME ---
Name: LifterLMS LaunchPad
Version: 2.4.6
Themeuri: https://lifterlms.com/launchpad/
Authoruri: https://gocodebox.com
Template:
Child Theme: No

--- PLUGINS ---
CartFlows: 1.3.0

LifterLMS: 3.36.3

WooCommerce: 3.7.1
  
```

</details>

### SOME OTHER INFO/THOUGHS

This issue has been introduced with:
https://github.com/gocodebox/lifterlms/commit/c7e7cc1d5e76323e9a92b945e822be6985f1a59a

That plugin runs its query alteration at priority 10.
Before the commit above, the CartFlows callback ran after our one, as you can image, now, it runs before. The issue, although, **doesn't regard our query alteration** as, in CartFlows checkout pages, none of our conditions to alter the query are matched.

I've actually found the main responsible of this issue to be this line:
https://github.com/gocodebox/lifterlms/blob/3.36.3/includes/class.llms.query.php#L238

My guess is that the callbacks re-sorting, which follows a callback removal, when called from within the callback to be removed (hence **within the same hook execution**) is the cause of the issue. Although I've been not able to track 

I guess the reason for that is to avoid altering the query twice?!
So instead of removing the callback we can use a static variable to be sure we perform the actual callback logic only once.

I'm going to propose a PR following this logic.

eri-trabiccolo:
fix #976

## Description
Rather use a static variable to make sure the business logic of the callback
is executed only once.

## How has this been tested?
Manually and existing unit tests.

## Types of changes
Plugin conflict fix

## Checklist:

### DIFF
--- a/includes/class.llms.query.php
+++ b/includes/class.llms.query.php
@@ -4,13 +4,13 @@
  * Handles queries and endpoints.
  *
  * @since 1.0.0
- * @version 3.36.3
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
 
 /**
- * LLMS_Query class
+ * LLMS_Query class.
  *
  * @since 1.0.0
  * @since 3.31.0 Deprecated `add_query_vars() method and added sanitizing functions when accessing `$_GET` vars.
@@ -18,6 +18,9 @@ defined( 'ABSPATH' ) || exit;
  * @since 3.36.3 Changed `pre_get_posts` callback from `10 (default) to `15`,
  *               so to avoid conflicts with the Divi theme whose callback runs at `10`,
  *               but since themes are loaded after plugins it overrode our one.
+ * @since [version] Don't remove `pre_get_posts` callback from within the callback itself.
+ *               Rather use a static variable to make sure the business logic of the callback
+ *               is executed only once.
  */
 class LLMS_Query {
 
@@ -177,13 +180,21 @@ class LLMS_Query {
 	 * @since 3.36.3 Changed `pre_get_posts` callback from `10 (default) to `15`,
 	 *               so to avoid conflicts with the Divi theme whose callback runs at `10`,
 	 *               but since themes are loaded after plugins it overrode our one.
+	 * @since [version] Don't remove this callback from within the callback itself.
+	 *               Rather use a static variable to make sure the business logic of this
+	 *               method is executed only once.
 	 *
 	 * @param WP_Query $query Main WP_Query Object.
 	 * @return void
 	 */
 	public function pre_get_posts( $query ) {
 
+		static $done      = false;
 		$modify_tax_query = false;
+
+		if ( $done ) {
+			return;
+		}
 
 		if ( ! is_admin() && $query->is_main_query() ) {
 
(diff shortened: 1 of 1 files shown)


thomasplevy:
@eri-trabiccolo 

I'm hesitant to make more changes here now that we've already created a new conflict by meddling.

I'd be more confident if the `LLMS_Query` class had significant test coverage but it doesn't. I spent a few hours trying to write tests but I'm stalled out trying to figure out how to cover modifications to the `$wp_query` global.

It does seem like this all works fine and if it solves a conflict cool but I'm still really hesitant.

If we can get some test coverage on the underlying function I'll be more confident. Maybe you can put a pair of eyes on how to get some coverage here because the road I've been going down just isn't working...

thomasplevy:
@eri-trabiccolo WIP on some tests (that are mostly not working) here: https://github.com/gocodebox/lifterlms/commit/d7f5d8bc911cf6798163c6349ad4bbddb8bad127

eri-trabiccolo:
Mmm not sure @thomasplevy 
I see what you mean, although I consider the change introduced here just a more "sane" way to accomplish what the current code aims to do (=> do it once), I can understand your perplexities on changing something we have little control on.
I wonder whether would be better to revert https://github.com/gocodebox/lifterlms/commit/73501c2a54d6b7c4334e55b3927cdc1020cd9869 until `LLMS_Query` class has significant test coverage. After all we've (I've) introduced a conflict to fix something that in theory can be an issue, but the only user who reported it once told about a workaround and applied it was able to "fix" the issue. So not ideal but acceptable.

I'm going to take a look at the query tests.

thomasplevy:
@eri-trabiccolo,

> I wonder whether would be better to revert 73501c2 until LLMS_Query class has significant test coverage.

I like this idea. Since we introduced this to fix a conflict with a specific theme and a theoretical but actually unknown number of other themes we could actually fix that conflict (without introducing any new conflicts) by adding a theme support class for Divi.

However, I'm not really a huge fan of adding theme support for any specific themes other than core themes (like the twenties). We have to draw a line somewhere.

Instead, we could add this to LifterLMS Labs as a Divi compatibility fix.

eri-trabiccolo:
Yes that makes sense, although we have to fix this anyways, adding this to LifterLMS Labs as Divi compat fix won't prevent avoiding the conflict with Divi + CartFlows :D 
Let's just revert it for the moment, what do you think?

PR #984, opened 2019-11-06
Make sure notices are printed on sales pages too.

### LINKED ISSUE #156: Analytics (Student and eComm)
Analytics are a broken mess, fix them

### LINKED ISSUE #983: Content restriction notices not shown on a "redirected" sales page.
### Reproduction Steps
(reported by a user who also provided a video)
+ set a course sales page redirection to a whatever wp page.
+ as visitor visit a lesson, which requires the above course enrollment, for which you have a direct link.
+ you get redirected to the course page, and then to the sales page you set in the first step.
**(first issue)**
+ repeat the second step above, say 2 times.
+ now visit a course page (that has no sales page redirection, of course), or the checkout page, whatever "page" we know that print notices, let's say we navigate to the account login page.
**(second issue)**

### Expected Behavior

**(first issue)**
+ I would expect to still see the restriction notice anyways.
But we could discuss on whether this actually makes sense.

**(second issue)**
+ I would expect nothing special :), e.g. I would expect to see the account login page normally.

### Actual Behavior

**(first issue)**
+ notices do not show up on the redirected page.

**(second issue)**
+ Two (equal to the number of the lesson access attempts) restriction notices are shown.

### Considerations 

This is due to the fact that those notices are not printed throughout the site but only in certain (llms) templates, hence while not shown on a "redirected sales page" if you navigate to a certain (llms) page like the login one (notices printed before the login form).

I'm not sure what would be the best solution here.
Does it make sense showing that kind of notice on a "redirected sales page"?
If so, maybe we can do something of the sort:
1) Pass a GET parameter here (`add_query_arg()`):
https://github.com/gocodebox/lifterlms/blob/3.36.4/includes/class.llms.template.loader.php#L97

2) Check the GET parameter here: 
https://github.com/gocodebox/lifterlms/blob/3.36.4/includes/class.llms.template.loader.php#L396
and if true, fire the `llms_print_notices` at `loop_start`

Otherwise I don't think it would make any sense showing these notices in other pages, maybe after you've navigated through 10 different pages on the website and you land on a "llms page"..., and we can just clear the notices (`llms_clear_notices()`)
here:
https://github.com/gocodebox/lifterlms/blob/3.36.4/includes/class.llms.template.loader.php#L95

### Error Messages / Logs

n/a

### System Information
LifterLMS 3.36.4
WordPress 5.2.4

n/a

eri-trabiccolo:
fixes #983
also fixes docs and cs

## How has this been tested?
manually (see reproduction steps in #983) and with existing unit tests

## Types of changes

## Checklist:

### DIFF
--- a/includes/class.llms.template.loader.php
+++ b/includes/class.llms.template.loader.php
@@ -1,26 +1,34 @@
 <?php
+/**
+ * Template loader.
+ *
+ * @since 1.0.0
+ * @version [version]
+ */
+
 defined( 'ABSPATH' ) || exit;
 
 /**
- * Template loader class
+ * Template loader class.
  *
- * @since    1.0.0
- * @version  3.20.0
+ * @since 1.0.0
+ * @since 3.20.0 Unknown.
+ * @since [version] Made sure notices are printed on sales pages too.
  */
 class LLMS_Template_Loader {
 
 	/**
 	 * Constructor
 	 *
-	 * @since    1.0.0
-	 * @version  3.20.0
+	 * @since 1.0.0
+	 * @since 3.20.0 Unknown.
 	 */
 	public function __construct() {
 
-		// do template loading
+		// do template loading.
 		add_filter( 'template_include', array( $this, 'template_loader' ) );
 
-		// restriction actions for each kind of restriction
+		// restriction actions for each kind of restriction.
 		$reasons = apply_filters(
 			'llms_restriction_reasons',
 			array(
@@ -45,13 +53,14 @@ class LLMS_Template_Loader {
 	}
 
 	/**
-	 * Add a notice and / or redirect during restriction actions
+	 * Add a notice and/or redirect during restriction actions
 	 *
-	 * @param    string $msg       notice message to display
-	 * @param    string $redirect  optional url to redirect to after setting a notice
-	 * @param    string $msg_type  type of message to display [notice|success|error|debug]
-	 * @since    3.0.0
-	 * @version  3.0.0
+	 * @since 3.0.0
+	 *
+	 * @param string $msg      Notice message to display.
+	 * @param string $redirect Optional. Url to redirect to after setting a notice. Default empty string.
+	 * @param string $msg_type Optional. Type of message to display [notice|success|error|debug]. Default 'notice'.
+	 * @return void
 	 */
 	private function handle_restriction( $msg = '', $redirect = '', $msg_type = 'notice' ) {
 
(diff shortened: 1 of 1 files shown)


includes/class.llms.template.loader.php
```diff
@@ -69,20 +78,21 @@ private function handle_restriction( $msg = '', $redirect = '', $msg_type = 'not
...
+	 * @since [version] Flag to print notices when landing on the redirected sales page.
+	 *
+	 * @return void
 	 */
 	public function maybe_redirect_to_sales_page() {
 
-		// only proceed for courses and memberships
-		if ( ! in_array( get_post_type(), array( 'course', 'llms_membership' ) ) ) {
+		// only proceed for courses and memberships.
+		if ( ! in_array( get_post_type(), array( 'course', 'llms_membership' ), true ) ) {
```
thomasplevy: A+

includes/class.llms.template.loader.php
```diff
@@ -437,23 +459,39 @@ public function template_loader( $template ) {
...
+	/**
+	 * Maybe print notices after redirection.
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	private function maybe_print_notices_on_sales_page_redirect() {
+
+		if ( ! empty( $_GET['llms_print_notices'] ) ) { // phpcs:ignore -- No need to verify nonce for this purpose.
```
thomasplevy: Use `llms_filter_input()` and you shouldn't run into nonce verification warnings.

And additionally when disabling phpcs on a single line it's preferable to disable phpcs for the specific rule instead of entirely. EG:

`// phpcs:disable WordPress.Security.NonceVerification.Missing`

This way if we happen to encounter other violations on the same line we can clean those up!

eri-trabiccolo: Of course you're right (on both).
I usually tend to disable `phpcs` for the single rule but this time I didn't, for no reason, I just did it wrong... sorry
I'm going to fix it tomorrow (I'm from my bed now 😅).

thomasplevy: No worries man!

eri-trabiccolo:
I'm re-testing this, look I don't like that `$_GET` variable appearing all the times in the address bar.
Because it'll appear also when you directly access to a course with a sales page set, even if there's no notice to print.
I want to add this variable only if there's at least one notice:
`count( LLMS()->session->get( 'llms_notices', array() ) ) > 1`

I don't think this will introduce an actual performance regression sine this check will only be performed in very specific cases.

I'm going to push this in the next commit.

includes/class.llms.template.loader.php
eri-trabiccolo: done in cfc8e6e

thomasplevy:
@eri-trabiccolo 

I like this a lot better, good call! I agree that it's not much of a performance impact to run this check and it's better to only add the notice check when we actually have a notice.

includes/class.llms.template.loader.php
```diff
@@ -93,8 +103,17 @@ public function maybe_redirect_to_sales_page() {
 			return;
 		}
 
+		$notices_to_print = LLMS()->session->get( 'llms_notices', array() );
```
thomasplevy: We can use `llms_notice_count()` instead.

Since we know we want to remove the reliance on the bundled session library using this function instead of checking the session data directly will make it easier to transition out in the future.

thomasplevy: https://developer.lifterlms.com/reference/functions/llms_notice_count/

eri-trabiccolo: I'm sorry, I totally forgot about this helper!

eri-trabiccolo: done in 3d1a8cc

PR #989, opened 2019-11-13
add filter, resolves #988

### LINKED ISSUE #988: Add a filter to allow customization of the frequency of `LLMS_Sessions::end_idle_sessions()`
**Code in question**

https://github.com/gocodebox/lifterlms/blob/914f43d5857348216022c3e838cfbe265410fd0d/includes/class-llms-sessions.php#L59-L62

Currently it's not possible to customize the frequency with this cleanup action is run on a background process.

Add a filter to the "every_five_mins" cron interval so that a custom interval could be used in its place.

Removing the action and re-adding it on a different hook but this is kind of wonky and may not actually work since the action is added via a class method instead using `array( $this, ... )`. I didn't setup a test case but I don't think it'll work.

thomasplevy:
## Description

Adds filter

Fixes #988 

## How has this been tested?

1. Passes existing unit tests
2. Manually tested with a filter applied:

A. Added code to custom plugin:
```
function my_llms_sessions_end_idle_cron_recurrence( $recurrence ) {
	return 'daily';
}
add_filter( 'llms_sessions_end_idle_cron_recurrence', 'my_llms_sessions_end_idle_cron_recurrence' );
```
B. Used WP Crontrol to delete the existing scheduled action on the 5 minute interval
C. Automatically rescheduled action moved to a once-per-day interval

## Types of changes

Adds a filter, backwards compatible and safe.

## Checklist:

### DIFF
--- a/includes/class-llms-sessions.php
+++ b/includes/class-llms-sessions.php
@@ -5,7 +5,7 @@
  * @package  LifterLMS/Classes
  *
  * @since 3.36.0
- * @version 3.36.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -14,6 +14,7 @@ defined( 'ABSPATH' ) || exit;
  * LLMS_Sessions class..
  *
  * @since 3.36.0
+ * @since [version] Add filter `llms_sessions_end_idle_cron_recurrence` to allow customization of the recurrence of the idle session cleanup cronjob.
  */
 class LLMS_Sessions {
 
@@ -49,6 +50,7 @@ class LLMS_Sessions {
 	 * Private Constructor.
 	 *
 	 * @since 3.36.0
+	 * @since [version] Add filter to the cleanup cronjob interval.
 	 *
 	 * @return void
 	 */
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
I'm going to review this tomorrow but, maybe, to improve the 'recurrence' filtering experience (avoiding 3rd party plugins to manage cron tasks or requiring extra coding) we could, instead of just comparing the existence of the scheduled event, do the following:
1) `$event = wp_get_scheduled_event( $hook )`
2) check the existence of the scheduled event
3) if it exists, also check that the recurrence ( $event->schedule) is different from the filtered recurrence.
3a) if 3) (different) reschedule the event with the new recurrence
4) if not 2) (no scheduled event) schedule the event with the filtered recurrence.

Well that's the idea that can be compacted of course.

What do you think?

thomasplevy:
@eri-trabiccolo I don't know if we really need to add additional checks (which will require additional database requests to be made on EVERY page load) in order to accommodate a very niche issue where someone has customized the default behavior of LifterLMS via a filter.

Sure the response is "You have to delete the existing cron in order to customize it" but I think it's more performant to do that ONE TIME than to programmatically do this on literally every page load...

thomasplevy:
Maybe there won't be additional database calls... I'd need to look at the actual calls made to know that for sure, that's my gut talking and right now I'm really tired and frustrated with myself...

thomasplevy:
@eri-trabiccolo if you think we should do it differently submit a pr tomorrow. I need to package up an emergency release and I was hoping to pull this in. This is a low priority and it can wait.

eri-trabiccolo:
Ok @thomasplevy sorry, maybe I did it too simple because `wp_next_scheduled()` internally calls `wp_get_scheduled_event()` so I didn't see any additional db call, but I didn't test the whole pseudo-code I theorized above, just gave a rapid look at the wp code.
If you want pack the current improvement in the emergency release and if my pseudo-code turned into a PR will be good as well it'll be shipped in the next release :).
... I know you're going to use this frustration to lift up like 1700 pounds (or something) shortly 😝

thomasplevy:
@eri-trabiccolo I should have read your idea more closely... Sorry I've been doing support *the entire day* and I skipped lunch so I'm in a foul mood.

Your idea is superior and makes no additioanl DB calls since `wp_next_scheduled()` actually uses `wp_get_scheduled_event()` under the hood.

thomasplevy:
Eh... I'm so torn about a small bit of code...

If we do it your way we then run the filter on every page load which again just doesn't feel necessary to me for something that most users are NEVER going to encounter (the customization)

Am I being dumb and hungry?

eri-trabiccolo:
Nope you're not dumb, that's what will happen. But I'm not aure why this particular filter is bugging you that much... I mean there are hundreds of filter hooks fired at each page load with no filter cbs attached, thie would be one of them. Actually, in my opinion, the possible useless overhead is in the case there's a filter cb attached, with my code this cb will be run at every page load, while with your code, only once.
Again I couldn't see it as a big deal, but you're still right, it's either not necessary.
Since we're in a "once in a blue moon" case, let's go with your solution. If more cases like this will arise, and I don't think (I hope) so, we'll improve it.

PR #998, opened 2019-11-26
Settings refactor

thomasplevy:
## Description

+ Adds tests for some untested area of the codebase (`LLMS_Settings_Page` and `LLMS_Certificates` classes)
+ Adds new utility methods to `LLMS_Settings_Page` to help reduce code redundancy across admin settings page classes
+ Adds logic used in the constructor of every core settings page class to the constructor of `LLMS_Settings_Page`
+ This additionally adds new actions to the `LLMS_Certificates` class as well as refactors some of the existing code. No functional changes have been made though.

## How has this been tested?

+ Manually tested everything to ensure changes don't break existing settings pages.
+ Added some new unit tests.

## Types of changes

This adds new methods in a backwards compatible way. The class `LLMS_Settings_Engagements` implements new methods but all the other settings page classes will continue to work as they previously did without any updates. Since these classes all define their own constructors the new constructor from the base class simply won't be used.

## Checklist:

### DIFF
--- a/includes/admin/settings/class.llms.settings.page.php
+++ b/includes/admin/settings/class.llms.settings.page.php
@@ -59,7 +59,7 @@ class LLMS_Settings_Page {
 	 */
 	public function __construct() {
 
-		$this->label = $this->get_label();
+		$this->label = $this->set_label();
 
 		add_filter( 'lifterlms_settings_tabs_array', array( $this, 'add_settings_page' ), $this->tab_priority );
 
@@ -123,7 +123,7 @@ class LLMS_Settings_Page {
 	 *
 	 * @return string
 	 */
-	public function get_label() {
+	protected function set_label() {
 		return $this->id;
 	}
 
--- a/includes/admin/settings/class.llms.settings.page.php
+++ b/includes/admin/settings/class.llms.settings.page.php
@@ -41,6 +41,15 @@ class LLMS_Settings_Page {
 	public $label = '';
 
 	/**
+	 * Tab priority
+	 *
+	 * Determines the order of the page when registered with the core settings array.
+	 *
+	 * @var int
+	 */
+	public $tab_priority = 20;
+
+	/**
 	 * Constructor.
 	 *
 	 * @since [version]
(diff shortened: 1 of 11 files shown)


thomasplevy:
@eri-trabiccolo can you have a quick once over on this when you get a chance.

If it all looks good we can update the remaining settings classes to utilize the functionality added here which will remove quite a bit of redundant code!

includes/admin/settings/class.llms.settings.page.php
```diff
@@ -78,59 +108,107 @@ public function flush_rewrite_rules() {
...
+	}
+
+	/**
+	 * Generates a group of settings.
+	 *
+	 * @since [version]
+	 *
+	 * @param string  $id Group ID. Used to create IDs for the start, end, and title fields.
+	 * @param string  $title Title of the group (should be translatable).
+	 * @param string  $title_desc (Optional) title field description text.
```
eri-trabiccolo: If we follow the wp doc standards, as I think we do (?), we should have:
```
* @param string  $title_desc Optional. Title field description text. Default empty.
* @param array[] $settings   Optional. Array of settings field arrays. Default empty array.
```

see: https://github.com/WordPress/WordPress/blob/5.3/wp-includes/rest-api/class-wp-rest-request.php#L108

includes/admin/settings/class.llms.settings.page.php
```diff
@@ -14,40 +14,70 @@
...
 	/**
-	 * Settings page link label
+	 * Settings page label / title.
 	 *
-	 * @var      string
-	 * @since    1.0.0
+	 * @var string
 	 */
-	public $label;
+	public $label = '';
```
eri-trabiccolo: do we need to set this to `''`?
My understanding is that the first thing we do is to set the `$label` property in the constructor so...

And now let me be a pain in the... I don't like the fact that the method `get_label()` doesn't actually return the `$label` property value, it's confusing...
I know that you created that method because you need to run the gettext function plus you wanted to make the override as small as possible, but let's use a different name, not sure... `build_label()` ? Maybe `protected` too?

What do you think? I know this is not the most important thing, but still I think it's better to change it now, if you agree.

thomasplevy: I don't think we have to set it but I like the idea of being able to say *this is literally always a string* which we can't if we don't set it. There is a world in which someone using the API decides they're not going to set it and I guess it's their problem then but I still think setting it as an empty string makes logical sense.

You're right about `get_label()` being wrong. I'm going to change it to `set_label()`

includes/admin/settings/class.llms.settings.page.php
```diff
@@ -57,15 +87,15 @@ public function add_settings_page( $pages ) {
...
 	public function flush_rewrite_rules() {
 
-		// add the updated endpoints
-		$q = new LLMS_Query();
-		$q->add_endpoints();
+		// Add the updated endpoints.
+		$query = new LLMS_Query();
+		$query->add_endpoints();
 
 		// flush rewrite rules
```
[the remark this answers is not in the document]
thomasplevy: yes.

includes/admin/settings/class.llms.settings.page.php
thomasplevy: This hasn't actually ever been explicitly defined in the coding standards and I haven't been doing it anymore since I rewrote them. Why? Because I haven't been able to find any way to automate this through PHPCBF or any other tools.

The WordPress core does not specify any amount of alignment either. If you have a look through the core codebase *sometimes* doc blocks are aligned and sometimes they're not. The same is true of LifterLMS.

I think they do *look* a whole lot nicer when aligned however without a way to automatically identify (and throw errors) as well as automatically fix it it feels like a large amount of work to enforce with little benefit as the alternative is still pretty easy to read as a human and the phpdoc parser that builds our developer.lifterlms.com site doesn't need any special alignment to parse.

If you can find a way to sniff and fix this I'd very much like to enforce this alignment but until then I'm not going to worry about it.

eri-trabiccolo: @thomasplevy I see your point, thanks.

PR #1004, opened 2019-12-12
Payments schedule

thomasplevy:
## Description

Refactors methodology for calculating the `date_next_payment` property on a recurring order.

We don't have an open issue for the problem this addresses because we have not been able to adequately come up with a set of recreation steps. However, the problem appears to be that after the *second* payment (the first recurring payment) the stored `date_next_payment` (the date of the transaction that was just processed) is a date in the *future*. As a result of this the calculation is treated like the initial payment calculation and calculated off of the order's `post_date` property.

This is problematic because it results in a new payment being scheduled in the past which is triggered again immediately following the transaction that was just processed.

In this PR we add a fallback condition which will "catch" this scenario and calculate the next payment off of the last transaction date.

## How has this been tested?

There are a ton of new integration tests which attempt to simulate the above issue as well as adding more robust test coverage for recurring payment calculations in "good" conditions.

I've have manually run recurring transactions from the frontend of the website (the way a customer/student would initiate a transaction). And then manually adjusted the next payment date to ensure that the recurring transaction fires as expected when triggered by the background processes.

I also have hundreds of recurring charges that run regularly on my local. I'll review my logs tomorrow to ensure they all continue to charge and reschedule as expected.

## Types of changes

## Checklist:

### DIFF
--- a/includes/models/model.llms.order.php
+++ b/includes/models/model.llms.order.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Models
  *
  * @since 3.0.0
- * @version 3.35.0
+ * @version [version]
  *
  * @property   $access_expiration  (string)  Expiration type [lifetime|limited-period|limited-date]
  * @property   $access_expires  (string)  Date access expires in m/d/Y format. Only applicable when $access_expiration is "limited-date"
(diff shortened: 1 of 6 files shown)


includes/models/model.llms.order.php
```diff
@@ -317,9 +337,18 @@ private function calculate_next_payment_date( $format = 'Y-m-d H:i:s' ) {
 		if ( 0 != $end_time && ( $next_payment_time + 23 * HOUR_IN_SECONDS ) > $end_time ) {
 			$ret = '';
 		} elseif ( $next_payment_time > 0 ) {
-			$ret = date_i18n( $format, $next_payment_time );
+			$ret = date( $format, $next_payment_time );
```
eri-trabiccolo: what's the reason for this change? Is this intended? Just a question.

eri-trabiccolo:
@thomasplevy about:

> However, the problem appears to be that after the second payment (the first recurring payment) the stored date_next_payment (the date of the transaction that was just processed) is a date in the future. As a result of this the calculation is treated like the initial payment calculation and calculated off of the order's post_date property.

Not sure about this, I mean the `date_next_payment` set in the future can justify transactions duplicated several times in a range of 10 minutes?
You know, looking at the code, a failure of `current_time()` seems to fit more, although not sure how that could happen.

Anyways the fall-back on the last transaction date as starting "time" for the next payment date should solve the issue, so we should be fine.

includes/models/model.llms.order.php
thomasplevy: My thinking is that this is an internal/private function that we request a specific format for so we actually don't want any modifications to that requested format adjusted by `date_i18n()`

eri-trabiccolo: I see, makes sense, thanks.

PR #1024, opened 2020-02-12
Sequential completion of lessons in a course.

tobeyadr:
## Description

This PR adds functionality to allow the user to define that all lessons must be completed sequentially.

With a course level setting, all lessons (except the first lesson) will have the lesson prerequisite set to the lesson preceding it.

## How has this been tested?

Tested on a localhost, latest version of WP and LifterLMS as well as test cases added to LLMS_Test_LLMS_Lesson class

## Types of changes

Current possible issues:
- In the course builder, the prerequisite is auto-set to the previous lesson. When saving, the lesson Id is saved to the prerequisite attribute. When disabling the course level settings, lessons will retain the prerequisite of the previous lesson ID. However, the pre-requisite checkbox will be disabled.

## Checklist:

### DIFF
--- a/includes/admin/post-types/meta-boxes/class.llms.meta.box.course.options.php
+++ b/includes/admin/post-types/meta-boxes/class.llms.meta.box.course.options.php
@@ -38,7 +38,7 @@ class LLMS_Meta_Box_Course_Options extends LLMS_Admin_Metabox {
 	 *
 	 * @since 1.0.0
 	 * @since 3.36.0 Allow some fields to store values with quotes.
-	 * @since 3.37.1 Added option to enable sequential completion
+	 * @since [version] Added option to enable sequential completion.
 	 *
 	 * @return array
 	 */
--- a/includes/admin/post-types/meta-boxes/class.llms.meta.box.course.options.php
+++ b/includes/admin/post-types/meta-boxes/class.llms.meta.box.course.options.php
@@ -14,6 +14,7 @@ defined( 'ABSPATH' ) || exit;
  * @since 1.0.0
  * @since 3.35.0 Verify nonces and sanitize `$_POST` data.
  * @since 3.36.0 Allow some fields to store values with quotes.
+ * @since 3.37.1 Added option to enable sequential completion
  */
 class LLMS_Meta_Box_Course_Options extends LLMS_Admin_Metabox {
 
@@ -37,6 +38,7 @@ class LLMS_Meta_Box_Course_Options extends LLMS_Admin_Metabox {
 	 *
 	 * @since 1.0.0
 	 * @since 3.36.0 Allow some fields to store values with quotes.
+	 * @since 3.37.1 Added option to enable sequential completion
 	 *
 	 * @return array
 	 */
@@ -350,7 +352,6 @@ class LLMS_Meta_Box_Course_Options extends LLMS_Admin_Metabox {
 						'label'            => __( 'Capacity Reached Message', 'lifterlms' ),
 						'type'             => 'text',
 					),
-					// Require sequential completetion
 					array(
 						'is_controller' => true,
 						'type'          => 'checkbox',
--- a/includes/admin/post-types/meta-boxes/class.llms.meta.box.course.options.php
+++ b/includes/admin/post-types/meta-boxes/class.llms.meta.box.course.options.php
@@ -14,7 +14,7 @@ defined( 'ABSPATH' ) || exit;
  * @since 1.0.0
  * @since 3.35.0 Verify nonces and sanitize `$_POST` data.
  * @since 3.36.0 Allow some fields to store values with quotes.
- * @since 3.37.1 Added option to enable sequential completion
+ * @since [version] Added option to enable sequential completion.
  */
 class LLMS_Meta_Box_Course_Options extends LLMS_Admin_Metabox {
 
--- a/includes/admin/post-types/meta-boxes/class.llms.meta.box.course.options.php
+++ b/includes/admin/post-types/meta-boxes/class.llms.meta.box.course.options.php
@@ -350,6 +350,17 @@ class LLMS_Meta_Box_Course_Options extends LLMS_Admin_Metabox {
 						'label'            => __( 'Capacity Reached Message', 'lifterlms' ),
 						'type'             => 'text',
 					),
+					// Require sequential completetion
+					array(
+						'is_controller' => true,
+						'type'          => 'checkbox',
+						'label'         => __( 'Enable Sequential Lesson Completion', 'lifterlms' ),
+						'desc'          => __( 'This will require that the lessons in the course be completed in order.', 'lifterlms' ),
+						'id'            => $this->prefix . 'complete_sequentially',
+						'class'         => '',
+						'value'         => 'yes',
+						'desc_class'    => 'd-3of4 t-3of4 m-1of2',
+					),
 				),
 			),
 		);
--- a/includes/functions/llms.functions.access.php
+++ b/includes/functions/llms.functions.access.php
@@ -353,8 +353,15 @@ function llms_is_post_restricted_by_prerequisite( $post_id, $user_id = null ) {
 		}
 	}
 
-	// otherwise return false
-	// no prereq
+	/**
+	 * Allow external plugins to specify pre_requisite restrictions
+	 *
+	 * @since 3.37.1
+	 *
+	 * @param bool whether the post is already restricted
+	 * @param int the ID of the post
+	 * @param int the ID of the current user
+	 */
 	return apply_filters( 'llms_is_post_restricted_by_prerequisite', false, $post_id, $user_id );
 
 }
--- a/includes/functions/llms.functions.access.php
+++ b/includes/functions/llms.functions.access.php
@@ -356,7 +356,7 @@ function llms_is_post_restricted_by_prerequisite( $post_id, $user_id = null ) {
 	/**
 	 * Allow external plugins to specify pre_requisite restrictions
 	 *
-	 * @since 3.37.1
+	 * @since [version]
 	 *
 	 * @param bool whether the post is already restricted
 	 * @param int the ID of the post
--- a/includes/functions/llms.functions.access.php
+++ b/includes/functions/llms.functions.access.php
@@ -355,7 +355,7 @@ function llms_is_post_restricted_by_prerequisite( $post_id, $user_id = null ) {
 
 	// otherwise return false
 	// no prereq
-	return false;
+	return apply_filters( 'llms_is_post_restricted_by_prerequisite', false, $post_id, $user_id );
 
 }
 
--- a/includes/models/model.llms.course.php
+++ b/includes/models/model.llms.course.php
@@ -773,7 +773,6 @@ class LLMS_Course
 	 * @since      Unknown
 	 * @return WP_Post[]
 	 * @deprecated 3.0.0
-	 *
 	 */
 	public function get_children_sections() {
 		llms_deprecated_function( 'LLMS_Course::get_children_sections()', '3.0.0', "LLMS_Course::get_sections( 'posts' )" );
@@ -787,7 +786,6 @@ class LLMS_Course
 	 * @since      Unknown
 	 * @return WP_Post[]
 	 * @deprecated 3.0.0
-	 *
 	 */
 	public function get_children_lessons() {
 		llms_deprecated_function( 'LLMS_Course::get_children_lessons()', '3.0.0', "LLMS_Course::get_lessons( 'posts' )" );
@@ -801,7 +799,6 @@ class LLMS_Course
 	 * @since      3.0.0
 	 * @return WP_User instance of WP_User
 	 * @deprecated 3.34.0
-	 *
 	 */
 	public function get_author() {
 		llms_deprecated_function( 'LLMS_Course::get_author()', '[version]' );
@@ -815,7 +812,6 @@ class LLMS_Course
 	 * @since      3.0.0
 	 * @return int
 	 * @deprecated 3.34.0
-	 *
 	 */
 	public function get_author_id() {
 		llms_deprecated_function( 'LLMS_Course::get_author_id()', '[version]', 'LLMS_Course::get( "author" )' );
@@ -829,7 +825,6 @@ class LLMS_Course
 	 * @since      3.0.0
 	 * @return string
 	 * @deprecated 3.34.0
-	 *
 	 */
 	public function get_author_name() {
 		llms_deprecated_function( 'LLMS_Course::get_author_name()', '[version]' );
@@ -844,7 +839,6 @@ class LLMS_Course
 	 * @since      Unknown
 	 * @return string
 	 * @deprecated 3.34.0
-	 *
 	 */
 	public function get_sku() {
 		llms_deprecated_function( 'LLMS_Course::get_sku()', '[version]', 'LLMS_Course::get( "sku" )' );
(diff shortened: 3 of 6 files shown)


thomasplevy:
@eri-trabiccolo I've spoken to @tobeyadr about this in Slack.

There's going to need to be some REST-related changes made to accommodate this enhancement.

The main change I see is that currently the affected lesson properties (`prerequisite` and `has_prerequisite` are mapped directly to database values. After the introduction of this code these properties will be conditionally "generated" based on the course-level setting.

For the REST API that means that create, update, and get requests to the lesson resource will all need to adjust their current behavior. I guess it means that the properties will be conditionally read-only?

EG: if the lesson's prereqs are managed by the course a API user should *not* be able to update the prerequisite via a POST request to the lesson.

Additionally the new property should be added to the course as well.

This is low priority but when you have a free moment I'd like your thoughts on the changes we'd need to make.

includes/admin/post-types/meta-boxes/class.llms.meta.box.course.options.php
```diff
@@ -350,6 +350,17 @@ public function get_fields() {
 						'label'            => __( 'Capacity Reached Message', 'lifterlms' ),
 						'type'             => 'text',
 					),
+					// Require sequential completetion
```
thomasplevy: This method & class needs a `@since` tag addition

includes/functions/llms.functions.access.php
```diff
@@ -355,7 +355,7 @@ function llms_is_post_restricted_by_prerequisite( $post_id, $user_id = null ) {
 
 	// otherwise return false
 	// no prereq
-	return false;
+	return apply_filters( 'llms_is_post_restricted_by_prerequisite', false, $post_id, $user_id );
```
thomasplevy: new filter requires filter docblock

function needs `@since` tag added for new update

includes/models/model.llms.course.php
```diff
@@ -548,6 +548,15 @@ public function has_prerequisite( $type = 'any' ) {
 
 	}
 
+	/**
+	 * Whether the course lessons must be completed sequentially
+	 *
```
thomasplevy: Needs `@since` tag.

Class needs `@since` and `@version` updates.

includes/models/model.llms.course.php
```diff
@@ -548,6 +548,15 @@ public function has_prerequisite( $type = 'any' ) {
...
+	 * @return bool
+	 */
+	public function must_complete_sequentially(){
```
thomasplevy: Need to add tests for this method in the course class

includes/models/model.llms.lesson.php
```diff
@@ -364,6 +364,33 @@ public function get_video() {
 		return $this->get_embed( 'video' );
 	}
 
+	/**
+	 * Get the lesson prerequisite
+	 *
```
thomasplevy: needs `@since` tag

includes/models/model.llms.lesson.php
```diff
@@ -364,6 +364,33 @@ public function get_video() {
...
+		}
+
+		// Good ol' fashioned way
+		if ( ! $pre_req ){
+			if ( $this->has_prerequisite ) {
+				$pre_req = $this->prerequisite;
+			}
+		}
+
+		return apply_filters( 'llms_lesson_get_prerequisite', $pre_req, $this );
```
thomasplevy: need filter doc block

includes/models/model.llms.lesson.php
```diff
@@ -373,7 +400,20 @@ public function get_video() {
...
+		// check course level setting
+		if ( $course && $course->must_complete_sequentially() ){
+			$has_pre_req = true;
+		}
+
+		if ( ! $has_pre_req ){
+			$has_pre_req = ( 'yes' == $this->get( 'has_prerequisite' ) && $this->get( 'prerequisite' ) );
+		}
+
+		return apply_filters( 'llms_lesson_has_prerequisite', $has_pre_req, $this );
```
thomasplevy: need filter docblock

includes/models/model.llms.lesson.php
```diff
@@ -373,7 +400,20 @@ public function get_video() {
 	 */
 	public function has_prerequisite() {
```
thomasplevy: method needs `@since` tag updates

includes/models/model.llms.lesson.php
```diff
@@ -522,17 +561,16 @@ public function toArrayAfter( $arr ) {
 			}
 		}
 
+		// handling in the event of course level prerequisite settings
```
thomasplevy: method needs `@since` tag updates

includes/models/model.llms.lesson.php
```diff
@@ -364,6 +364,33 @@ public function get_video() {
...
+		$course = $this->get_course();
+
+		// Check course level setting
+		if ( $course && $course->must_complete_sequentially() ){
+			// might be false so we are okay
+			$pre_req = $this->get_previous_lesson();
+		}
+
+		// Good ol' fashioned way
+		if ( ! $pre_req ){
```
thomasplevy: this condition seems unnecessary to me and would be more concisely written as

```php
if ( $course && $course->must_complete_sequentially() ) [
 ...
} elseif ( $this->has_prerequisite ) {
 ...
}

File: includes/models/model.llms.lesson.php [hunk 11]
```diff
@@ -364,6 +364,33 @@ public function get_video() {
...
+			if ( $this->has_prerequisite ) {
```
thomasplevy: this could be a pre-existing bug that's gone unnoticed / reported: this line should be checking the method `has_prerequisite()` instead of the property `has_prerequisite()`

Same for the `prerequisite` on the next line.

tests/phpunit/unit-tests/models/class-llms-test-model-llms-lesson.php
```diff
@@ -417,4 +417,46 @@ public function test_is_orphan() {
 
 	}
 
+	/**
+	 * Test the new functionality for the course level setting for prerequiesites
+	 */
+	public function test_course_level_prerequisite()
```
thomasplevy: this particular test is (mostly) okay. It would be better to move this test to the course tests class.

And then add tests for the two lesson methods that have been affected by the pr.


Despite what a lot of our existing tests put forth as a "style" we've learned that making smaller more "atomic" tests is better and more maintainable over time.

Therefore the lesson tests class should have a few new tests to test the various of each method. EG: don't write one test to test all the functions in one test function, instead write multiple test functions to test each potential condition of each new function.

For example `has_prerequisite()` could have multilpe tests:

test_has_prerequisite_course_sequential -- tests the return when the course has sequential enabled
test_has_prerequisite_on -- no course prereq but does have lesson enabled
test_has_prereq_off -- no course && no lesson

thomasplevy (changes requested):
CS issues identified by code climate can be fixed manually if you don't want to install dev deps or install dev deps and run `composer run-script fix-cs`

There's some documentation standards issue to resolve to, which cannot be automated. Yes, these are nitpicky and yes our existing codebase is in a transitional phase where it often doesn't follow it's own documentation standards. We're transitioning files as we update them and new code should always adhere to the doc standards: https://github.com/gocodebox/lifterlms/blob/master/docs/documentation-standards.md

includes/functions/llms.functions.access.php
tobeyadr: For the @since, do you want 3.37.1 or the current version 3.37.0?

includes/admin/post-types/meta-boxes/class.llms.meta.box.course.options.php
tobeyadr: Done

includes/functions/llms.functions.access.php
tobeyadr: Done

eri-trabiccolo: you can use `@since [version]`
then `[version]` will be replaced by the release version when building the new release.

includes/models/model.llms.course.php
tobeyadr: The use of @version seems a bit ambiguous, can you define how you want it used?

includes/models/model.llms.course.php
tobeyadr: Done

includes/models/model.llms.lesson.php
tobeyadr: Since we are in the `get_prerequiste()` here I don't think it should use the `has_prerequisite()` method. 

The `has_prerequiste()` method will return true if the course level setting is available OR the lesson has a manually set prerequisite from the course builder.

This means that if we were to use `has_prerequiste()` here we do not know if it's the sequential completion setting or course builder setting.

Also, can't use `get_prerequiste()` that would put us into a recursion.

Does that make sense?

includes/models/model.llms.lesson.php
tobeyadr: Done

includes/models/model.llms.lesson.php
tobeyadr: Done

includes/models/model.llms.lesson.php
tobeyadr: Done

includes/models/model.llms.course.php
tobeyadr: NVM

includes/models/model.llms.lesson.php
tobeyadr: Done

includes/models/model.llms.lesson.php
tobeyadr: Done

includes/models/model.llms.lesson.php
tobeyadr: I agree

includes/models/model.llms.lesson.php
tobeyadr: Resolved based on your other comment.

includes/models/model.llms.lesson.php
tobeyadr: I've made a similar change to the `has_prerequisite()` method as well

tests/phpunit/unit-tests/models/class-llms-test-model-llms-lesson.php
tobeyadr: Done & Added new test cases

eri-trabiccolo:
Some thoughts about this.

The following two are really minor and I just wanted to mention them, since I spent some sugar to think about them :P 
1) I think we should add a customizable message that will be shown after the redirect to inform the user what's happening, exactly like the other "restriction" options.
I know this is a kind of an edge case, but still someone can have a direct link to a lesson which requires the previous 4 lessons to be completed, being redirected there and don't know what's happening...

2) Even if we're talking about "sequential completion", actually we are implementing a different concept, which is a probably acceptable compromise :D 
I mean, here we're setting as prerequisite that the previous lesson is completed, while actually we should make sure that ALL the previous lessons are completed (or free). Once again this is an edge case, with, as far as I can see, practically 0 real consequences if not this:
I have completed only lesson 1,
I have the link for the lesson 10, I put it in the address bar, I'm redirected to lesson 9, not completed, so redirected to 8, not completed, so redirected to... until lesson 2.
------
I'm then not sure about this piece:

https://github.com/gocodebox/lifterlms/pull/1024/files#diff-50f370269662ef9d5a22820e422278a6R583-R587

This is "conceptually" tricky to me, also considering the point 2 above.
Why when I access the `$instance[$prop]` as array should I see `has_prerequisite` and `prerequisite` in a way which is different from when I access the model as `$instance->get($prop)`? I don't think that's right, I should get the same value, and, to me, it should be the db - sanitized yes - value.
Also, the `toArray()` method is used on exporting + cloning right? This means that those values will go in the cloned (or imported) lesson's metas (db), so they must be the the db values.
If I clone a course 1 - lesson 3 which has no prerequisite meta but the course requires sequential completion, I don't want the cloned lesson 3 to have has prerequisite course 1 - lesson 2. It makes no sense, right? But that's what is going to happen.

-----

About REST API @thomasplevy your solution sounds good to me, although... I don't know, I'm not fully convinced, because, again point 2 above, that's not the "data value" of the property of that resource. If I'm a simple reader that's fine, but at least in edit context I might want to know the real value there for any reason, without having to change a property in another resource, then get the actual value of the lesson property, then change back the parent course property... you know.

What if we turn the `prerequisite` into an object, with the following subprops?
- lesson_id : int - The `prerequisite` meta value
- requires_sequential_completion (or a smaller one) : readonly - bool - the parent course option value

Or something of the sort, we might even have 3 subprops: db value, sequential completion, actual prerequisite id (result of `get_prerequisite()` ). The latter 2 must be readonly.

In any case we need to document that 

Then what will actually change should be the `prerequisite` **link** (under __links_), which will be the link to the lesson returned by `get_prerequisite()`.

tobeyadr:
A possible solution to your second point @eri-trabiccolo.

When redirecting to the pre-requisite lesson, would it be possible to check if the pre-requisite has a pre-requisite itself and see if it's complete before redirecting?

You can use a recursive function to check each lesson with the exit case being the redirect to the first available lesson.

This way, 1 redirect instead of 8.

eri-trabiccolo:
yeah @tobeyadr but actually even that won't be 100% precise.
The first available lesson should be the first in the sequence (ASC) that has no prerequisite.
Point 2) is too much an edge case, I just needed to share it because I though about it and maybe it helped some speculation :D  
I think your solution (previous lesson's completion) is good enough.

includes/models/model.llms.lesson.php
```diff
@@ -364,16 +364,73 @@ public function get_video() {
...
+		return apply_filters( 'llms_lesson_get_prerequisite', $pre_req, $this );
+	}
+
 	/**
 	 * Determine if lesson prereq is enabled and a prereq lesson is selected
 	 *
 	 * @return   boolean
 	 * @since    3.0.0
+	 * @cince    3.37.1 Add filter and compatibility for course level prerequisite setting.
 	 * @version  3.0.0
```
eri-trabiccolo:
```suggestion
	 * @since 3.0.0
	 * @since [version] Add filter and compatibility for course level prerequisite setting.
	 *
	 * @return bool
```
and so on...

eri-trabiccolo (changes requested):
I've requested some changes regarding to docs standards. :)
Except for that and the thing about `toArray` I would like to have Thomas' feedback on, and maybe the message addition (for both see https://github.com/gocodebox/lifterlms/pull/1024#issuecomment-586380115), it looks good to me.

tobeyadr:
Ok, I have refactored the PHP docs. I use PHPStorm, so I calibrated my IDE to observe your docs standards. I took the liberty of auto-refactoring the whole llms_lesson and llms_course class to meet your doc standards.

thomasplevy:
I'll have a look at this again either Friday or Monday. Sorry for the delays.

eri-trabiccolo (commented):
Hello Adrian,
from the functional point of view, my main doubt is about the `toArray()` thing as said in
let's wait for Thomas' response :)
https://github.com/gocodebox/lifterlms/pull/1024#pullrequestreview-359111969
The impact this PR will have on the rest api is something that we can handle one way or another.

About the docs, there would still be things to fix, but honestly we know that respecting the doc standards, that are stated [here](https://github.com/gocodebox/lifterlms/blob/master/docs/documentation-standards.md#), but that are not respected in most of the codebase, since we're in a transitional state, can be hard, cannot ask you to spend some other time on that. But if you want, let me know and I will review the doc standards compliance in depth.

Thanks again!

thomasplevy:
@tobeyadr 

We're switching the names of our branches and I neglected to switch the base branch before deleting master (that no longer exists).

Would you kindly reopen this against the "dev" branch when you have some time.

I know this has been hanging for an awfully long time and we will (eventually) get it merged in we just haven't had capacity to prioritize this.

Thanks (and sorry!)

Issue #1044, opened 2020-03-05
Feature Addition: Native Akismet Spam Integration


thomasplevy:
Mostly functional WIP at https://github.com/gocodebox/lifterlms/tree/feature/akismet

Purpose:

Send user registration data through the Akismet API prior to user signup during checkout and/or open registration.

Todo:

#### Bugs

- [ ] Bug: integrations list shows the integration as enabled regardless of whether or not it is

#### Moderation

Akismet recommends having a "moderation" queue to allow admins to review "iffy" signups. There's not "moderation" status for WordPress users so with these "iffy" signups we should probably just register the user and then add a spam flag to the usermeta table.

- [ ] Admins should be notified
- [ ] the WP User Profile (admin panel) add an area to submit the user as spam via the "Submit Spam" api endpoint: https://akismet.com/development/api/#submit-spam (this should be enabled for any user always. Even if a user doesn't come in through LifterLMS users could still submit them to Akismet via the API. Historical spam can be handled in this way too.
- [ ] The users table could output spam status information like the existing comments moderation links

#### Ham

Akismet recommends allowing users to be able to submit false-positives (ham) back to Akisment: https://akismet.com/development/api/#submit-ham

- [ ] Using same interface cues as above admins should be able to mark things identified as spam as not spam.

#### Tests

- [ ] php unit and integration tests should be written

PR #1045, opened 2020-03-05
Early review for Akismet Feature Addition

thomasplevy:
For code review, per #1044 

Low priority.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ dist/
 tmp/
 wordpress/
 build/
+
+phpunit.xml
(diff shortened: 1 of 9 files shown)


eri-trabiccolo:
We should consider leveraging the wp api to add "suggestions" on updating the privacy policy page, when this feature is enabled.

eri-trabiccolo:
why did you close this?
I'm sorry for being late but I didn't think I was thaaat late...

thomasplevy:
@eri-trabiccolo changing priorities. It didn't work for the test user (at all) to block any amount of spam

eri-trabiccolo:
well yeah I was surprised, it's kinda impossible to detect a spam if you cannot check the presence of "enlarger" anywhere, unless Enlarger is your last name, and we all know what would be the first...

PR #1055, opened 2020-03-10
fix comments better

thomasplevy:
## Description

Fixes comment related issues introduced in (unreleased) #1054 

New issues would result in our methods overwriting comment counts on the same filter (like WooCommerce)

Updated methodology reverts to a subtraction method but also introduces a new case to handle scenarios where the incoming stats is an empty array (because no other plugins have modified the object yet).

In this scenario we leverage `get_comment_count()` to let the WordPress core calculate the comment counts and then we subtract our comments the same way as we used to.

## How has this been tested?

Unit tests

Manually tested this against WooCommerce to ensure that WooCommerce comment types are still subtracted (by Woo's methods)

Manually tested without WooCommerce.

Manually tested with WPForms Lite which (which isn't actually creating an issue but surfaces a conflict with the 3.x version of the Action Scheduler library)

## Types of changes

Preemptive plugin conflict fix

## Checklist:


includes/class.llms.comments.php
```diff
@@ -126,79 +134,149 @@ public static function exclude_order_comments_from_feed_where( $where ) {
...
+	 * Remove order notes from an existing comment count stats object.
+	 *
+	 * This method accepts a stats object, generated by another plugin (like WooCommerce) or using core information
+	 * from `get_comment_counts()` and then subtracts LifterLMS order note comments from the existing comment counts
+	 * which would have included order notes in the counts.
+	 *
+	 * @since [version]
+	 *
+	 * @param  stdClass $stats Comment stats object. See the return of LLMS_Comments::wp_comment_counts() for object details.
+	 * @return stdClass        See LLMS_Comments::wp_comment_counts() for return object details.
```
eri-trabiccolo:
```suggestion
	 * @return stdClass See LLMS_Comments::wp_comment_counts() for return object details.
```

PR #1063, opened 2020-03-13
Persist the tracking events via ajax when reaching the cookie size li…

eri-trabiccolo:
…mit.

@thomasplevy 

This is a WIP.
I need to add unit tests (functional mostly) and wanted to finish fixing the cs in `class.llms.ajax.handler.php`
I'm not going to finish it today.

Just wanted to have your opinion on the approach - IF AND WHEN you got time -, which I think is basically what we discussed except for the condition that "triggers" the ajax action.

Fixes gocodebox/lifterlms-advanced-videos#23

Bye

### DIFF
--- a/assets/js/app/llms-tracking.js
+++ b/assets/js/app/llms-tracking.js
@@ -7,6 +7,7 @@
  * @since 3.36.2 Fix JS error when settings aren't loaded.
  * @since 3.37.2 When adding an event to the storae also make sure the nonce is set for server-side verification.
  * @since 3.37.9 Fix IE compatibility issue related to usage of `Object.assign()`.
+ * @since [version] Persist the tracking events via ajax when reaching the cookie size limit.
  */
 LLMS.Tracking = function( settings ) {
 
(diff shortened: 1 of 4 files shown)


thomasplevy:
@eri-trabiccolo I didn't do a full review but I read through it and it looks like a solid approach to me. I don't see any obvious issues.

I'll do a more thorough review early next week, good work so far

eri-trabiccolo:
Thanks @thomasplevy
Hopefully early next week there will be a final version of this for you to review.

thomasplevy:
@eri-trabiccolo this looks good but I haven't actually tested it yet and I want to run some manual tests before merging. I'm going to do that tomorrow.

thomasplevy:
@eri-trabiccolo I'm seeing an issue here... 

I'm adding fake events to the tracking data by repeatedly running this in the console:

`llms.tracking.addEvent( 'page.focus' )`

Watching the cookie in the sources I can watch the cookie size increase in the Application dev tools tab.

It looks like when the cookie size reaches it max it clears but the AJAX event is not triggered and (I think) the events are just lost.

thomasplevy:
I think I was being dumb...

thomasplevy:
Yep, works perfect... I had some text in the filter box on the XHR tab and therefore didn't see the request going through. This is great <3

eri-trabiccolo:
@thomasplevy ah well that's THE thing that should work... also the cleaning and the ajax call are under the same condition so... I'm going to check in 2 minutes.

PR #1066, opened 2020-03-18
Debug log

thomasplevy:
## Description

+ Hardened admin "tools" by requiring the `manage_lifterlms` cap in addition to nonce verifications
+ Use strict comparisons where possible
+ Added the ability to view (but not delete) the WP core debug.log file (if it exists)

I wrote this because @nrherron92 was asking if there's an easier way to get the debug.log file from a non-technical user who's having an issue. This is really a feature to help us do better support than anything else. If we need to review logs (which we frequently do) we can just review them instead of trying to explain to somehow how to find an error log on their server.

## How has this been tested?

+ Manually tested
+ Added some new unit tests to test related functionality

## Types of changes

Feature enhancements
THe "hardening" of user capabilities won't affect any users since that capability is already required to view the page. This would harden against someone trying to exploit the code through a custom plugin that calls our methods.

## Checklist:

### DIFF
--- a/includes/admin/class.llms.admin.page.status.php
+++ b/includes/admin/class.llms.admin.page.status.php
@@ -27,6 +27,7 @@ class LLMS_Admin_Page_Status {
 	 *
 	 * @since 3.11.2
 	 * @since 3.35.0 Sanitize input data.
+	 * @since [version] Use `llms_redirect_and_exit()` in favor of `wp_safe_redirect()`.
 	 *
 	 * @return void
 	 */
@@ -39,7 +40,11 @@ class LLMS_Admin_Page_Status {
 		$tool = llms_filter_input( INPUT_POST, 'llms_tool', FILTER_SANITIZE_STRING );
 
 		/**
-		 * Custom Tools can hook into this action to do the tool action
+		 * Custom and 3rd party tools can use this action to perform the tool's action
+		 *
+		 * @since Unknown
+		 *
+		 * @param string $tool Tool name or ID.
 		 */
 		do_action( 'llms_status_tool', $tool );
 
@@ -52,7 +57,7 @@ class LLMS_Admin_Page_Status {
 
 			case 'clear-cache':
 				global $wpdb;
-				$wpdb->query(
+				$wpdb->query( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
 					$wpdb->prepare(
 						"DELETE FROM {$wpdb->prefix}usermeta WHERE meta_key = %s or meta_key = %s;",
 						'llms_overall_progress',
@@ -71,9 +76,8 @@ class LLMS_Admin_Page_Status {
 				break;
 
 			case 'setup-wizard':
-				wp_safe_redirect( esc_url_raw( admin_url( '?page=llms-setup' ) ) );
-				exit;
-			break;
+				llms_redirect_and_exit( esc_url_raw( admin_url( '?page=llms-setup' ) ) );
+				break;
 
 		}
 	}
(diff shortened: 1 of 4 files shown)


thomasplevy:
@eri-trabiccolo if you have some time could you have a quick glance through the changes here and let me know if you see any stupid mistakes or issues!

includes/admin/views/status/view-log.php
```diff
@@ -0,0 +1,51 @@
...
+// Nonce URL to delete a log file.
+$delete_url = 'debug-log' === $current ? '' : wp_nonce_url(
+	add_query_arg(
+		array(
+			'llms_delete_log' => $current,
+		),
+		admin_url( 'admin.php?page=llms-status&tab=logs' )
+	),
+	'delete_log'
+);
```
eri-trabiccolo: 1) I think this part can be moved into includes/admin/class.llms.admin.page.status.php 
2) the condition here is a little bit weak. The "wp" debug log file can be named in other ways, that's why we get its actual name via `ini_get()`. Right? So I guess that here we should get the current file name (`$logs[$current]`) and compare it with `ini_get( 'error_log' )`. What do you think?

includes/admin/views/status/view-log.php
```diff
@@ -0,0 +1,51 @@
...
+				<?php echo esc_html( basename( $file ) ); ?>
+				(<?php echo date_i18n( $date_format, filemtime( $file ) ); ?>)
+			</option>
+		<?php endforeach; ?>
+	</select>
+	<button class="llms-button-secondary small" type="submit"><?php _e( 'View Log', 'lifterlms' ); ?></button>
+</form>
+
+<h2>
+	<?php printf( esc_html__( 'Viewing: %s', 'lifterlms' ), basename( $logs[ $current ] ) ); ?>
```
eri-trabiccolo: Missing `// Translators...`

tests/phpunit/unit-tests/admin/class-llms-test-admin-page-status.php
```diff
@@ -0,0 +1,241 @@
...
+/**
+ * Test Admin Status page
+ *
+ * @package LifterLMS/Tests/Admin
+ *
+ * @group admin
+ * @group status
+ *
+ * @since [version]
+ */
```
eri-trabiccolo: not totally sure because we don't use a cs for the tests, but I guess we also use to add the `@version` tag, right?

thomasplevy: @eri-trabiccolo I've stopped adding `@version` to test files since it's redundant (eg: the test files dont' have a fileheader and a class definition). I should probably document it but I really don't care.

includes/admin/class.llms.admin.page.status.php
```diff
@@ -115,23 +117,38 @@ public static function get_url( $tab = null ) {
...
+					continue;
 				}
+
+				$result[ sanitize_title( $value ) ] = LLMS_LOG_DIR . $value;
+
 			}
 		}
+
+		// Add the site's debug.log file if it exists.
+		if ( defined( 'WP_DEBUG_LOG' ) && WP_DEBUG_LOG ) {
```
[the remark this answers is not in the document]
thomasplevy: potentially true... but I want to output the log file if it exists, not if it's explicitly enabled currently. That makes sense, right?

includes/admin/class.llms.admin.page.status.php
```diff
@@ -115,23 +117,38 @@ public static function get_url( $tab = null ) {
...
+			$path = ini_get( 'error_log' );
+			if ( $path ) {
+				$result['debug-log'] = ini_get( 'error_log' );
+			}
```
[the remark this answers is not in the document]
thomasplevy: Smart yes thanks.

includes/admin/views/status/view-log.php
thomasplevy: 1. I guess lol
2. This is referencing the "key" that is given to the log in the array of available logs which is hardcoded as "debug-log" in the `get_logs()` method so regardless of what the file is actually named the "slug" we give to that file will always be "debug-log" so this condition will always work.

thomasplevy: Fixed 1

includes/admin/class.llms.admin.page.status.php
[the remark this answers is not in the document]
eri-trabiccolo: Yeah I thought that but it seemed to be incomplete anyway. We're assuming that someone would disable the WP_DEBUG while keeping WP_DEBUG_LOG.
Also at this point let's check the `ini_get('error_log')` anyways then... What do you think?!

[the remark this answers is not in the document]
thomasplevy: hmmm... I'm not sure about that... On my local if I switch `WP_DEBUG_LOG` to `false` and call `ini_get( 'error_log' )` I end up exposing my local server `php-fpm.error` log file. I guess that's actually okay though, right? That error log still provides us with the errors and that's probably a beneficial thing to be able to look at if users have WP_DEBUG_LOG off but their server is still logging errors in a different place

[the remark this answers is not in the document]
thomasplevy: Are there any inherit security issues present with getting this log file when WP_DEBUG_LOG is off?

includes/admin/views/status/view-log.php
eri-trabiccolo: 1. I'm not using the 'I guess' the right way, I guess :)
2. Oh mmm sorry, totally stupid me :(

PR #1092, opened 2020-04-08
Refactor password reset form submission controller and add Add `lostpassword_post` hook

### LINKED ISSUE #846: Programmatic reset password keys fail with LifterLMS reset-password form
LifterLMS has implemented its own version of the WordPress forgot password functionality that is no-longer compliant with the WordPress core system.  This prevents any additional plugins that generate reset password keys programmatically from working.

Specifically, WordPress' built-in `get_password_reset_key()` is not compatible with LifterLMS' reset password functionality, as WordPress has been changed to include a timestamp in the database as part of the key, and Lifter does not check this.

LifterLMS should ideally use WordPress' built-in `get_password_reset_key()` functionality instead of generating its own key, allowing it to work with other plugins and login-based functionality.

### Reproduction Steps

To reproduce, generate a forgot password URL for LifterLMS using WordPress' functionality:

```php
$resetKey = network_site_url(
        "/dashboard/forgot-password?key=" . get_password_reset_key($user) . "&login=" . rawurlencode($user->user_login),
        'login'
);
```

This URL will fail with the message `Invalid key`, as the database includes a timestamp from the `get_password_reset_key()` method that is not checked in LifterLMS here:

https://github.com/gocodebox/lifterlms/blob/f93a34750bfc8b2390239129a112fb5f3686b43d/includes/llms.functions.core.php#L1027

Point in WordPress code where the timestamp is included:

https://core.trac.wordpress.org/browser/tags/5.1.1/src/wp-includes/user.php#L2342

### Expected Behavior

LifterLMS should accept keys generated by WordPress for resetting password keys.

### Actual Behavior

Lifter returns an error from this line, specifically because the timestamp interferes with the check:

https://github.com/gocodebox/lifterlms/blob/f93a34750bfc8b2390239129a112fb5f3686b43d/includes/llms.functions.core.php#L1030

### Error Messages / Logs

N/A

### System Information

<details>
  
N/A

</details>

WordPress Docker image; latest.

### LINKED ISSUE #1089: Reset password use of `lostpassword_post` doesn't pass `WP_Error`.
The core `lostpassword_post` hook (https://developer.wordpress.org/reference/hooks/lostpassword_post/) added a second parameter (a `WP_Error` object in version 4.4.0.

We haven't updated our password flow to mimic this behavior which means any developer hooking into this hook would have to add a custom handler to incorporate our password reset flow into their code.

https://github.com/gocodebox/lifterlms/blob/09c3353aaeb2a8785868f2cf1cf67c4146e94068/includes/forms/frontend/class.llms.frontend.password.php#L51

thomasplevy:
## Description

Refactors the lost password form submission handler.

Adds hook `lostpassword_post` as used by the WordPress core during password reset form submissions.

Fixes #1089 and Fixes #846 

## How has this been tested?

+ Manually tested password reset flows for both successes and failures
+ Added new phpunit tests to handle all potential error and success cases and scenarios

## Types of changes

Non-breaking
Adds hooks
Refactors code for readability
Fixes docblocks throughout the file

## Checklist:


thomasplevy:
71 is failing because of an issue in lifterlms-tests, will be resolved shortly...

thomasplevy:
@eri-trabiccolo please have a quick look through this refactor. I mentioned a team was working on a recaptcha integration this morning and this resolves #1089 which is added by their request to normalize our reset flow to be more like the WP Core for simplicity's sake.

This adds the requested hook and refactors the whole method because it was pretty darn messy. This cherry-picks some changes out of the stale 3.38 branch and actually does a bit better job (I think) than the changes there.

Thanks

includes/forms/controllers/class.llms.controller.account.php
```diff
@@ -121,108 +128,149 @@ public function update() {
...
+		 * Mimics WordPress core behavior so 3rd parties don't need to add special handlers for LifterLMS
+		 * password reset flows.
+		 *
+		 * @since [version]
+		 *
+		 * @link https://developer.wordpress.org/reference/hooks/lostpassword_post/
+		 *
+		 * @param WP_Error $err A WP_Error object containing any errors generated by using invalid credentials.
+		 */
+		do_action( 'lostpassword_post', $err );
```
eri-trabiccolo, line 213: At this point, why we don't also pass the 3rd parameter (`WP_User` or `false`), as the original wp-core hook? https://developer.wordpress.org/reference/hooks/lostpassword_post/
In my opinion we always should ensure the original number of parameters, it's up to the listener deciding how many it needs.

eri-trabiccolo:
@thomasplevy 
> This adds the requested hook and refactors the whole method because it was pretty darn messy. This cherry-picks some changes out of the stale 3.38 branch and actually does a bit better job (I think) than the changes there.
> 
> Thanks

I think you already know this but I do want to be "prudent" here: 3.38 branch "lost password" it's overall still better because it updates the way we generate the pw reset key according to the wp-core. I mean this PR adds the plugin compatibility we need, but still there's something from 3.38 to take into account.
<3

includes/forms/controllers/class.llms.controller.account.php
TimothyBJacobs: I believe this should be fired before `allow_password_reset` to match the WordPress core hook order: https://github.com/WordPress/wordpress-develop/blob/master/src/wp-login.php#L388

eri-trabiccolo: @TimothyBJacobs you mean at line 169 then, makes sense...

thomasplevy: > At this point, why we don't also pass the 3rd parameter (`WP_User` or `false`), as the original wp-core hook? https://developer.wordpress.org/reference/hooks/lostpassword_post/
> In my opinion we always should ensure the original number of parameters, it's up to the listener deciding how many it needs.

Yea, this isn't an opinion. This is an objective fact. I had the second paraemeter in there and then I removed it right before committing... Why? Because the docs I was looking at are out of date. Why do I try to save time by using stuff like this: https://devdocs.io/wordpress/hooks/lostpassword_post

I'll fix it, thanks

thomasplevy:
> @thomasplevy
> 
> > This adds the requested hook and refactors the whole method because it was pretty darn messy. This cherry-picks some changes out of the stale 3.38 branch and actually does a bit better job (I think) than the changes there.
> > Thanks
> 
> I think you already know this but I do want to be "prudent" here: 3.38 branch "lost password" it's overall still better because it updates the way we generate the pw reset key according to the wp-core. I mean this PR adds the plugin compatibility we need, but still there's something from 3.38 to take into account.
> <3

Agreed. I wasn't trying to pull all of the 3.38 changes in since that would require spending a lot more time than warranted by @TimothyBJacobs's request. There's a ton more work to do on that branch and it'll be done when we have some capacity to work on it. For now we can stick with the inferior flow and the issues addressed by 3.38 since most of them are non-breaking.

thomasplevy:
@eri-trabiccolo and @TimothyBJacobs 

This fixes the issues identified by your review.

@eri-trabiccolo I also just got frustrated with the dumbness of this whole flow (this is how we built it back in version 1.x before the wp core had abstracted some of the logic in wp-login.php into callable function (which explains why there's duplicated core functionality re-written into the LifterLMS core).

So I pulled #846 into this and deprecated the offending functions identified there in favor of WP core functions.

The only backwards incompatibility I can find is that if someone upgrades in the middle of a users password reset the reset key generated by LifterLMS before the upgrade will no longer work after the upgrade. I think this is an acceptable sacrifice to make in favor of closing #846 and making LifterLMS's code generally less redundant.

Could you have a re-review and let me know if you spot any issues again.

includes/functions/llms-functions-deprecated.php
```diff
@@ -168,3 +168,50 @@ function llms_expire_membership() {
...
+ * @since 3.8.0
+ * @deprecated [version] Use wp core `check_password_reset_key()` instead.
+ *
+ * @param string $key   Plain text activation key.
+ * @param string $login User login (username).
+ * @return boolean
+ */
+function llms_verify_password_reset_key( $key = '', $login = '' ) {
+
+	llms_deprecated_function( 'llms_verifyr_password_reset_key()', '3.37.17', 'check_password_reset_key()' );
```
eri-trabiccolo, line 207:
```suggestion
	llms_deprecated_function( 'llms_verify_password_reset_key()', '3.37.17', 'check_password_reset_key()' );
```

man I think you have a problem with the 'r' on your keyboard, please check :P

thomasplevy: I actually literally do have a problem with the 'r' on my keyboard but in this case it's not the keyboard's fault, I'm just dumb I guess.

includes/functions/llms.functions.person.php
```diff
@@ -23,7 +23,7 @@
  * @param  string $lastname          [user last name]
  * @param  string $password          [user password]
  * @param  string $password2         [user verify password]
- * @param  string $billing_address_1 [user billing address 1]
+ * @param  string $billing_address_1 [user billinrg address 1]
```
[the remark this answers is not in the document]
thomasplevy: I don't think this was my keyboards fault but whatever.

Issue #849, opened 2019-05-06
Student Dashboard endpoints (lost password, order history, & edit account) 404 with WooCommerce 3.6 or later


thomasplevy:
### Reproduction Steps

+ Install LifterLMS and WooCommerce 3.6 or later
+ Access the "lost password" endpoint on the LifterLMS student dashboard

### Expected Behavior

+ Should be able to reset password via the password reset form

### Actual Behavior

+ A 404 page is encountered

### Error Messages / Logs

N/A

### System Information

LifterLMS (any version) and WooCommerce 3.6 or later

thomasplevy:
Due to changes in WC 3.6 ([commit](https://github.com/woocommerce/woocommerce/commit/a0f47fb429266e1d55fbe5f77db928e478682973)) any WC endpoint accessed outside of the WC account page will 404.

Shared LifterLMS endpoints will no longer be accessible as a result

thomasplevy:
Above commit resolves this in a somewhat acceptable manner. I'd really like a filter in WC that would allow us to not lie to WC via the `woocommerce_is_account_page` filter.

Will adjust this based on results of this issue in the WC GitHub: https://github.com/woocommerce/woocommerce/issues/23622

eri-trabiccolo:
This PR of yours @thomasplevy has been accepted:
https://github.com/woocommerce/woocommerce/pull/23660

that's the filter we actually needed, and it will also solve #1037!

PR #1094, opened 2020-04-10
Fix dashboard endpoints conflicts with WooCommerce

### LINKED ISSUE #1037: WooCommerce causing LifterLMS password reset looping
### Reproduction Steps

Turn on LifterLMS and WooCommerce.
Do not install or turn on LifterLMS WooCommerce integration.
Set LifterLMS Dashboard to its own dashboard page.
Set WooCommerce my account to its own my account page.
From the logged out LifterLMS Dashboard click password reset and run through the process.
With WooCommerce active, even with no crossing of my account page/dashboard settings or shortcodes LifterLMS password reset loops back to the original password input page.

### Expected Behavior

LifterLMS password reset should operate as expected.

### Actual Behavior

It loops back to the input email page instead of going to the new password input page.

### System Information

LifterLMS
WooCommerce

HS-111422

eri-trabiccolo:
## Description

Fixes #1037 and #849 (which was already fixed)

I removed a public method of the `LLMS_PlayNice` class instead of making it deprecated, I thought it was good since that class instance is not accessible from outside that file.

Additionally, we have also this in our integration:
https://github.com/gocodebox/lifterlms-integration-woocommerce/blob/2.0.15/includes/class-llms-wc-my-account.php#L30

I don't think we really need that anymore... I mean, it's not technically needed.
Shall we remove it?

## How has this been tested?
I tested manually that issue described in #1037 occurred before and that both #1037 and #849 do not occur with this PR applied.

## Types of changes

## Checklist:

### DIFF
--- a/includes/class.llms.playnice.php
+++ b/includes/class.llms.playnice.php
@@ -16,6 +16,7 @@ defined( 'ABSPATH' ) || exit;
  * @since 3.1.3
  * @since 3.31.0 Resolve dashboard endpoint 404s resulting from changes in WC 3.6.
  * @since [version] Changed the way we handle the dashboard endpoints conflict, using a different wc filter hook.
+ *                 Deprecated `LLMS_PlayNice::wc_is_account_page()`.
  */
 class LLMS_PlayNice {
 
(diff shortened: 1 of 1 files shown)


thomasplevy:
@eri-trabiccolo 

I don't mean to be a pain but it is technically possible to call the public functions from outside the file so while we really do not need it anymore it would be better not to remove it. 

I want us to continue to do our best to follow semantic versioning and if we do remove a public function we need to bump to 4.0 and I don't think think this warrants a 4.0 bump.

Just leave the old method and deprecate it and add the new one. It's "bloat" but it's safer than removing it -- even though it's very likely no one in the world is calling I'd still feel better about not messing with it.

Thanks

PR #1115, opened 2020-04-20
fix deprecation warning resulting from theme support

thomasplevy:
## Description

This PR quiets ~~@eri-trabiccolo~~ deprecation logs encountered when using themes that add theme support for `lifterlms-quizzes` and do not utilize the deprecated filter `llms_get_quiz_theme_settings` 

## How has this been tested?

Undesired behavior (deprecation warning logged to the `llms.log` file despite the deprecated methods not being used) can be seen by adding theme support to any theme (like twentytwenty) using the following code:

```php
add_action( 'after_setup_theme', function() {
	add_theme_support( 'lifterlms-quizzes' );
} );
```

Without this PR a deprecated message is logged despite not using the deprecated method. After this PR the depercation warning is no longer logged.

When the filter is being used the warning is still recorded:

Add this code to the same theme:

```php
add_filter( 'llms_get_quiz_theme_settings', function( $settings ) {
	$settings['layout']['options'] = array( 'Full Width', 'Left', 'Right' );
	return $settings;
} );
```

With both snippets the custom fields will be displayed on the builder and the warning will also be logged to the llms.log file (as expected).

## Checklist:

### DIFF
--- a/includes/admin/class.llms.admin.builder.php
+++ b/includes/admin/class.llms.admin.builder.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Admin/Classes
  *
  * @since 3.13.0
- * @version 3.37.12
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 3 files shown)


thomasplevy:
@eri-trabiccolo updated

PR #1117, opened 2020-04-22
Refactor the notification background processor dispatching methods

thomasplevy:
## Description

This update refactors the code associated with the asynchronous scheduling of notification processor classes.

This doesn't actually fix any issues in the LifterLMS core, but in working on https://github.com/gocodebox/lifterlms-integration-twilio/issues/19 I discovered that there is a potential issue with the way processors are scheduled.

The SMS notification processor in the Twilio plugin is unreliable when tested on my local machine (although these issues cannot be reliably reproduced outside of my environment as @eri-trabiccolo has been unable to find any errors on his local machine).

In any event the issue appears to be related to having multiple processors scheduled to be dispatched using the background processor in the core plugin.

This refactor offloads the async scheduling to the (more reliable) action scheduler library included with the LifterLMS core.

There are some minor changes in this code (like using strict comparisons and wp recommended functions) as well as doc / cs updates to some of the modified files.

EDIT: 
There is a really small downside to this: this method is *slightly* slower than the old method. Using the old method it looks like *mostly* emails were never pushed to a queue where the cron healthcheck would end up sending them. Instead, they would be sent almost instantly via the background process that's spawned upon dispatch. With this method it will *always* be triggered by a cron (via the action schedule) which runs once per minute. So there's the possibility of having at 1 minute delay on the notifications. I don't think this is really a concern *at all* but it's worth noting...

EDIT EDIT:

Furthermore since it will *always* be handled by a cron, sites with malfunctioning or broken crons will never send emails at all. So that's something. But if the cron is broken they're going to get in touch with us about it at some point anyway and we can provide them with steps they can take to get the cron working

## How has this been tested?

+ I have written new unit tests to test all the individual methods introduced in this PR 
+ I have manually tested the existing (and Twilio) notification processors continue to work in the following ways: 

1) Tested notifications processed during enrollment via manual enrollment on the admin panel
2) Notifications processed when a student enrolls in free or paid courses from the frontend
3) Other notifications (like course complete) continue to work

## Types of changes

This introduces a new method of scheduling the background processes that is backwards compatible
However, it does add a new function and deprecates the previous function. The previous function is still a publicly callbable function, however it now logs a deprecation notice when used.

A new filter as been introduced which will allow users to switch to the "old" method that doesn't utilize the action scheduler. This will allow us to quickly switch users back to the old method if for some reason that I can't conceive of currently but someone will likely come up with someone wants the old method.

I am not aware of any 3rd party plugins we recommend that have implemented their own processors but unless they are doing something very custom these changes should improve the reliability of their processors with no negative impact.

## Checklist:

### DIFF
--- a/includes/notifications/class.llms.notifications.php
+++ b/includes/notifications/class.llms.notifications.php
@@ -25,14 +25,14 @@ class LLMS_Notifications {
 	/**
 	 * Singleton instance
 	 *
-	 * @var  LLMS_Notifications
+	 * @var LLMS_Notifications
 	 */
 	protected static $_instance = null;
 
 	/**
 	 * Controller instances
 	 *
-	 * @var  LLMS_Abstract_Notification_Controller[]
+	 * @var LLMS_Abstract_Notification_Controller[]
 	 */
 	private $controllers = array();
 
@@ -46,30 +46,30 @@ class LLMS_Notifications {
 	/**
 	 * Background processor instances
 	 *
-	 * @var  LLMS_Abstract_Notification_Processor[]
+	 * @var LLMS_Abstract_Notification_Processor[]
 	 */
 	private $processors = array();
 
 	/**
 	 * Array of processors needing to be dispatched on shutdown
 	 *
-	 * @var  string[]
+	 * @var string[]
 	 */
 	private $processors_to_dispatch = array();
 
 	/**
 	 * [string $view_classname => string $trigger ]
 	 *
-	 * @var  string[]
+	 * @var string[]
 	 */
 	private $views = array();
 
 	/**
 	 * Main Instance
 	 *
-	 * @return    LLMS_Notifications
-	 * @since     3.8.0
-	 * @version   3.8.0
+	 * @since 3.8.0
+	 *
+	 * @return LLMS_Notifications
 	 */
 	public static function instance() {
 		if ( is_null( self::$_instance ) ) {
@@ -220,9 +220,9 @@ class LLMS_Notifications {
 	/**
 	 * Get the directory path for core notification classes
 	 *
-	 * @return   string
-	 * @since    3.8.0
-	 * @version  3.8.0
+	 * @since 3.8.0
+	 *
+	 * @return string
 	 */
 	private function get_directory() {
 		return LLMS_PLUGIN_DIR . 'includes/notifications/';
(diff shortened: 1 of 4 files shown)


includes/notifications/class.llms.notifications.php
```diff
@@ -423,19 +482,96 @@ private function name_to_file( $name ) {
...
+		// Save items in the queue.
+		$processor->save();
+
+		// Check if there's already a scheduled event.
+		$timestamp = as_next_scheduled_action( $hook, $args );
+
+		// If there's no event scheduled already, schedule one.
+		if ( ! $timestamp ) {
+
+			$timestamp = llms_current_time( 'timestamp' );
```
eri-trabiccolo, line 561: Ok we have a problem here:
1) `llms_current_time( 'timestamp' )` calls `current_time( 'timestamp' )` which is _deprecated_ (it adjusts the UTC unix timestamp computed using`time()` adding the UTC offset as per WP settings. )
2) the action scheduler expects the time (`$timestamp` here) parameter to be UTC, while we're passing here the local time. This means that if in our WP settings we set the timezone to UTC+2 the action will be scheduled to be fired in 2 hours from now (I can reproduce this).

So let's go with:
`llms_current_time( 'timestamp', 1 )` that will allow us to continue testing it (which is why we have `llms_current_time` here, right?)
1) will call `current_time( 'timestamp', 1 )` that returns just `time()` and it's not _deprecated_
2) also this fixes 2
3) we all are happy

includes/notifications/class.llms.notifications.php
```diff
@@ -18,20 +18,21 @@
  * @since 3.8.0
  * @since 3.24.0 Unknown.
  * @since 3.36.1 Record notifications as read during the `wp_print_footer_scripts` hook.
+ * @since [version] Updated processor scheduling for increased performance and reliability.
```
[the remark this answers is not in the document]
thomasplevy: I'm going to stop duplicating every single set of minor changes that are logged on methods into the class too. The redundancy seems unnecessary, thoughts?

includes/notifications/class.llms.notifications.php
thomasplevy: Great catch!

includes/notifications/class.llms.notifications.php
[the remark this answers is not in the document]
eri-trabiccolo: I'm totally fine with this!

includes/notifications/class.llms.notifications.php
thomasplevy: @eri-trabiccolo updated, good to merge?

PR #1134, opened 2020-05-04
Avoid certificate bg alignment cut-offs

### LINKED ISSUE #1133: Certificate image can be cut off
### Reproduction Steps

It can be reproduced with the Launchpad theme and the default certificate, even if it's not that visible (or maybe not visible at all because the default certificate has some white gap all around).

It gets more visible if you set a bigger body/content font-size and/or line-height.
Or even more if the theme doesn't align images vertically to the middle (e.g. **Divi**).

All this is because images are inline (inline-block) elements, by default aligned on the baseline - and because we use a negative margin on the image.

### Expected Behavior

+ certificate image not cut off

### Actual Behavior

+ certificate image cut off

### Screenshot
Launchpad theme - default certificate.
To highlight the issue I bordered the certificate container with a 2px solid black border, allow its children overflowing it and bordered the certificate image with a 2px solid red border.

Schermata 2020-05-01 alle 19 02 36

### Error Messages / Logs

n/a

### System Information

n/a

n/a

### Related User Information
HS-119882

### Possible solution
Force the img to display as a block:
`.llms-certificate-container .certificate-background { display: block }`

eri-trabiccolo:
## Description

Fixes #1133 

## How has this been tested?
manually

## Types of changes

## Checklist:

### DIFF
--- a/assets/scss/certificates.scss
+++ b/assets/scss/certificates.scss
@@ -67,6 +67,7 @@ body {
 		position: relative;
 		z-index: 1;
 		width: 100%;
+		display: block;
 		// position: absolute;
 		// z-index: 1;
 		// top: 0;


thomasplevy:
@eri-trabiccolo 

Can you confirm that this both fixes the issue in LP and also does not create new issues in other popular and recommended themes?

Can you check astra, twentytwenty, and oceanwp at a minimum please

thomasplevy:
(I can't imagine it does cause issues in any theme but just in case...)

eri-trabiccolo:
@thomasplevy
I checked Astra, LP, Divi, OceanWP, TwentyTwenty

PR #1137, opened 2020-05-06
Fix bbpress forums not private

### LINKED ISSUE #1132: Bbpress Forums not private
### Reproduction Steps

Install Bbpress and activate
Create a forum and attach the forum to a course.

### Expected Behavior

The forum should be accessed by students only enrolled in that course

### Actual Behavior

Anyone with the forum link can access the course(read-only)

HS-119870

eri-trabiccolo:
## Description

Fixes #1132 

## How has this been tested?
Manually and with new unit tests

## Types of changes

The new query to retrieve the courses associated to a fourm takes care of the backward compatibility, when we used to save forum ids as array of integers.
It also fixes a possible very edge case where when looking for forum ids we could also erroneously match the index of the serialized array.

@thomasplevy in our bug scrub we talked about an alternative method to avoid the wrong match, well we were tired, that just couldn't work - we said that we would have checked the the retrieved posts were actually courses - because the edge case of course would have matched courses since only courses have the meta key we match against :D

## Checklist:

### DIFF
--- a/includes/integrations/class.llms.integration.bbpress.php
+++ b/includes/integrations/class.llms.integration.bbpress.php
@@ -3,7 +3,7 @@
  * bbPress Integration
  *
  * @since 3.0.0
- * @version 3.37.11
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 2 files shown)


eri-trabiccolo:
Ops, travis fails for a db problem with the regex, why so?! this is something I can't reproduce locally.

thomasplevy:
travis is now failing because of composer timeouts... i need to figure out how to fix that now...

eri-trabiccolo:
only the test coverage though.
That was kinda slow on my machine too I remember...

PR #1138, opened 2020-05-06
Membership associations

thomasplevy:
## Description

In working on https://github.com/gocodebox/lifterlms-groups/issues/61 I determined that it's currently impossible to easily determine what content on a LLMS-powered site is considered to be "part" of a membership.

This PR tackles the very early stages of answering this question.

This code puts forth that what belongs to a membership is:

https://github.com/gocodebox/lifterlms/blob/4da2a032b25bdb46ac3415df2e40f9c034232236/includes/models/model.llms.membership.php#L99-L102

This PR adds a single public method, `LLMS_Membership::get_associated_posts()` which queries for associations as defined by the above statement.

Initially, this method will be used to solve the groups reporting issue identified by https://github.com/gocodebox/lifterlms-groups/issues/61 but in the future we should build an editor block and possibly a widget and maybe a shortcode (even though I really don't want to build any new shortcodes ever) that can be used on membership pages and etc... to provide membership members with lists of content available through the membership, instead of forcing site admins to list this content out manually.

## How has this been tested?

Since the new method won't actually be used anywhere in the core codebase there's no way to actually see it in action except via the new tests that have been written and which are included in this PR.

## Types of changes

 New public function added to the internal api.

## Checklist:

### DIFF
--- a/includes/models/model.llms.membership.php
+++ b/includes/models/model.llms.membership.php
@@ -32,8 +32,7 @@ defined( 'ABSPATH' ) || exit;
  */
 class LLMS_Membership
 extends LLMS_Post_Model
-implements LLMS_Interface_Post_Instructors
-		 , LLMS_Interface_Post_Sales_Page {
+implements LLMS_Interface_Post_Instructors, LLMS_Interface_Post_Sales_Page {
 
 	/**
 	 * Membership post meta.
@@ -293,12 +292,13 @@ implements LLMS_Interface_Post_Instructors
 	 * Determine if sales page redirection is enabled
 	 *
 	 * @since 3.20.0
+	 * @since [version] Use strict array comparison.
 	 *
 	 * @return string
 	 */
 	public function has_sales_page_redirect() {
 		$type = $this->get( 'sales_page_content_type' );
-		return apply_filters( 'llms_membership_has_sales_page_redirect', in_array( $type, array( 'page', 'url' ) ), $this, $type );
+		return apply_filters( 'llms_membership_has_sales_page_redirect', in_array( $type, array( 'page', 'url' ), true ), $this, $type );
 	}
 
 	/**
@@ -379,9 +379,9 @@ implements LLMS_Interface_Post_Instructors
 				   AND metas.meta_key = %s
 				   AND metas.meta_value REGEXP %s;",
 					$post_type,
-				    $enabled_key,
-				    $enabled_value,
-				    $list_key,
+					$enabled_key,
+					$enabled_value,
+					$list_key,
 					'a:[0-9][0-9]*:{(i:[0-9][0-9]*;(i|s:[0-9][0-9]*):"?[0-9][0-9]*"?;)*(i:[0-9][0-9]*;(i|s:[0-9][0-9]*):"?' . $this->get( 'id' ) . '"?;)'
 				)
 			);
--- a/includes/models/model.llms.membership.php
+++ b/includes/models/model.llms.membership.php
@@ -2,7 +2,7 @@
 /**
  * LifterLMS Membership Model
  *
- * @package LifterLMS/Models/Classes
+ * @package LifterLMS/Models
  *
  * @since 3.0.0
  * @version [version]
--- a/includes/models/model.llms.membership.php
+++ b/includes/models/model.llms.membership.php
@@ -2,9 +2,10 @@
 /**
  * LifterLMS Membership Model
  *
- * @package  LifterLMS/Models
- * @since    3.0.0
- * @version  3.36.3
+ * @package LifterLMS/Models
+ *
+ * @since 3.0.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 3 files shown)


thomasplevy:
@eri-trabiccolo your regexp done to fix that bbPress issue makes a surprise appearance!

includes/models/model.llms.membership.php
```diff
@@ -92,6 +93,65 @@ public function add_auto_enroll_courses( $course_ids, $replace = false ) {
...
+		$posts = apply_filters( 'llms_membership_get_associated_posts', $posts, $post_type, $this );
+
+		// If a single post type was requested, return only that.
+		if ( $post_type ) {
+			// Return the request post type array and fallback to an empty array if that post type doesn't exist.
+			return isset( $posts[ $post_type ] ) ? $posts[ $post_type ] : array();
+		}
+
+		// Otherwise return everything.
+		return $posts;
```
eri-trabiccolo, line 151: At a certain point I think we should remove the empty arrays, don't you think?
What's the meaning of having something like `array('course'=>array())` as result?

```suggestion
		return array_filter( $posts );
```

tests/phpunit/unit-tests/models/class-llms-test-model-llms-membership.php
```diff
@@ -68,6 +68,73 @@ protected function get_data() {
...
+	 * @return void
+	 */
+	public function test_get_associated_posts_has_associations() {
+
+		$membership = $this->factory->membership->create_and_get();
+
+		// Add a post.
+		$post = $this->factory->post->create();
+		update_post_meta( $post, '_llms_is_restricted', 'yes' );
+		update_post_meta( $post, '_llms_restricted_levels', array( $membership->get( 'id' ), 1, 1008, '183' ) );
```
eri-trabiccolo, line 110: how mean are you to stress my regexp with mixed types?

includes/models/model.llms.membership.php
```diff
@@ -251,6 +312,92 @@ public function instructors() {
...
+	 * @return int[]
+	 */
+	protected function query_associated_posts( $post_type, $enabled_key, $enabled_value, $list_key ) {
+
+		global $wpdb;
+
+		// See if we have a cached result first.
+		$cache = sprintf( 'membership_%1$d_associated_%2$s', $this->get( 'id' ), $post_type );
+		$found = null;
+		$ids   = wp_cache_get( $cache, '', false, $found );
```
eri-trabiccolo, line 363: we rely on wp to invalidate (clean) the object cache, do you think this can be a problem with persistent object caching? you know object-cache.php drop-in...

tests/phpunit/unit-tests/models/class-llms-test-model-llms-membership.php
thomasplevy: :-p

includes/models/model.llms.membership.php
thomasplevy: Of course. Any amount of caching is in danger of being screwed up by drop-ins and caching plugins...

This particular area though caches a complicated and potentially slow query in order to prevent the same query from running multiple times on the same page load which I forsee could potentially happening in future utilizations of this.

I think the fact that a drop-in could make this not work as intended is an acceptable risk...

If we don't cache it someone eventually will tell us were not caching a complicated query that we should be and if we do cache it someone will drop in persistent object caching without thinking about the consequences of doing so...

However... given the fact that all persistent object caches do eventually expire this will cause delays in a new post association showing up on the froentend somewhere but won't really break any existing functionality...

Yea that all causes support but caching in this way exists for a reason and I don't want to shy away from using it because shared webhosts cache things forever all the time and then also blame us for being too resource intensive..

includes/models/model.llms.membership.php
thomasplevy: I was thinking you could have an expected response that always has the keys so that you wouldn't have to always check the presence of a key before interacting the the response object. But now that I'm thinking about it I guess that wouldn't exactly matter because you couldn't always guarantee that it's going to be courses/posts/pages as a theme or plugin could modify post type support so you'd want to check the presence anwyay.

thomasplevy: I'll take a look at some modifications here and testse are going to have to be rewritten

includes/models/model.llms.membership.php
eri-trabiccolo: Yeah I see, I was more thinking about a way to clear this at shutdown, or using a you know class property as "cache". But yeah I see your point.

thomasplevy: I don't know how to respond to this...

We're ALLOWED and SHOULD use the WP cache API where possible. Yes, there might be caching issues in the world but that doesn't mean we should start building some kind of crazy cache handler code that we then need to maintain and support in order to accommodate a theoretical issue like this...

I don't know man... That feels really crazy to me

includes/models/model.llms.membership.php
thomasplevy: After reviwing I think your suggestion is perfect. I just updated the tests to accommodate this return.

includes/models/model.llms.membership.php
eri-trabiccolo: I think you're right.
Mine was more something like: "Can we do something with little effort that will avoid us getting support requests about object caching issues? e.g. - different approach - flushing this cache on the specific post metas update."
This was more a general thought and whether it was something to explore, you know, if it was worthwhile.

That's all, fine for me! :)

PR #1150, opened 2020-05-15
Make sure a question's question_type is always available, 'choice' by…

### LINKED ISSUE #1143: LifterLMS Quiz Screen not Loading in Builder
### Reproduction Steps

+ Open the builder for a course
+ Attempt to look at a quiz in the course

### Expected Behavior

+ The quiz settings and questions should load

### Actual Behavior

+ there appears to be a javascript error that loads and loads
+ you have to look at a different lesson setting and come back to the quiz to get it to load

### Error Messages / Logs

```
Uncaught TypeError: Cannot read property 'choices' of undefined
    at N.d.initialize (Question.js:115)
    at N.d.b.Model (backbone.min.js?ver=1.2.3:1)
    at new N.d (backbone.min.js?ver=1.2.3:1)
    at N.d._prepareModel (backbone.min.js?ver=1.2.3:1)
    at N.d.set (backbone.min.js?ver=1.2.3:1)
    at N.d.add (backbone.min.js?ver=1.2.3:1)
    at N.d.add_question (Quiz.js:126)
    at underscore.min.js?ver=1.8.3:1
    at Function.s.each.s.forEach (underscore.min.js?ver=1.8.3:1)
    at Object.success (Quiz.js:202)

```

<details>

```
Wordpress

Home Url: [removed]
Site Url: [removed]
Login Url: [removed]/wp-login.php
Version: 5.2.4
Debug Mode: No
Debug Log: No
Debug Display: Yes
Locale: en_US
Multisite: No
Page For Posts: Not Set
Page On Front: Home (#207) [[removed]/]
Show On Front: page
Wp Cron: Yes


Settings

Version: 3.38.1
Db Version: 3.38.1
Course Catalog: Course Catalog (#8596) [[removed]/courses/]
Membership Catalog: Membership Catalog (#8597) [[removed]/memberships/]
Student Dashboard: My Courses (#8599) [[removed]/my-courses/]
Checkout Page: Purchase (#8598) [[removed]/purchase/]
Course Catalog Per Page: 9
Course Catalog Sorting: menu_order
Membership Catalog Per Page: 9
Membership Catalog Sorting: menu_order
Site Membership: Not Set
Courses Endpoint: my-courses
Edit Endpoint: edit-account
Lost Password Endpoint: lost-password
Vouchers Endpoint: redeem-voucher
Autogenerate Username: yes
Password Strength Meter: yes
Minimum Password Strength: strong
Terms Required: no
Terms Page: Not Set
Checkout Names: required
Checkout Address: required
Checkout Phone: optional
Checkout Email Confirmation: yes
Open Registration: no
Registration Names: required
Registration Address: optional
Registration Phone: hidden
Registration Voucher: optional
Registration Email Confirmation: no
Account Names: required
Account Address: required
Account Phone: optional
Account Email Confirmation: yes
Confirmation Endpoint: confirm-payment
Force Ssl Checkout: no
Country: CA
Currency: CAD
Currency Position: left
Thousand Separator: ,
Decimal Separator: .
Decimals: 2
Trim Zero Decimals: no
Recurring Payments: yes
Email From Address: [removed]
Email From Name: [removed]
Email Footer Text: 
Email Header Image: 
Cert Bg Width: 800
Cert Bg Height: 616
Cert Legacy Compat: no


Gateways

Manual: Disabled
Manual Logging: 
Manual Order: 1


Server

Mysql Version: 5.5.5
Php Curl: Yes
Php Default Timezone: UTC
Php Fsockopen: Yes
Php Max Input Vars: 1000
Php Max Upload Size: 128 MB
Php Memory Limit: 260M
Php Post Max Size: 128M
Php Soap: Yes
Php Suhosin: No
Php Time Limt: 30
Php Version: 7.3.17
Software: Apache
Wp Memory Limit: 40M


Browser

HTTP USER AGENT: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/81.0.4044.138 Safari/537.36


Theme

Name: Twenty Twenty
Version: 1.2
Themeuri: https://wordpress.org/themes/twentytwenty/
Authoruri: https://wordpress.org/
Template: 
Child Theme: No
Llms Support: No


Plugins

LifterLMS: 3.38.1
User Switching: 1.5.4


Integrations

BbPress: No
BuddyPress: No


Template Overrides




```

</details>

This issue has be recreated:
+ [ ] Locally
+ [X] On a staging site
+ [X] On a production website
+ [X] With only LifterLMS and a default theme

HS-121255

eri-trabiccolo:
… default

## Description
Fixes #1143 

## How has this been tested?

First reproduce the issue described in #1143 by emptying the question_type of a question in the db (either manually, with code, I used an exported course that had this problem). [EDIT: `emptying` here was a wrong concept, I meant: set the post meta `_llms_question_type` to an empty string]
Then apply this patch and see that the questions are correctly loaded in the builder.
Also make some change to the "broken" question so that on save the question type will be saved as 'choice'.
Check that this actually happened, either manually in the db or getting the question's `question_type` raw value:
`$question->get( 'question_type', true );`

## Types of changes
Bug fix (non-breaking change which fixes an issue).
Well I think this is borderline, it's not a bug fix, but something that takes into account possible faults of the stored data, possibly because of a db update not went completely through!?

## Checklist:

@thomasplevy 
I haven't written any unit test yet, can you tell me if you think this is a valid way to approach the issue?

### DIFF
--- a/includes/admin/class.llms.admin.builder.php
+++ b/includes/admin/class.llms.admin.builder.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Admin/Classes
  *
  * @since 3.13.0
- * @version 3.38.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -20,6 +20,7 @@ defined( 'ABSPATH' ) || exit;
  *                Added new filter, `llms_builder_{$post_type}_force_delete` to allow control of how post type deletion is handled
  *                when deleted via the builder.
  * @since 3.38.0 Improve backwards compatibility handling for the `llms_get_quiz_theme_settings` filter.
+ * @since [version] On quiz saving, made sure that a question as a type set, otherwise set it by default to `'choice'`.
  */
 class LLMS_Admin_Builder {
 
@@ -1051,11 +1052,13 @@ class LLMS_Admin_Builder {
 	/**
 	 * Update quiz questions from heartbeat data
 	 *
-	 * @param    array $questions  question data array
-	 * @param    obj   $parent    instance of an LLMS_Quiz or LLMS_Question (group)
-	 * @return   array
-	 * @since    3.16.0
-	 * @version  3.16.11
+	 * @since 3.16.0
+	 * @since 3.16.11 Unknown.
+	 * @since [version] Make sure that a question as a type set, otherwise set it by default to `'choice'`.
+	 *
+	 * @param array $questions Question data array.
+	 * @param obj   $parent    Instance of an LLMS_Quiz or LLMS_Question (group).
+	 * @return array
 	 */
 	private static function update_questions( $questions, $parent ) {
 
@@ -1070,16 +1073,16 @@ class LLMS_Admin_Builder {
 				)
 			);
 
-			// remove temp id if we have one so we'll create a new question
+			// Remove temp id if we have one so we'll create a new question.
 			if ( self::is_temp_id( $q_data['id'] ) ) {
 				unset( $q_data['id'] );
 			}
 
-			// remove choices because we'll add them individually after creation
+			// Remove choices because we'll add them individually after creation.
 			$choices = ( isset( $q_data['choices'] ) && is_array( $q_data['choices'] ) ) ? $q_data['choices'] : false;
 			unset( $q_data['choices'] );
 
-			// remove child questions if it's a question group
+			// Remove child questions if it's a question group.
 			$questions = ( isset( $q_data['questions'] ) && is_array( $q_data['questions'] ) ) ? $q_data['questions'] : false;
 			unset( $q_data['questions'] );
 
@@ -1096,6 +1099,14 @@ class LLMS_Admin_Builder {
 
 				$question = $parent->questions()->get_question( $question_id );
 
+				/**
+				 * When saving a question, make sure that it has a question type set
+				 * otherwise set it by default to `'choice'`.
+				 */
+				if ( ! $question->get( 'question_type', true ) ) {
+					$question->set( 'question_type', 'choice' );
+				}
+
 				if ( $choices ) {
 
 					$ret['choices'] = array();
@@ -1111,7 +1122,7 @@ class LLMS_Admin_Builder {
 
 						unset( $c_data['question_id'] );
 
-						// remove the temp ID so that we create it if it's new
+						// Remove the temp ID so that we create it if it's new.
 						if ( self::is_temp_id( $c_data['id'] ) ) {
 							unset( $c_data['id'] );
 						}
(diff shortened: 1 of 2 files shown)


includes/models/model.llms.question.php
```diff
@@ -175,6 +176,29 @@ protected function get_creation_args( $args = null ) {
 
 	}
 
+
```
thomasplevy, line 179: Instead of adding this logic by overriding the function, can we use the default property values to automatically fallback here?

From the `___get()` method:

https://github.com/gocodebox/lifterlms/blob/1d61d0b2f54b3ce5578f796ac611a04f66d2d229/includes/abstracts/abstract.llms.post.model.php#L437

Then define the default value in the default properties array:

https://github.com/gocodebox/lifterlms/blob/master/includes/abstracts/abstract.llms.post.model.php#L112

Here's a real example of setting default properties: https://github.com/gocodebox/lifterlms/blob/237bf80035a77bacf36ca4d8e99a8f3df92407d2/includes/models/model.llms.lesson.php#L72-L82

includes/admin/class.llms.admin.builder.php
```diff
@@ -1096,6 +1099,14 @@ private static function update_questions( $questions, $parent ) {
 
 				$question = $parent->questions()->get_question( $question_id );
 
+				/**
```
thomasplevy, line 1102: Without testing this I'm not 100% positive but if we have a "default" value then we don't actually need this bit of code, right?

What I mean is doing both this and then also setting a fallback value when the value isn't set is a bit redundant.

I think the idea is that this would "fix" the problem from happening in the future and setting a default would help with existing items. However I don't think we technically need both sets of code.

Since having a default value will protect against malformed API requests and already malformed questions I'm partial to that one if we had to choose.

But maybe I'm missing something?

eri-trabiccolo: @thomasplevy 

No and yes.
No: you're not missing anything from what I said - sorry.
Yes: because I used in many places "set" and "not set" concepts but I HAD to say that the culprit `question_type` - according to the course imported - IS SET but IS EMPTY.
So setting the `$property_defaults` doesn't work - that's a fallback for non set property.

The redundancy I added was just to provide a way to normalize data with minimum effort.
Do you think I can remove it then?

I think I can also add the default property as you suggest to be more strict...

thomasplevy: hmmm... that's weird then, i see the reasoning and in that case this (as it is without the default properties) should be good.

I know we don't know but I'm really curious how these questions are ending up with no type in the database... Seems weird

PR #1158, opened 2020-05-20
move update functions into their own files grouped by update version

thomasplevy:
## Description

Reorganizes update functions by moving each version's update functions into their own files.

## How has this been tested?

Manually rolled the db back to ensure db updates still get queued and ran on update 

## Types of changes

Organizes code so the updates file isn't so massive

## Checklist:


thomasplevy:
@eri-trabiccolo don't need a code review, exactly, it's a lot of code but it's actually just existing code moved around and re-organized.

Does this organizational strategy make sense to you?

PR #1171, opened 2020-05-29
remove classnames that conflict with bootstrap

### LINKED ISSUE #888: Login form doesn't display fine when bootstrap4 style is enqueued
Need to enqueue bootstrap4 CSS or use a theme/plugin which enqueues it.
Schermata 2019-06-24 alle 14 18 37

The user reporting this issue is using the theme Airi Theme which is based on bootstrap4 as far as I can see.

The issue resides on the fact that one of the llms form wrapper CSS classes is `col-1` which in bootstrap4 is styled as following:
`.col-1 {
    -ms-flex: 0 0 8.333333%;
    flex: 0 0 8.333333%;
    max-width: 8.333333%;
}`
which means that the form box width will take 8.3% of its container...

I think we need to "namespace" that class to avoid this kind of issues.

thomasplevy:
## Description

Resolves #888

Removes classnames from student dashbord login and registration form wrapper elements which conflict with bootstrap causing visual issues.

These classes are not used by the LifterLMS core or add-ons and are a legacy class that hasn't been removed for fear of creating backwards compatibility issues with any custom css, 3rd party themes, etc...

## How has this been tested?

Visually inspected forms with common recommended themes and LaunchPad after removal

## Types of changes

**Breaking change**: removes css classes from the following templates:

+ templates/global/form-login.php: Removes `col-1` class from the `div.llms-person-login-form-wrapper` element.
+ templates/global/form-registration.php: : Removes `col-2` class from the `div.llms-new-person-form-wrapper` element.

## Checklist:

### DIFF
--- a/templates/global/form-login.php
+++ b/templates/global/form-login.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Templates
  *
  * @since 3.0.0
- * @version 3.21.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -31,7 +31,7 @@ if ( ! empty( $message ) ) {
 
 <?php do_action( 'llms_before_person_login_form' ); ?>
 
-<div class="col-1 llms-person-login-form-wrapper">
+<div class="llms-person-login-form-wrapper">
 
 	<form action="" class="llms-login" method="POST">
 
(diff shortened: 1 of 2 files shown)


thomasplevy:
@eri-trabiccolo what do you think? I think it's time.

eri-trabiccolo:
@thomasplevy, I think that too!
We always struggle to be compatible with everyone else in the wp galaxy, let's go kick some asses :P

PR #1223, opened 2020-07-01
Updated supported versions

thomasplevy:
Added 4.x as the only supported version. Declared 3.x as unsupported.


eri-trabiccolo (approved):
I don't think we should formalize on the indentation, right?!

PR #1225, opened 2020-07-01
LifterLMS 4.1.0 (Dev to Trunk)

thomasplevy:
+ PR #1224


thomasplevy:
@eri-trabiccolo I don't believe we're working on anything else that should be included in 4.1.0, if so let me know, otherwise if you approve we can shoot to release tomorrow

eri-trabiccolo (approved):

PR #1226, opened 2020-07-01
Return the author's display name as the default instructor's name 

nicolas-jaussaud:
## Description
The get_instructors() method of the LLMS_Post_Instructors() is not returning the same result when there is no instructors set and it fallback to the post author.

When instructors are set, the method return for each instructor an array with the following indexes: label, visibility, id and name.

However, when there is no instructors set, the array of the instructor only contain 3 indexes: label, visibility and id. 

The objective of this merge is to return the name as well for this case.

## Checklist:
- [ ] My code passes all existing automated tests 
The code needs a modification in the automated tests (see in the commit)

### DIFF
--- a/includes/models/model.llms.post.instructors.php
+++ b/includes/models/model.llms.post.instructors.php
@@ -93,10 +93,12 @@ class LLMS_Post_Instructors {
 
 		// if empty, respond with the course author in an array
 		if ( ! $instructors ) {
+			$author = get_userdata( $this->post->get( 'author' ) );
 			$instructors = array(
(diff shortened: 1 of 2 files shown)


nicolas-jaussaud:
Hi @thomasplevy,

As requested, I recreated [this pull request](https://github.com/gocodebox/lifterlms/pull/929#issuecomment-651926228) here, for the dev branch.

thomasplevy:
Closing in favor of #1233 Thanks @nicolas-jaussaud!

PR #1242, opened 2020-07-17
Allow completion customization

thomasplevy:
## Description

The goal of this PR is to make it possible for a 3rd party to use the LifterLMS "progress" system (marking post/taxonomy objects as complete/incomplete) using our internal API and some hooks and filters.

This PR expands the extensibility of the API with a new filter and "DRYs" some functionality by using functions for determining whether or not an object is completeable (instead of having the filters hardcoded into the method).

In a forthcoming PR we will use these functions in the REST API to extend these functionality to the `/students/{id}/progress` endpoint.

## How has this been tested?

+ Existing unit tests
+ Manually tested mark complete / incomplete buttons on a test site

## Types of changes
+ Non-breaking

## Checklist:

### DIFF
--- a/includes/models/model.llms.student.php
+++ b/includes/models/model.llms.student.php
@@ -1817,7 +1817,7 @@ class LLMS_Student extends LLMS_Abstract_User_Data {
 		 * Lessons have binary completion (complete or incomplete).
 		 *
 		 * Other objects are dependent on their children's statuses. These other object types
-		 * must check the combined progress of their children to see if it's complete / incomplete
+		 * must check the combined progress of their children to see if it's complete / incomplete.
 		 */
 		$complete = ( 'lesson' === $object_type ) ? ( 'complete' === $status ) : ( 100 == $this->get_progress( $object_id, $object_type, false ) );
 
--- a/includes/models/model.llms.student.php
+++ b/includes/models/model.llms.student.php
@@ -24,6 +24,7 @@ defined( 'ABSPATH' ) || exit;
  * @since 3.37.9 Added filters `llms_user_enrollment_allowed_post_types` & `llms_user_enrollment_status_allowed_post_types` which allow 3rd parties to enroll users into additional post types via core enrollment methods.
  * @since 4.0.0 Remove previously deprecated methods.
  * @since [version] The `$enrollment_trigger` parameter was added to the `'llms_user_enrollment_deleted'` action hook.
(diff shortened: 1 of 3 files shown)


includes/models/model.llms.student.php
```diff
@@ -1767,83 +1768,115 @@ public function delete_enrollment( $product_id, $trigger = 'any' ) {
...
-		$parent_type = false;
-
-		// lessons are complete / incomplete automatically
-		// other object types are dependent on their children's statuses
-		// so the other object types need to check progress manually (bypassing cache) to see if it's complete / incomplete
+		/**
+		 * Lessons have binary completion (complete or incomplete).
+		 *
+		 * Other objects are dependent on their children's statuses. These other object types
+		 * must check the combined progress of their children to see if it's complete / incomplete
```
eri-trabiccolo:
```suggestion
		 * must check the combined progress of their children to see if it's complete / incomplete.
```

includes/models/model.llms.student.php
```diff
@@ -1856,41 +1889,60 @@ private function update_completion_status( $status, $object_id, $object_type, $t
...
+			 * @param int    $student_id  WP_User ID of the student.
+			 * @param int    $object_id   WP_Post ID of the object.
+			 * @param string $object_type The type of object. A lesson, section, course, or course_track.
+			 * @param string $trigger     String describing the reason for the status change.
 			 */
-			do_action( 'after_llms_mark_' . $status, $this->get_id(), $object_id, $object_type, $trigger );
+			do_action( "after_llms_mark_{$status}", $student_id, $object_id, $object_type, $trigger );
 
-		}// End if().
+		}
```
eri-trabiccolo, line 1945: Do you think adding to all the action hooks also `$this` instance maybe would be more helpful for 3rd parties?
I mean yeah they can always create a new instance of an LLMS_Student starting from the `$student_id` if needed but you know...

eri-trabiccolo (changes requested):

includes/models/model.llms.student.php
thomasplevy: since historically the actions didn't reference objects I don't know if it's essential to add another object to a set of triggers that already have a large number of arguments. Yes, it could be helpful but it's honestly not essential. Anyone hooking into the action has a at most 3 lines of code to instantiate (and check against a falsy).

I think we're going to leave it as is.

PR #1244, opened 2020-07-20
Reduce complexity of LLMS_Student::delete_enrollment()

thomasplevy:
Quick refactor of the method to improve code quality issues as reported by codeclimate.

### DIFF
--- a/includes/models/model.llms.student.php
+++ b/includes/models/model.llms.student.php
@@ -1710,7 +1710,7 @@ class LLMS_Student extends LLMS_Abstract_User_Data {
 
 			$delete = true;
 
-		} else if ( ! $enrollment_trigger ) {
+		} elseif ( ! $enrollment_trigger ) {
 
 			/**
 			 * Customize the behavior of enrollment deletion for "legacy" orders.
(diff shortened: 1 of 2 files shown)


eri-trabiccolo:
@thomasplevy 

I think the first commit (https://github.com/gocodebox/lifterlms/pull/1244/commits/5a1529d5e973f6370a106809a908eb376c61b67f) shouldn't be part of this PR :D 
The code looks good to me, but consider that I originally followed the structure of the `enroll` and the `unenroll` methods, then yes I complicated it a little bit more in https://github.com/gocodebox/lifterlms/commit/26b33bd3a84be85e7c1ae9dd22091415fcc40d8b
I think the reason why the `enroll` and `unenroll` methods, and hence this method too, were more complex was _maybe_ to avoid  the call to `get_enrollment_trigger()` (so a potentially useless db query) method when looking for **'any'** trigger?!
Anyways, if we don't care about it this PR looks good.  But, even if I think it doesn't introduce any drawbacks, I haven't tested it and since I'm tired I prefer to take another look at it tomorrow. <3

thomasplevy:
@eri-trabiccolo doing what I've done is never really

> I think the reason why the `enroll` and `unenroll` methods, and hence this method too, were more complex was _maybe_ to avoid the call to `get_enrollment_trigger()` (so a potentially useless db query) method when looking for **'any'** trigger?!

You're correct but I still wanted to fix this to fix issues reported by CC on the dev branch.

I believe that you are correct about avoiding the call to get the enrollment trigger when it's not necessary but this code is a bit simpler this way and I think it's okay to have that call unnecessarily -- we *still* want the trigger when running the action so having it early in favor of fewer conditions i think is okay. 

I messed up my commit history and I'm not going to rebase, just going to merge this in, have a look again tomorrow (dev isn't going live yet) and let me know if you see any pressing issues.

PR #1263, opened 2020-07-29
Refactor access-related functions

### LINKED ISSUE #1262: Search is accessible when site is restricted to a membership
### Reproduction Steps

+ Create a membership
+ Set the site to be restricted to this membership via the "Restrict site by membership level" global setting
+ As a logged out user visit the search page for the site (append ?s=whatever to the site's home url)

### Expected Behavior

+ Visitor will be served the sitewide restriction notice / redirect / etc...

### Actual Behavior

+ User can see search results

### Error Messages / Logs

N/A

<details>

```
LifterLMS 4.3.0, this will likely happen on 3.16 or later (possibly earlier)
```
</details>

This issue has be recreated:
+ [x] Locally
+ [ ] On a staging site
+ [ ] On a production website

### Additional Information and Context

I believe this is a bug because we have explicit logic in place to *allow* searching when a sitewide restriction is NOT enabled

https://github.com/gocodebox/lifterlms/blob/64622439a618aca8be8829eaf499a413bac317c1/includes/functions/llms.functions.access.php#L54-L56

However there's no inverse logic to serve a restriction when searching and sitewide membership *is* enabled.

thomasplevy:
## Description

The goal of this PR is to allow usage of `llms_page_restricted()` usable outside of the main WP loop. It relies on `is_singular()` and that means that in order to check post restrictions using this function in contexts like a REST request we need to mock the `$wp_query` with fake data to make this work.

In order to make this function generally useable as a way to check page restrictions I've rewritten the conditions and removed the usage of `is_singular()`. There are still references to `is_search()` and `is_home()` which are needed to restrict these areas when a sitewide membership is enabled.

This change did cause a test to fail but on review of the test the test was basically broken anyway so that test has been adjusted.

I have manually reviewed various restriction scenarios to ensure this doesn't break anything. It doesn't, as far as I can see.

Changes:

+ Refactored the `llms_page_restricted()` method and some related methods.
+ Improved documentation in the file.

**Added caching** -- I noticed while testing this that `llms_page_restricted()` runs 3 times per page load. I attempted to determine if we can remove the triplicate calls and it looks like we do actually need this information in a few different places. In order to improve the performance I've added a caching layer to the function. This will ensure that while the function may be called multiple times per page load, it only needs to run once (the first time). There is a concern in my mind that adding a caching layer could cause **more** caching issues than we already have in LifterLMS. This would be *dangerous* if for example a site was persisting the object cache. In this scenario we'd need to also purge page restriction data when enrollment changes. This is maybe something that should be added anyway (purging of this cache during enrollment) but in reality when enrollments happen the page reloads in all scenarios that I can find so this shouldn't be required unless there's a persistent cache.

I added a simple flag that can be used to kill the caching layer globally: `add_filter( 'llms_page_restricted_disable_caching', '__return_true' );` -- this can be used to help alleviate a caching issue quickly if we encounter one.

This is an area of preemptive concern in my mind. I think we want to improve performance even if that may mean creating caching issues for a subset of sites that add additional caching layers. 

Added a lot of tests

Fixes #1262 

## How has this been tested?

+ Existing tests 
+ New tests (written before refactoring to improve overall coverage and ensure that refactor maintains all previous functionality and conditions)

## Types of changes

+ Code quality improvements (I hope)
+ Bug fix

## Checklist:


thomasplevy:
@eri-trabiccolo I think this is ready for a careful review, please have at it when available.

includes/functions/llms.functions.access.php
```diff
@@ -5,24 +5,56 @@
...
  * @since 1.0.0
  * @since 3.16.11 Unknown.
+ * @since [version] Major refactor; added caching mechanisms.
  *
- * @param int      $post_id WordPress Post ID of the content.
- * @param int|null $user_id Optional. WP User ID (will use get_current_user_id() if none supplied). Default `null`.
- * @return array Restriction check result data.
+ * @param int      $post_id   WordPress Post ID of the content.
+ * @param int|null $user_id   Optional. WP User ID. Defaults to the current user if none supplied.
+ * @param bool     $use_cache If `true`, uses data stored in object cache (where available).
```
eri-trabiccolo:
```suggestion
 * @param bool     $use_cache Optional. If `true`, uses data stored in object cache (where available). Default `true`. 
```

includes/functions/llms.functions.access.php
```diff
@@ -5,24 +5,56 @@
...
+	 *
+	 * @since [version]
+	 *
+	 * @param bool $disable If `true`, caching will be disabled regardless of the value of the function's `$use_cache` parameter.
+	 */
+	$use_cache = apply_filters( 'llms_page_restricted_disable_caching', false ) ? false : $use_cache;
+
+	$cache_key = sprintf( '%1$d::%2$d', $post_id, $user_id );
+
+	// If we are using caching
```
eri-trabiccolo:
```suggestion
	// If we are using caching.
```

includes/functions/llms.functions.access.php
```diff
@@ -31,149 +63,196 @@ function llms_page_restricted( $post_id, $user_id = null ) {
...
+ */
+function llms_get_post_membership_restrictions( $post_id ) {
+
+	$memberships = array();
+
+	/**
+	 * Filter the post types which cannot be restricted to a membership.
+	 *
+	 * These LifterLMS core post types are restricted via enrollment into that
+	 * post (or it's parent post) directly so there won't be any related
```
eri-trabiccolo:
```suggestion
	 * post (or its parent post) directly so there won't be any related
```

includes/functions/llms.functions.access.php
```diff
@@ -310,121 +389,123 @@ function llms_is_page_restricted( $post_id, $user_id = null ) {
...
 		$quiz      = llms_get_post( $post_id );
-		$lesson_id = $quiz->get( 'lesson_id' );
-		if ( ! $lesson_id ) {
-			return false;
-		}
-	} else {
-		// dont pass other post types in here dumb dumb.
-		return false;
+		$lesson_id = $quiz ? $quiz->get( 'lesson_id' ) : false;
 	}
```
eri-trabiccolo: what about?
```php
	switch ( get_post_type( $post_id ) ) {
		case 'lesson':
			$lesson_id = $post_id;
			break;
		case 'llms_quiz':
			$quiz      = llms_get_post( $post_id );
			$lesson_id = $quiz ? $quiz->get( 'lesson_id' ) : false;
			break;
		default:
			$lesson_id = false;
	}
```

includes/functions/llms.functions.access.php
```diff
@@ -310,121 +389,123 @@ function llms_is_page_restricted( $post_id, $user_id = null ) {
...
 
+	/**
+	 * Customize the restriction information for a post's drip restrictions
+	 *
+	 * @since [version]
+	 *
+	 * @param int|bool $restriction False if the lesson is available.
+	 *                              WP Post ID of the lesson if it is not.
+	 * @param int      $post_id WP Post ID of a lesson or quiz.
+	 * @param int|null $user_id Optional. WP User ID (will use get_current_user_id() if none supplied). Default `null`.
```
eri-trabiccolo:
```suggestion
	 * @param int      $post_id     WP Post ID of a lesson or quiz.
	 * @param int|null $user_id     WP User ID or `null` if none supplied to the function.
```

includes/functions/llms.functions.access.php
```diff
@@ -310,121 +389,123 @@ function llms_is_page_restricted( $post_id, $user_id = null ) {
  * @since 3.0.0
  * @since 3.16.11 Unknown.
  * @since 3.37.10 Use strict comparison '===' in place of '=='.
+ * @since [version] Refactored and added filter on return value.
  *
  * @param int      $post_id WP Post ID of a lesson or quiz.
  * @param int|null $user_id Optional. WP User ID (will use get_current_user_id() if none supplied). Default `null`.
```
eri-trabiccolo:
```suggestion
 * @param int|null $user_id Optional. WP User ID. Default `null`.
```

includes/functions/llms.functions.access.php
```diff
@@ -310,121 +389,123 @@ function llms_is_page_restricted( $post_id, $user_id = null ) {
...
- * @return array|false False if the post is not restricted or the user has completed the prereq
- *                     associative array with prereq type and prereq id
- *                     array(
- *                         type => [course|course_track|lesson]
- *                         id => int (object id)
- *                     ).
+ * @return array|bool Returns `false` if the post is not restricted or the user has completed all applicable prerequisites,
+ *                     when at least one incomplete prerequisite has been found, returns an associative array describing the
+ *                     prerequisite type and object id.
```
eri-trabiccolo:
```suggestion
 *                    when at least one incomplete prerequisite has been found, returns an associative array describing the
 *                    prerequisite type and object id.
```

includes/functions/llms.functions.access.php
```diff
@@ -310,121 +389,123 @@ function llms_is_page_restricted( $post_id, $user_id = null ) {
...
+	if ( 'llms_quiz' === $post_type ) {
 		$quiz      = llms_get_post( $post_id );
-		$lesson_id = $quiz->get( 'lesson_id' );
-		if ( ! $lesson_id ) {
-			return false;
-		}
-	} else {
-		return false;
+		$lesson_id = $quiz ? $quiz->get( 'lesson_id' ) : false;
 	}
```
eri-trabiccolo: This is exactly the same piece of code I suggested above to refactor with a switch-case construct.
Either you like the switch-case refactoring or you don't (but secretly you love it), we can move the whole block in another function which would have an horrible name like `llms_get_lesson_id_from_lesson_or_quiz( $lesson_or_quiz_id )`.

includes/functions/llms.functions.access.php
```diff
@@ -310,121 +389,123 @@ function llms_is_page_restricted( $post_id, $user_id = null ) {
...
+	/**
+	 * Filter restriction data determining if a post is restricted by a prerequisite
+	 *
+	 * @since [version]
+	 *
+	 * @param array|bool $restriction Restriction result: `false` if the post is not restricted or the user has
+	 *                                completed all applicable prerequisites, when at least one incomplete prerequisite
+	 *                                has been found, returns an associative array describing the prerequisite type and object id.
+	 * @param int        $post_id     WP Post ID of a lesson or quiz.
+	 * @param int|null   $user_id     Optional. WP User ID (will use get_current_user_id() if none supplied). Default `null`.
```
eri-trabiccolo:
```suggestion
	 * @param int|null   $user_id     WP User ID or `null` if none supplied to the function.
```

includes/functions/llms.functions.access.php
```diff
@@ -433,50 +514,34 @@ function llms_is_post_restricted_by_prerequisite( $post_id, $user_id = null ) {
...
-		$course_id = $lesson->get_parent_course();
-
-	} elseif ( 'llms_quiz' === $post_type ) {
-		$quiz      = llms_get_post( $post_id );
-		$lesson_id = $quiz->get( 'lesson_id' );
-		if ( ! $lesson_id ) {
-			return false;
-		}
-		$lesson = llms_get_post( $lesson_id );
-		if ( ! $lesson_id ) {
```
eri-trabiccolo, line 459: ohh this was very nice :D

includes/functions/llms.functions.access.php
```diff
@@ -496,67 +561,46 @@ function llms_is_post_restricted_by_time_period( $post_id, $user_id = null ) {
...
-		// if no user, return the first membership from the array as the restriction id.
-		if ( ! $user_id ) {
-
-			$restriction_id = array_shift( $memberships );
+	$memberships = llms_get_post_membership_restrictions( $post_id );
+	if ( $memberships ) {
 
+		$student = $user_id ? llms_get_student( $user_id ) : false;
+		if ( ! $student ) {
+			$restriction = array_shift( $memberships );
```
eri-trabiccolo: golf playing: since we're not going to do anything else with the `$memberships` array, why don't we use `reset( $memberships )` or even `$memberships[0]`?

includes/functions/llms.functions.access.php
```diff
@@ -578,50 +624,60 @@ function llms_is_post_restricted_by_membership( $post_id, $user_id = null ) {
...
+		 * additional IDs are added to the array as integers, not numeric strings!
+		 *
+		 * @since Unknown
+		 *
+		 * @param int[] $bypass_ids Array of WP_Post IDs.
+		 */
+		$allowed = apply_filters( 'lifterlms_sitewide_restriction_bypass_ids', $bypass_ids );
 
-		return false;
+		$restriction = in_array( $post_id, $allowed, true ) ? false : $membership_id;
```
eri-trabiccolo:
```suggestion
		$restriction = in_array( absint( $post_id ), $allowed, true ) ? false : $membership_id;
```

includes/functions/llms.functions.access.php
```diff
@@ -5,24 +5,56 @@
...
+	$use_cache = apply_filters( 'llms_page_restricted_disable_caching', false ) ? false : $use_cache;
+
+	$cache_key = sprintf( '%1$d::%2$d', $post_id, $user_id );
+
+	// If we are using caching
+	$cached = $use_cache && $user_id ? wp_cache_get( $cache_key, 'llms_page_restricted' ) : false;
+
+	// Return early if we have cached data & cached is enabled.
+	if ( $cached ) {
+		return $cached;
```
eri-trabiccolo: I don't know, we don't run the filter on the cached result?
The `'llms_page_restricted`' filter is the one used by the view manager to grant access to the administrators...
I know it's an edge case but if do something like this:
```
add_action(
	'init',
	function() {
		llms_page_restricted( '123' );
	},
       9
);
```
and you visit a (non free, non orphan) lesson with id 123 as admin, you'll be treated as a visitor, because the view manager filter that grants access adds the filter at init|10, but with the code above, we cache the restriction at init 9 and never filter it.
Not sure if it's better to filter the cached value as well, or, alternatively, make sure that the cache is clean when the view manager adds its filter...

includes/functions/llms.functions.access.php
```diff
@@ -31,149 +63,196 @@ function llms_page_restricted( $post_id, $user_id = null ) {
...
 	$sitewide_membership_id = llms_is_post_restricted_by_sitewide_membership( $post_id, $user_id );
-	$membership_id          = llms_is_post_restricted_by_membership( $post_id, $user_id );
-
-	if ( is_home() && $sitewide_membership_id ) {
-		$restriction_id = $sitewide_membership_id;
-		$reason         = 'sitewide_membership';
-		// if it's a search page and the site isn't restricted to a membership bypass restrictions.
-	} elseif ( ( is_search() ) && ! get_option( 'lifterlms_membership_required', '' ) ) {
-		return apply_filters( 'llms_page_restricted', $results, $post_id );
-	} elseif ( is_singular() && $sitewide_membership_id ) {
```
eri-trabiccolo, line 57: from here on this PR introduces a regression (bug) by removing the checks on `is_singular()`.
Because on `template_include` we run this function, see: https://github.com/gocodebox/lifterlms/blob/4.3.0/includes/class.llms.template.loader.php#L101

and you know what `get_the_ID()` returns on archives at `template_loader` => the first post ID of the list.
You can easily reproduce this bug:
1) Create a post A and assign it the newly created category Cat_1
2) Create a post B and assign it the category Cat_1, then limit this post to a membership.
3) Visit category Cat_1 archive page
You'll see no list and the restriction message. This is not what we want (although, we probably might want to filter the content/excerpt of the restricted post in archives/search as well, but out of the scope of this PR).

What can we do?
Probably we shouldn't check against `get_the_ID()` in the template loader class, but against `get_queried_object_id()`, what do you think?
We should stress out these edge cases.

eri-trabiccolo (changes requested):
There are minor ones and a more important one, about the removal of `is_singular()`.

includes/functions/llms.functions.access.php
thomasplevy: this is a good point. I initially ran the filter on the cached result and the initial (non-cached) result but then I changed it thinking "The filter doesn't need to run every time".

But what happens if a result is filtered, cached, and then filtered again? I suppose whoever is adding a filter just needs to know. We could also pass an extra value with the filter "$cached" to let the develoepr know whether they're getting a result from the cache or a fresh one, right? And if they need to disregard the cache and regen they could.

I think you're right, let's filter both of them

includes/functions/llms.functions.access.php
thomasplevy: this introduces an interesting (related?) issue... We're passing in a $post_id but we're actually checking a "page's" (as in a generated screen's) not a post's restrictions.

In this scenario the queried object ID would be a category ID, right?

If we pass that in we should be adjusting the signature to accept an object id instead of a post id. And then, we need another argument to pass in the object type because we may have a category that shares an id with a post.

This would make the method more robust but it would also constitute some backwards incompatibilities, I think.

Maybe instead of doing this we can leave it as it is and add taxonomy archive checks before the "singular" post checks at the end...

Before this change what happens in your example? (I know I can load it up but I haven't yet)

includes/functions/llms.functions.access.php
thomasplevy: yea, this does belong in a utility function with a gross name. Good call <3

includes/functions/llms.functions.access.php
eri-trabiccolo: yeah I thought about adding another parameter as well for that filter to inform that's a cached version. Nice!

includes/functions/llms.functions.access.php
thomasplevy: right?

includes/functions/llms.functions.access.php
thomasplevy: is `reset()` faster than `array_shift()`? Or is `$memberships[0]` faster? As far as I can tell all 3 will do the same thing. What's the advantage to one the other over the other?

includes/functions/llms.functions.access.php
eri-trabiccolo: > this introduces an interesting (related?) issue... We're passing in a $post_id but we're actually checking a "page's" (as in a generated screen's) not a post's restrictions.

yeah 

> In this scenario the queried object ID would be a category ID, right?

right

> If we pass that in we should be adjusting the signature to accept an object id instead of a post id. And then, we need another argument to pass in the object type because we may have a category that shares an id with a post.

yeah, why not?

> This would make the method more robust but it would also constitute some backwards incompatibilities, I think.

Not sure it will, because it basically doesn't change the behavior... Before this PR categories were not affected by a restriction if not the sitewide one, or similar, for all the others there was the check on `is_singular()` that excluded them.

> Maybe instead of doing this we can leave it as it is and add taxonomy archive checks before the "singular" post checks at the end...
> 
well yeah we need something like that. You know that the nature of the current context is always tricky in wp, because of plugin's query alterations... it might really break our _compatibility balls_ :D 
You know if `! is_singular()` we don't need to know the `id` of the current object. If we pass a falsy post_id (which is not optional), we should only run those checks on the sitewide membership restriction, right?

> Before this change what happens in your example? (I know I can load it up but I haven't yet)

Well the category displayed all the posts...

includes/functions/llms.functions.access.php
eri-trabiccolo: I think, but I have no benchmarks, that `array_shift() `is slower that `reset()` because PHP doc say that `array_shift()` also performs a `reset()` on the resulting array.
https://www.php.net/manual/en/function.array-shift.php
But maybe on small arrays is not appreciable.
I also think that, since I don't need to reset a pointer of an array nor I need to have a smaller array but just get the first element of it `$memberships[0]` is more appropriate. :)

includes/functions/llms.functions.access.php
eri-trabiccolo: eheh was stupid because we checked again `$lesson_id` while there we probably wanted to check `$lesson` that we just retrieved :D

includes/functions/llms.functions.access.php
thomasplevy: FINE. <3

thomasplevy: This is called a "heated agreement"

includes/functions/llms.functions.access.php
thomasplevy: Now that i'm thinking about it catalogs are *not restricted* by design when sitewide membership is enabled.

So the real issue (I realize this now) is that we are restricting something incorrectly. `is_singular()` ensures that we don't run these checks against these "screens"

So I think we should change the behavior so that we do not run the formerly `is_singular()` checks when viewing archives of any kind.

eri-trabiccolo: Oh shi... you're right about the catalogs... Yeah but why?
Well anyways, at this point, if we don't want catalogs to be restricted we just can stop calling `llms_page_restricted()` in  https://github.com/gocodebox/lifterlms/blob/4.3.0/includes/class.llms.template.loader.php#L101
right?

thomasplevy: > Yeah but why?

Since courses and memberships themselves bypass sitewide membership restrictions, so should the catalogs. The idea is that this is how a user determines what is available to them if they join the membership that provides them with access to the content.

If memberships cannot be found/discovered then it's impossible to join the closed site at all.

There have been a few requests to lock down *everything* including these areas of the site but not very many. I would chalk this up to the fact that this is a pretty misunderstood feature of LifterLMS in general (and one that I wish we never added as it's a niche feature that generally confuses users instead of empowers them to solve a problem).

In any event, whether or not this is the *correct* behavior is a different conversation. It is, however, the current behavior and I'd like to maintain that.

eri-trabiccolo: I'm sorry, I thought that you were using "catalogs" in a more abstract way to identify any kind of archives, because we don't apply restrictions to any archives.

So yes I understood why our memberships and courses catalogs where not restricted, and my "but why?" question was more about other archives. And wanted to understand this to see whether doing what I said above made any sense:

> Well anyways, at this point, if we don't want catalogs to be restricted we just can stop calling llms_page_restricted() in https://github.com/gocodebox/lifterlms/blob/4.3.0/includes/class.llms.template.loader.php#L101
> right?

And for catalogs I wrongly meant all kind of archives.

thomasplevy: @eri-trabiccolo 

After reviewing this closer to try to fix the regression, I've found a ton of inconsistencies in the rule(s) here with regards to sitewide memberships...

I think we can take this as an opportunity to possibly right some broken behavior and possibly make a new option alongside the sitewide membership restriction option to allow end users to customize this behavior.

My thinking is as follows, when a sitewide membership is enabled:

+ All site content (excluding the exceptions listed [here](https://github.com/gocodebox/lifterlms/blob/b54afbd510c7b1c6ad0465671d569e9044e42221/includes/functions/llms.functions.access.php#L630-L638) should be restricted. This includes the site's homepage, blog page, any post type archives, etc... Courses are restricted. Additionally, the membership catalog page should also be restricted. The exception pages exist to facilitate user self-onboarding onto the site. Users can view the (restricted) view of the membership and pages that support enrollment into that membership.
+ Add a textarea field where users can enter a list of URLs that they wish to add to this bypass list. Users can add one url per line. These URLs will be automatically excluded from restrictions.

The end result will be what I believe is supposed to be the current behavior but is not working properly (both without this PR and with it).

Thoughts?

eri-trabiccolo: I'm with you.

PR #1270, opened 2020-08-04
update default images, per #1081

### LINKED ISSUE #1081: Update default images

thomasplevy:
# Description

+ Replaces default images for certificate background & achievement
+ Replaces default course/membership placeholder image
+ Adds source SVG files to repository in the `.source/` directory.

Fixes #1081 

## How has this been tested?

+ No testing required really

## Screenshots 

https://github.com/gocodebox/lifterlms/blob/be99a7eaa669ed91067cb2be185a332c4b9342e5/assets/images/optional_achievement.png
https://github.com/gocodebox/lifterlms/blob/be99a7eaa669ed91067cb2be185a332c4b9342e5/assets/images/optional_certificate.png
https://github.com/gocodebox/lifterlms/blob/be99a7eaa669ed91067cb2be185a332c4b9342e5/assets/images/placeholder.png

## Types of changes

Updated static image assets

## Checklist:


thomasplevy:
Opening this for design feedback on updated images.

eri-trabiccolo (commented):
The new certificate image has a different aspect ratio compared to the old one. The certificate image width/height are by default (settings that impact on how they'll be displayed) 800/616.

The new achievement image is bigger than the old one. Which isn't bad per se but when displayed in the notification widget in front, it takes far more space, maybe too much!?

thomasplevy:
@eri-trabiccolo I changed the aspect ratio because users are quite literally always yelling that the aspect ration doesn't make sense. I do NOT KNOW where 800/616 came from. I don't know if it's any actual legitimate aspect ratio but the updated ratio is A4 in landscape. I KNOW that a4 is not the most common WORLDWIDE but it makes more sense than an aspect ratio that doesn't exist.

I don't see any danger in changing this, but maybe I'm being short-sighted. Do you see an issue in changing the ratio?

As far as the achievement I made it bigger because the current default gets stretched out and looks pixelated in common scenarios. I'll have a look at the achievement within a notification and add some CSS to ensure it's not too big in that context.

eri-trabiccolo:
@thomasplevy 
> @eri-trabiccolo I changed the aspect ratio because users are quite literally always yelling that the aspect ration doesn't make sense. I do NOT KNOW where 800/616 came from. I don't know if it's any actual legitimate aspect ratio but the updated ratio is A4 in landscape. I KNOW that a4 is not the most common WORLDWIDE but it makes more sense than an aspect ratio that doesn't exist.

I agree with you, but we use the default values to write the wrapper's inline CSS here:
https://github.com/gocodebox/lifterlms/blob/4.3.0/templates/content-certificate.php#L22
(that will also set a "maximum width" for the certificate when printing it)
and we use the height as negative margin bottom for the certificate image element here:
https://github.com/gocodebox/lifterlms/blob/4.3.0/templates/content-certificate.php#L23

we might want to change the default placeholder's values at least:
https://github.com/gocodebox/lifterlms/blob/4.3.0/includes/functions/llms.functions.certificate.php#L68-L69
What do you think?

> I don't see any danger in changing this, but maybe I'm being short-sighted. Do you see an issue in changing the ratio?

No, I don't see any _danger_
I don't really see any big issue, just that the wrapper, using the default values, will have a bigger height, so leaving space below the image, if we don't change the default placeholder values (see the two filters above).

> As far as the achievement I made it bigger because the current default gets stretched out and looks pixelated in common scenarios. I'll have a look at the achievement within a notification and add some CSS to ensure it's not too big in that context.

ok.

chrisbadgett:
This is the one that got the most votes by the community
LifterLMS Certificate Background Default

chrisbadgett:
Also here is the default LifterLMS Achievement Badge 
LifterLMS Default Achievement Badge

chrisbadgett:
Updated certificate border that goes right to the edge to allow correct printing on background without border overlap 

LifterLMS Default Certiificate Background

PR #1271, opened 2020-08-04
Remove usage of deprecated jQuery.on( 'ready', ... )

### LINKED ISSUE #1268: JQMIGRATE: 'ready' event is deprecated
### Reproduction Steps

+ Open browser's console in front

### Expected Behavior

+ No warnings logged

### Actual Behavior

+ Warning logged

### Error Messages / Logs

```
JQMIGRATE: 'ready' event is deprecated
migrateWarn @ jquery-migrate.js?ver=1.4.1:45

```

```
LifterLMS 4.3.0
```

This issue has be recreated:
+ [x] Locally
+ [x] On a staging site
+ [x] On a production website

n/a

### What to do
First of all let's look at the jQuery doc:
https://api.jquery.com/ready/

> 
> jQuery offers several ways to attach a function that will run when the DOM is ready. All of the following syntaxes are equivalent:
> 
```
$( handler )
$( document ).ready( handler )
$( "document" ).ready( handler )
$( "img" ).ready( handler )
$().ready( handler )
```
> As of jQuery 3.0, only the first syntax is recommended; the other syntaxes still work but are deprecated. This is because the selection has no bearing on the behavior of the .ready() method, which is inefficient and can lead to incorrect assumptions about the method's behavior. For example, the third syntax works with "document" which selects nothing. The fourth syntax waits for the document to be ready but implies (incorrectly) that it waits for images to become ready.
> 
> There is also $(document).on( "ready", handler ), deprecated as of jQuery 1.8 and removed in jQuery 3.0. Note that if the DOM becomes ready before this event is attached, the handler will not be executed.

Now considering that the built-in WordPress jQuery version is 1.12 (at the moment), I don't think we should really worry about what will happen in 3.0, but since we're here, we could just replace, occurrences of deprecated once with:
`$(handler)
`

thomasplevy:
## Description

Fixes #1268 

## How has this been tested?

Manual tests of affected screens (student dashboard)

## Types of changes

Removes usage of deprecated jQuery syntax

## Checklist:

### DIFF
--- a/assets/js/app/llms-achievements.js
+++ b/assets/js/app/llms-achievements.js
@@ -18,9 +18,10 @@ LLMS.Achievements = {
 	 */
 	init: function() {
 
-		var self = this;
-
 		if ( $( '.llms-achievement' ) ) {
+
+			var self = this;
+
 			$( function() {
 				self.bind();
 				self.maybe_open();
--- a/assets/js/app/llms-achievements.js
+++ b/assets/js/app/llms-achievements.js
@@ -3,8 +3,8 @@
  *
  * @package LifterLMS/Scripts
  *
- * @since    3.14.0
- * @version  3.14.0
+ * @since 3.14.0
+ * @version [version]
  */
 
 LLMS.Achievements = {
@@ -12,9 +12,9 @@ LLMS.Achievements = {
 	/**
 	 * Init
 	 *
-	 * @return   void
-	 * @since    3.14.0
-	 * @version  3.14.0
+	 * @since 3.14.0
+	 *
+	 * @return void
 	 */
 	init: function() {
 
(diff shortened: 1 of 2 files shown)


eri-trabiccolo (approved):
I approve this because the original aim was to fix `jQuery( document ).on('ready',`
but I see there are other occurrences of `jQuery( document ).ready(` in our codebase, either in js (well in llms-metabox-certificate.js that you're going to remove in #1272) and php files.

thomasplevy:
Yes. This fixes the immediate problem that will be encountered in WP Core 5.6 when jQuery is upgraded to 3.x since this handler was *removed* in 1.8 and will no longer function in 3.x. The other handlers are deprecated (but will still work -- although it seems they will not work "as expected" so those will be fixed accordingly but that "fix" got out of hand which is why I'm separating the PRs.

PR #1282, opened 2020-08-07
exclude 'protected' post types from wp 5.5. xml sitemap

### LINKED ISSUE #1281: Inaccessible post types should not be listend in the WP 5.5 sitemap.xml file
### Reproduction Steps

+ On WP 5.5 visit wp-sitemap.xml

### Expected Behavior

+ "Protected" post types (lessons, quizzes, etc...) are not exposed in the sitemap

### Actual Behavior

+ They are

### Error Messages / Logs

N/A

<details>

```
WP 5.5-rc.2
LLMS latest

```

</details>

thomasplevy:
## Description

Fixes #1281 

## How has this been tested?

Manually (use recreation steps)
Added new unit test for added method

## Types of changes

5.5 compatibility fix

## Checklist:

### DIFF
--- a/includes/class.llms.post-types.php
+++ b/includes/class.llms.post-types.php
@@ -5,7 +5,7 @@
  * @package  LifterLMS\Classes
  *
  * @since 1.0.0
- * @version 3.37.12
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -26,6 +26,7 @@ class LLMS_Post_Types {
 	 *
 	 * @since 1.0.0
 	 * @since 3.0.4 Unknown.
+	 * @since [version] Add filter to deregister protected post types.
 	 *
 	 * @return void
 	 */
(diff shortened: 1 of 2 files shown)


includes/class.llms.post-types.php
```diff
@@ -102,6 +105,27 @@ public static function add_thumbnail_support() {
...
+	/**
+	 * De-register protected post types from wp-sitemap.xml
+	 *
+	 * @since [version]
+	 *
+	 * @param WP_Post_Type[] $post_types Array of post types.
+	 * @return WP_Post_Type[]
+	 */
+	public static function deregister_sitemap_post_types( $post_types ) {
+
```
eri-trabiccolo, line 117: can we add a filter here to which we pass the list of post types to be ignored, so that from our add-ons we use our own filter?
I like it more. :)

eri-trabiccolo (changes requested):

includes/class.llms.post-types.php
thomasplevy: I thought about that but like... does that save us any work, really? I've checked our other add-ons and we don't actually have an issue with any of our other add-ons, except groups, which actually requires a more complicated and conditional solution based on the way groups work (see: https://github.com/gocodebox/lifterlms-groups/issues/126)

eri-trabiccolo: I see, yeah pa posts are not public(ly queryable) I see.
Well I was even thinking that maybe someone would have lessons available in sitemaps? Why? yeah good question, free lessons?
But yeah, nevermind we'll add it later if requested.

eri-trabiccolo (approved):

PR #1298, opened 2020-08-19
Removes "duplicate" logic from `llms_deprecated_function()` in favor of core method

thomasplevy:
## Description

The `llms_deprecated_function()` method originally copied the logic of the WP core method `_deprecated_function()`. This function is prefixed with an underscore and __previously__ came with a note that it was for internal use only. This language has been removed (I don't know when) but it looks like it's "acceptable" for plugins to use this function now.

As a citation, WC uses the function: https://github.com/woocommerce/woocommerce/blob/0699022a46c4750e0b2574de9fccc85795e8e332/includes/wc-deprecated-functions.php#L54

Our function has always logged to the `llms-$hash.log` file in favor of the wp debug.log file. In order to maintain this behavior, this update uses `ini_get()` and `ini_set()` to log to our log file when a deprecated function call is made.

I do not see any downside to this change but the behavior *does change* slightly. Previously no error was triggered by our deprecated functions and now an error *will* be triggered. There is some "danger" in this change but assuming we don't trigger any deprecated functions with our own code we should be okay. Prior to release we should ensure that no add-ons are creating any deprecated warning or logs.

The main motivation for this PR is to make it possible to use the `@expectedDeprecated` annotation provided by the WP Core's phpunit testing lib. This will allow us to leave tests for deprecated functions intact instead of removing them.

## How has this been tested?

+ Manually tested
+ Added new unit tests

## Types of changes

+ Developer improvement

## Checklist:

### DIFF
--- a/includes/llms.functions.core.php
+++ b/includes/llms.functions.core.php
@@ -115,47 +115,24 @@ if ( ! function_exists( 'llms_content' ) ) {
 /**
  * Provide deprecation warnings
  *
- * Very similar to https://developer.wordpress.org/reference/functions/_deprecated_function/
+ * This function uses WP core's `_deprecated_function()`, logging to the LifterLMS log file
+ * located at `wp-content/updloads/llms-logs/llms-{$hash}.log` instead of `wp-content/debug.log`.
  *
- * @param   string $function    name of the deprecated class or function
- * @param   string $version     version deprecation occurred
- * @param   string $replacement function to use in it's place (optional)
- * @return  void
- * @since   2.6.0
- * @version 3.6.0
+ * @since 2.6.0
+ * @since 3.6.0 Unknown.
+ * @since [version] Uses WP `_deprecated_function()` instead of duplicating it's logic.
+ *
+ * @param string $function    Name of the deprecated function.
+ * @param string $version     LifterLMS version that deprecated the function.
+ * @param string $replacement Replacement function.
+ * @return void
  */
 function llms_deprecated_function( $function, $version, $replacement = null ) {
 
-	// only warn if debug is enabled
-	if ( ! defined( 'WP_DEBUG' ) || ! WP_DEBUG ) {
-		return;
-	}
-
-	if ( function_exists( '__' ) ) {
-
-		if ( ! is_null( $replacement ) ) {
-			$string = sprintf( __( '%1$s is <strong>deprecated</strong> since version %2$s! Use %3$s instead.', 'lifterlms' ), $function, $version, $replacement );
-		} else {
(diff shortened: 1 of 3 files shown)


thomasplevy:
@eri-trabiccolo this is submitted as a PR to the `asset-handler` branch because this "fixes" broken tests on the branch. However, I wanted this to exist as a separate PR to be reviewed on it's own. It's possible that I'm missing something huge or obvious by making this change and I want your thoughts on it before committing to it too thoroughly.

includes/llms.functions.core.php
```diff
@@ -115,47 +115,24 @@ function llms_content( $content = '' ) {
 /**
  * Provide deprecation warnings
  *
- * Very similar to https://developer.wordpress.org/reference/functions/_deprecated_function/
+ * This function uses WP core's `_deprecated_function()`, logging to the LifterLMS log file
+ * located at `wp-content/updloads/llms-logs/llms-{$hash}.log` instead of `wp-content/debug.log`.
```
eri-trabiccolo, line 119: I was wondering if we could take this as an occasion to stop logging deprecation warnings in llms-log.
As developer I wouldn't want to go look into any plugin log file for that kind of stuff. What do you think?
But anyways, for the purpose of this PR it's fine so.

tests/phpunit/unit-tests/functions/class-llms-test-functions-core.php
```diff
@@ -73,6 +73,54 @@ public function test_llms_assoc_array_insert() {
...
 
+	/**
+	 * [test_llms_deprecated_function description]
+	 *
+	 * @since [version]
+	 *
+	 * @expectedDeprecated DEPRECATED
+	 *
+	 * @return [type] [description]
+	 */
```
eri-trabiccolo: placeholders

eri-trabiccolo (approved):
I made a comment, and actually requested a change in the tests where you left the placeholders in a function documentation block. I'm going to approve this anyways.

includes/llms.functions.core.php
thomasplevy: Good point, I think you're right and realistically I was thinking we shouldn't change because of a "bacwards compat" issue but it's actually not a breaking change unless you consider "breaking" to mean "I have to look where I'd expect to look as a new dev and now it's there instead of somewhere else"

It's possible that a veteran might be slightly confused but a veteran is likely looking at debug.log too so there's nothing actually broken.

Updated and pushed.

tests/phpunit/unit-tests/functions/class-llms-test-functions-core.php
thomasplevy: woops

PR #1324, opened 2020-09-09
Attempt to sideload images during post generation

### LINKED ISSUE #1196: Importer does not import images from the post_content or custom fields (except featured images)
### Reproduction Steps

+ Export the course

### Expected Behavior

### Actual Behavior

### Error Messages / Logs

N/A

thomasplevy:
## Description

During course imports via the `LLMS_Generator`, parses the post_content of courses, lessons, quizzes, and access plans and sideload any `` tags found within the content.

Parses imported image question choice types and sideloads these images into the site as well.

Fixes #1196 

## How has this been tested?

+ Added new unit tests
+ Existing unit tests
+ Manually tested with course export containing images in post content of course, lesson, quiz, and picture choices

## Types of changes

+ Update

## Checklist:


thomasplevy:
@eri-trabiccolo for review when you return

includes/class.llms.generator.php
```diff
@@ -860,7 +1008,17 @@ public function get_author_id_from_raw( $raw, $fallback_author_id = null ) {
 	 * @return string
 	 */
 	public function get_default_post_status() {
+
+		/**
+		 * Filter the default status used for generating posts
+		 *
+		 * @since 3.7.3
+		 *
+		 * @param string         $post_status The default post status
```
eri-trabiccolo, line 1017:
```suggestion
		 * @param string         $post_status The default post status.
```

includes/class.llms.generator.php
```diff
@@ -41,6 +42,20 @@ class LLMS_Generator {
 	 */
 	private $generator = '';
 
+	/**
+	 * Array of images that have been sideloaded during generation
+	 *
+	 * Each array key will be the original source URL and the array value will be the new
+	 * URL of the image that has been sideloaded into the current site.
```
eri-trabiccolo, line 49: the array value is the attachment id.

includes/class.llms.generator.php
```diff
@@ -41,6 +42,20 @@ class LLMS_Generator {
...
+	 *
+	 * This array is checked prior to sideloading an image to ensure that if the same image is
+	 * used multiple times throughout an import, the image is only sideloaded a single time.
+	 *
+	 * @var string[]
```
eri-trabiccolo, line 54: This is actually an associative array.

includes/class.llms.generator.php
```diff
@@ -1188,6 +1435,83 @@ public function set_generator( $generator = null ) {
...
+	 * @return null|boolean Returns `true` on success, `false` if there were no images to update, or `null` if sideloading is disabled.
+	 */
+	protected function sideload_images( $post ) {
+
+		// Sideloading is disabled.
+		if ( ! $this->is_image_sideloading_enabled() ) {
+			return null;
+		}
+
+		$content  = $post->post->post_content;
```
eri-trabiccolo, line 1487:
```suggestion
		$content  = $post->get( 'content', true );
```
just to use our own api :D

includes/class.llms.generator.php
```diff
@@ -1079,6 +1244,53 @@ protected function is_generator_valid( $generator ) {
...
+	 */
+	public function is_image_sideloading_enabled() {
+
+		/**
+		 * Filter the status of image sideloading for the generator.
+		 *
+		 * @since [version]
+		 *
+		 * @param boolean $enabled Whether or not sideloading is enabled.
+		 */
```
eri-trabiccolo, line 1262:
```suggestion
		 * @param boolean        $enabled   Whether or not sideloading is enabled.
		 * @param LLMS_Generator $generator Generator instance.
		 */
```

eri-trabiccolo (changes requested):
Aside from the minor (doc related) changes requested, I want to add some considerations.
I understand the main purpose of this PR, and for that very case (sample course import) works great, but I think we're missing something here.

1) I'm on site A and I'm going to export a course, that will be imported in the site B.
In my course's content there's an external image, e.g. it's located on amazon.
With this PR, when importing the course on site B that external image will be downloaded and added to the media library of site B.
I'm not sure this is what we want... 
(Note: the same will happen when cloning a course on the same site: cloning a lesson L whose content contains an external image will result in the external image downloaded and added to the media library).

Solution: We should reverse the approach here, e.g. we can parse the post content when "exporting" rather than on "importing", and create a _media_ field: an array of all the images (media in general) that should be downloaded when importing. In this case external images will not be added to this array.
(When cloning a post, on the other hand, we could just skip this array, but see the point 2 below for another consideration).

2) We know that this same code is used also when cloning our posts. At the moment we're not duplicating same site images, this might be technically _not perfect_. The thing is that we're not (and actually never did) taking into account "the post attachments", that as we know can have only one parent. So a clone of a lesson L1 that has M1 and M2 media attachments - images in its content, will be cloned to L2 which will still show M1 and M2 in the content but, of course, they won't be attached to it.

(weakly) related to 2) but not strictly to this PR, we always clone the post featured image, while I think we should do it only if the featured image is an attachment too, as in WP is totally possible to use the same image as featured image for many posts. This is actually something reported by one of our users too, now that I think about it:  HS-128125

includes/class.llms.generator.php
thomasplevy: Fine.

thomasplevy:
This PR has been integrated into #1341

Issue #946, opened 2019-09-24
Coding Standards Updates Mega Issue


thomasplevy:
We currently exclude quite a few coding standards rules. Most of these rules are related to docblock and comment styling. Some require updates to code and tests. This issue is a mega issue to track all of the updates that should be made.

To see files that need updates, check the [phpcs.xml](https://github.com/gocodebox/lifterlms/blob/master/phpcs.xml) file and delete/comment out the corresponding rule. Then run `composer run-script check-cs-errors`.

When submitting pull requests to work on parts of this issue please submit PRs for no more than 15 files per PR. That means if you wish to update 100 files you'll need to submit 7 different PRs.

In the PR please Add `Per #945` to the PR description to reference (but not close) this issue.

When making updates please adhere to our [coding](https://github.com/gocodebox/lifterlms/blob/master/docs/coding-standards.md) and [documentation](https://github.com/gocodebox/lifterlms/blob/master/docs/documentation-standards.md) standards.

- [ ] Add translation placeholder comments, currently excluded by `<exclude name="WordPress.WP.I18n.MissingTranslatorsComment" />`
- [ ] Inline parameter comments should end in a full stop, excluded by `<exclude name="Squiz.Commenting.InlineComment.InvalidEndChar" />`
- [ ] Function parameter comments should end in a full stop, excluded by `<exclude name="Squiz.Commenting.FunctionComment.ParamCommentFullStop" />`
- [ ] Coment descriptions should start with a capital letter: `<exclude name="Generic.Commenting.DocComment.LongNotCapital" />` and `<exclude name="Generic.Commenting.DocComment.ShortNotCapital" />`

- [ ] File header comments should obey [documentation standards](https://github.com/gocodebox/lifterlms/blob/master/docs/documentation-standards.md#file-headers-1):
		`<exclude name="LifterLMS.Commenting.FileComment.PackageTagOrder" />`
		`<exclude name="LifterLMS.Commenting.FileComment.SinceTagOrder" />`
		`<exclude name="LifterLMS.Commenting.FileComment.MissingSinceTag" />`
		`<exclude name="LifterLMS.Commenting.FileComment.MissingVersionTag" />`
		`<exclude name="LifterLMS.Commenting.FileComment.VersionTagOrder" />`
		`<exclude name="LifterLMS.Commenting.FileComment.MissingPackageTag" />`
		`<exclude name="LifterLMS.Commenting.FileComment.DuplicateSinceTag" />`
[...]

PR #1328, opened 2020-09-16
add a feature for student to make his certificate link public to can share it

alaa-alshamy:
## Description

This PR, makes a feature for student to make his certificate link public to can share it with others

- Ability to make the certificate link shareable with others or not

Fixes #

## How has this been tested?

- Manually

## Checklist:

### DIFF
--- a/includes/controllers/class.llms.controller.certificates.php
+++ b/includes/controllers/class.llms.controller.certificates.php
@@ -124,12 +124,12 @@ class LLMS_Controller_Certificates {
 	}
 
 	/**
-	 * Change shareable settings of a certificate
+	 * Change shareable settings of a certificate.
 	 *
 	 * @since [version]
 	 *
-	 * @param int  $cert_id WP Post ID of the llms_my_certificate
-	 * @param bool $is_allowed Allow share the certificate or not
+	 * @param int  $cert_id WP Post ID of the llms_my_certificate.
+	 * @param bool $is_allowed Allow share the certificate or not.
 	 * @return void
 	 */
 	private function change_sharing_settings( $cert_id, $is_allowed ) {
@@ -140,7 +140,7 @@ class LLMS_Controller_Certificates {
 			return;
 		}
 
-		update_post_meta( $cert_id, '_llms_allow_sharing', $is_allowed ? 'yes' : 'no' );
+		$cert->set( 'allow_sharing', $is_allowed ? 'yes' : 'no' );
 
 	}
 
--- a/includes/controllers/class.llms.controller.certificates.php
+++ b/includes/controllers/class.llms.controller.certificates.php
@@ -117,7 +117,31 @@ class LLMS_Controller_Certificates {
 			$this->download( $cert_id );
 		} elseif ( isset( $_POST['llms_delete_cert'] ) ) {
 			$this->delete( $cert_id );
+		} elseif ( isset( $_POST['llms_change_shareable_cert'] ) ) {
+			$this->change_shareable_settings( $cert_id, isset( $_POST['llms_is_shareable_cert'] ) );
 		}
+
+	}
+
+	/**
+	 * Change shareable settings of a certificate
+	 *
+	 * @since 4.0.0
+	 *
+	 * @param int $cert_id WP Post ID of the llms_my_certificate
+	 * @param bool $is_allow Allow share the certificate or not
+	 * @return void
+	 */
+	private function change_shareable_settings( $cert_id, $is_allow ) {
+
+		$uid = get_current_user_id();
+		$cert = new LLMS_User_Certificate( $cert_id );
+
+		if ( $uid != $cert->get_user_id() && ! is_admin() ) {
+			return;
+		}
+
+		update_post_meta($cert_id, '_llms_allow_sharing', $is_allow ? 'yes' : 'no');
 
 	}
 
--- a/includes/controllers/class.llms.controller.certificates.php
+++ b/includes/controllers/class.llms.controller.certificates.php
@@ -117,8 +117,8 @@ class LLMS_Controller_Certificates {
 			$this->download( $cert_id );
 		} elseif ( isset( $_POST['llms_delete_cert'] ) ) {
 			$this->delete( $cert_id );
-		} elseif ( isset( $_POST['llms_change_shareable_cert'] ) ) {
-			$this->change_shareable_settings( $cert_id, isset( $_POST['llms_is_shareable_cert'] ) );
+		} elseif ( isset( $_POST['llms_enable_cert_sharing'] ) || isset( $_POST['llms_disable_cert_sharing'] ) ) {
+			$this->change_sharing_settings( $cert_id, isset( $_POST['llms_enable_cert_sharing'] ) );
 		}
 
 	}
@@ -126,22 +126,21 @@ class LLMS_Controller_Certificates {
 	/**
 	 * Change shareable settings of a certificate
 	 *
-	 * @since 4.0.0
+	 * @since [version]
 	 *
-	 * @param int $cert_id WP Post ID of the llms_my_certificate
-	 * @param bool $is_allow Allow share the certificate or not
+	 * @param int  $cert_id WP Post ID of the llms_my_certificate
+	 * @param bool $is_allowed Allow share the certificate or not
 	 * @return void
 	 */
-	private function change_shareable_settings( $cert_id, $is_allow ) {
+	private function change_sharing_settings( $cert_id, $is_allowed ) {
 
-		$uid = get_current_user_id();
 		$cert = new LLMS_User_Certificate( $cert_id );
 
-		if ( $uid != $cert->get_user_id() && ! is_admin() ) {
+		if ( ! $cert->can_user_manage() ) {
 			return;
 		}
 
-		update_post_meta($cert_id, '_llms_allow_sharing', $is_allow ? 'yes' : 'no');
+		update_post_meta( $cert_id, '_llms_allow_sharing', $is_allowed ? 'yes' : 'no' );
 
 	}
 
--- a/includes/controllers/class.llms.controller.certificates.php
+++ b/includes/controllers/class.llms.controller.certificates.php
@@ -117,8 +117,8 @@ class LLMS_Controller_Certificates {
 			$this->download( $cert_id );
 		} elseif ( isset( $_POST['llms_delete_cert'] ) ) {
 			$this->delete( $cert_id );
-		} elseif ( isset( $_POST['llms_enable_cert_sharing'] ) || isset( $_POST['llms_disable_cert_sharing'] ) ) {
-			$this->change_sharing_settings( $cert_id, isset( $_POST['llms_enable_cert_sharing'] ) );
+		} elseif ( isset( $_POST['llms_enable_cert_sharing'] ) ) {
+			$this->change_sharing_settings( $cert_id, (bool) $_POST['llms_enable_cert_sharing'] );
 		}
 
 	}
@@ -128,7 +128,7 @@ class LLMS_Controller_Certificates {
 	 *
 	 * @since [version]
 	 *
-	 * @param int  $cert_id WP Post ID of the llms_my_certificate.
+	 * @param int  $cert_id    WP Post ID of the llms_my_certificate.
 	 * @param bool $is_allowed Allow share the certificate or not.
 	 * @return void
 	 */
--- a/includes/models/model.llms.user.certificate.php
+++ b/includes/models/model.llms.user.certificate.php
@@ -26,6 +26,7 @@ class LLMS_User_Certificate extends LLMS_Post_Model {
 		'certificate_image'    => 'absint',
 		// 'certificate_content' => 'html', // use get( 'content' )
 		'certificate_template' => 'absint',
+		'allow_sharing'        => 'yesno',
 	);
 
 	/**
@@ -121,10 +122,19 @@ class LLMS_User_Certificate extends LLMS_Post_Model {
 	 */
 	public function can_user_manage( $user_id = null ) {
 
-		$user_id = $user_id ?? get_current_user_id();
+		$user_id = $user_id ? $user_id : get_current_user_id();
 		$result  = ( $user_id == $this->get_user_id() || llms_can_user_bypass_restrictions( $user_id ) );
 
-		return apply_filters( 'llms_certificate_can_user_manage', $result, $this );
+		/**
+		 * Filter whether or not a user can manage a given certificate.
+		 *
+		 * @since [version]
+		 *
+		 * @param boolean               $result      Whether or not the user can manage certificate.
+		 * @param int                   $user_id     WP_User ID of the user viewing the certificate.
+		 * @param LLMS_User_Certificate $certificate Certificate class instance.
+		 */
+		return apply_filters( 'llms_certificate_can_user_manage', $result, $user_id, $this );
 
 	}
 
@@ -138,22 +148,41 @@ class LLMS_User_Certificate extends LLMS_Post_Model {
 	 */
 	public function can_user_view( $user_id = null ) {
 
-		$user_id = $user_id ?? get_current_user_id();
-		$result  = $this->can_user_manage( $user_id ) || $this->is_sharing_allowed();
+		$user_id = $user_id ? $user_id : get_current_user_id();
+		$result  = $this->can_user_manage( $user_id ) || $this->is_sharing_enabled();
 
-		return apply_filters( 'llms_certificate_can_user_view', $result, $this );
+		/**
+		 * Filter whether or not a user can view a user's certificate.
+		 *
+		 * @since [version]
+		 *
+		 * @param boolean               $result      Whether or not the user can view the certificate.
+		 * @param int                   $user_id     WP_User ID of the user viewing the certificate.
+		 * @param LLMS_User_Certificate $certificate Certificate class instance.
+		 */
+		return apply_filters( 'llms_certificate_can_user_view', $result, $user_id, $this );
 
 	}
 
 	/**
-	 * Is sharing settings allowed
+	 * Is sharing enabled
 	 *
 	 * @since [version]
 	 *
 	 * @return bool
 	 */
-	public function is_sharing_allowed() {
-		return llms_parse_bool( $this->get( 'allow_sharing' ) );
+	public function is_sharing_enabled() {
+
+		/**
+		 * Filter whether or not sharing is enabled for a certificate.
+		 *
+		 * @since [version]
+		 *
+		 * @param boolean               $enabled     Whether or not sharing is enabled.
+		 * @param LLMS_User_Certificate $certificate Certificate class instance.
+		 */
+		return apply_filters( 'llms_certificate_is_sharing_enabled', llms_parse_bool( $this->get( 'allow_sharing' ) ), $this );
+
 	}
 
 }
--- a/includes/models/model.llms.user.certificate.php
+++ b/includes/models/model.llms.user.certificate.php
@@ -111,4 +111,49 @@ class LLMS_User_Certificate extends LLMS_Post_Model {
 		);
 	}
 
+	/**
+	 * Can user manage and make some actions on the certificate
+	 *
+	 * @since [version]
+	 *
+	 * @param int|null $user_id Optional. WP User ID (will use get_current_user_id() if none supplied). Default `null`.
+	 * @return bool
+	 */
+	public function can_user_manage( $user_id = null ) {
+
+		$user_id = $user_id ?? get_current_user_id();
+		$result  = ( $user_id == $this->get_user_id() || llms_can_user_bypass_restrictions( $user_id ) );
+
+		return apply_filters( 'llms_certificate_can_user_manage', $result, $this );
+
+	}
+
+	/**
+	 * Can user view the certificate
+	 *
+	 * @since [version]
+	 *
+	 * @param int|null $user_id Optional. WP User ID (will use get_current_user_id() if none supplied). Default `null`.
+	 * @return bool
+	 */
+	public function can_user_view( $user_id = null ) {
+
+		$user_id = $user_id ?? get_current_user_id();
+		$result  = $this->can_user_manage( $user_id ) || $this->is_sharing_allowed();
+
+		return apply_filters( 'llms_certificate_can_user_view', $result, $this );
+
+	}
+
+	/**
+	 * Is sharing settings allowed
+	 *
+	 * @since [version]
+	 *
+	 * @return bool
+	 */
+	public function is_sharing_allowed() {
+		return llms_parse_bool( $this->get( 'allow_sharing' ) );
+	}
+
 }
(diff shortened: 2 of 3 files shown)


thomasplevy:
@alaa-alshamy thanks so much for your contribution!

I've moved this from a PR against `trunk` to the `dev` branch

I'll be reviewing this in more detail later today but there's a few things you can get started on:

1) The Travis builds are failing due to coding standards violations. You can quickly fix these by running PHPCBF, please have a look at our [installing for development[(https://github.com/gocodebox/lifterlms/blob/trunk/docs/installing.md) guide, specifically the section [Running PHPCS](https://github.com/gocodebox/lifterlms/blob/trunk/docs/installing.md#running-phpcs)

2) There's no phpunit/integration tests included in your PR. That's *okay* but before this is merged and shipped __someone__ will have to write some tests. Writing tests is a little bit of a burden if you're not familiar so we're not going to force that on your shoulders -- unless you want to write them and are willing to learn!

There's a few areas I think we can improve this PR so I'll be leaving some more specific notes on the code later today.

Thanks again

includes/controllers/class.llms.controller.certificates.php
```diff
@@ -117,10 +117,34 @@ public function maybe_handle_reporting_actions() {
...
+		} elseif ( isset( $_POST['llms_change_shareable_cert'] ) ) {
+			$this->change_shareable_settings( $cert_id, isset( $_POST['llms_is_shareable_cert'] ) );
 		}
 
 	}
 
+	/**
+	 * Change shareable settings of a certificate
+	 *
+	 * @since 4.0.0
```
thomasplevy: Per doc standards: https://github.com/gocodebox/lifterlms/blob/trunk/docs/documentation-standards.md#using-placeholders

```suggestion
	 * @since [version]
```

includes/controllers/class.llms.controller.certificates.php
```diff
@@ -117,10 +117,34 @@ public function maybe_handle_reporting_actions() {
...
+	private function change_shareable_settings( $cert_id, $is_allow ) {
+
+		$uid = get_current_user_id();
+		$cert = new LLMS_User_Certificate( $cert_id );
+
+		if ( $uid != $cert->get_user_id() && ! is_admin() ) {
+			return;
+		}
+
+		update_post_meta($cert_id, '_llms_allow_sharing', $is_allow ? 'yes' : 'no');
```
thomasplevy: Please use "allowed" in favor of "allow" so the variables read more naturally. 

```suggestion
	 * @param bool $is_allowed Allow share the certificate or not
	 * @return void
	 */
	private function change_shareable_settings( $cert_id, $is_allowed ) {

		$uid = get_current_user_id();
		$cert = new LLMS_User_Certificate( $cert_id );

		if ( $uid != $cert->get_user_id() && ! is_admin() ) {
			return;
		}

		update_post_meta($cert_id, '_llms_allow_sharing', $is_allowed ? 'yes' : 'no');
```

templates/content-certificate.php
```diff
@@ -13,7 +13,11 @@
 $cert = new LLMS_User_Certificate( get_the_ID() );
 $uid  = get_current_user_id();
 
-if ( $uid != $cert->get_user_id() && ! llms_can_user_bypass_restrictions( $uid ) ) {
+$is_owner_or_admin 		= ($uid == $cert->get_user_id() || llms_can_user_bypass_restrictions( $uid ));
+$is_shareable					= (get_post_meta( get_the_ID(), '_llms_allow_sharing', true ) === 'yes');
+$is_allow_to_display	= $is_owner_or_admin || $is_shareable;
```
thomasplevy: I think this logic can be made more generally useful (possibly internally in the future but certainly for any custom development) by moving it into class methods on the `LLMS_User_Certificate` model.

I envision two methods (pseduo code below):

```php
public function can_user_manage( $user_id = null ) {
 
  $user_id = $user_id or current user

  return existing logic of  $is_owner_or_admin var in PR

}
```

and

```php
public function can_user_view( $user_id = null ) {
 
  $user_id = $user_id or current user
  
  return $this->can_user_manage() || llms_parse_bool( $this->get( 'allow_sharing' ) )

}
```

Both methods should additionally have filters on their returns!

templates/content-certificate.php
```diff
@@ -13,7 +13,11 @@
...
+
+if ( ! $is_allow_to_display ) {
```
thomasplevy: Assuming class methods are added (as noted earlier in review):

```suggestion
if ( ! $cert->can_user_view() ) {
```

templates/content-certificate.php
```diff
@@ -38,19 +42,29 @@
...
-		<?php echo _e( 'Print', 'lifterlms' ); ?>
-		<i class="fa fa-print" aria-hidden="true"></i>
-	</button>
-
-	<form action="" method="POST">
-		<button class="llms-button-secondary" type="submit" name="llms_generate_cert">
-			<?php echo _e( 'Save', 'lifterlms' ); ?>
-			<i class="fa fa-cloud-download" aria-hidden="true"></i>
+<?php
+if ( $is_owner_or_admin ) {
```
thomasplevy:
```suggestion
if ( $cert->can_user_manage() ) {
```

templates/content-certificate.php
```diff
@@ -38,19 +42,29 @@
...
-</div>
+		<form action="" method="POST">
+			<button class="llms-button-secondary" type="submit" name="llms_generate_cert">
+		  	<?php echo _e( 'Save', 'lifterlms' ); ?>
+				<i class="fa fa-cloud-download" aria-hidden="true"></i>
+			</button>
+
+			<?php echo _e( 'Make this link public', 'lifterlms' ); ?>
+			<input type="checkbox" name="llms_is_shareable_cert" <?php echo ( $is_shareable ? 'checked' : '' ); ?> onchange="document.getElementsByName('llms_change_shareable_cert')[0].click()" />
+			<button class="llms-button-secondary" style="visibility: hidden;" type="submit" name="llms_change_shareable_cert"></button>
```
thomasplevy: Something about using inline JS and a hidden button with no content feels "hacky" (for lack of a better word).

I think that this would *feel* better to me if there was a button that acted as a toggle depending on the current state of the certificate's sharing settings.

Something like:

```php
<button ...><?php echo llms_parse_bool( $cert->get( 'allow_sharing' ) ? 'Enable sharing' : 'Disable Sharing'; ?></button>
```

What do you think?

thomasplevy (changes requested):
@alaa-alshamy thank you for your hard work here.

I'm requesting some changes here to make the feature a bit more extensible and reusable by external devs and also improve the general internal useability.

The notes on using a button in favor of a checkbox come from a place of *my feelings* and don't have any standards or facts to back them up so I'm curious what your thoughts are on my thoughts on how to handle this. I'm not necessarily stating it *must* be as I've requested!

templates/content-certificate.php
alaa-alshamy: it's ok for me, but i thought the checkbox will be more obvious in favor of change the text of the button

and in button way i have to make tow different actions one for enable and one for disable the sharing settings

alaa-alshamy:
@thomasplevy Thanks for your suggestions, done take a look right now please

includes/models/model.llms.user.certificate.php
```diff
@@ -111,4 +111,49 @@ public function get_user_postmeta() {
...
+	 * Can user view the certificate
+	 *
+	 * @since [version]
+	 *
+	 * @param int|null $user_id Optional. WP User ID (will use get_current_user_id() if none supplied). Default `null`.
+	 * @return bool
+	 */
+	public function can_user_view( $user_id = null ) {
+
+		$user_id = $user_id ?? get_current_user_id();
```
thomasplevy: Use ternary in favor of null coalesce operator.

Why?

While I have been called a dinosaur for adhering to rules like this that exist to support unsupported versions of PHP, I prefer to stay out of fights and arguments about progressive PHP in favor of focusing on what I'm personally better at.

This is to say that currently, the WordPress core does not allow usage of the null coalesce operating so neither do we.

Source if interested: https://make.wordpress.org/core/2020/03/20/updating-the-coding-standards-for-modern-php/ see the "Operators" section

```suggestion
		$user_id = $user_id ? $user_id : get_current_user_id();
```

includes/controllers/class.llms.controller.certificates.php
```diff
@@ -117,10 +117,33 @@ public function maybe_handle_reporting_actions() {
...
+	 */
+	private function change_sharing_settings( $cert_id, $is_allowed ) {
+
+		$cert = new LLMS_User_Certificate( $cert_id );
+
+		if ( ! $cert->can_user_manage() ) {
+			return;
+		}
+
+		update_post_meta( $cert_id, '_llms_allow_sharing', $is_allowed ? 'yes' : 'no' );
```
thomasplevy: Can use certificate setter in favor of directly using postmeta apis (this will make this automatically take advantage of built-in LifterLMS hooks).

```suggestion
		$cert->set( 'allow_sharing', $is_allowed ? 'yes' : 'no' );
```

thomasplevy (changes requested):
@alaa-alshamy 

This looks really great!

Thanks for updating the CS issues and the changes I've suggested.

These are two minor changes requested to finalize this initial work. It looks very good though and thank you again.

In addition to these two inline comments we'd also need to add `allow_sharing` as a class property on the `LLMS_User_Certificate` model in the `$properties` class variable.

Since this is a boolean yes/no variable we need to add `allow_sharing => 'yesno'` to this array

thomasplevy:
@alaa-alshamy This looks very good save the 3 minor items noted in my latest review.

Would you like to take a stab at covering the new code with tests or should we take it from here?

includes/models/model.llms.user.certificate.php
```diff
@@ -111,4 +111,49 @@ public function get_user_postmeta() {
...
+		$result  = $this->can_user_manage( $user_id ) || $this->is_sharing_allowed();
+
+		return apply_filters( 'llms_certificate_can_user_view', $result, $this );
```
thomasplevy: Add `$user_id` to parameters passed to filter and include docblock for the new filter.

```suggestion
		/**
		 * Filter whether or not a user can view a user's certificate.
		 *
		 * @since [version]
		 *
		 * @param boolean               $result      Whether or not the user can view the certificate.
		 * @param int                   $user_id     WP_User ID of the user viewing the certificate.
		 * @param LLMS_User_Certificate $certificate Certificate class instance.
		 **/
		return apply_filters( 'llms_certificate_can_user_view', $result, $user_id, $this );
```

includes/models/model.llms.user.certificate.php
```diff
@@ -111,4 +111,49 @@ public function get_user_postmeta() {
...
+
+	/**
+	 * Is sharing settings allowed
+	 *
+	 * @since [version]
+	 *
+	 * @return bool
+	 */
+	public function is_sharing_allowed() {
+		return llms_parse_bool( $this->get( 'allow_sharing' ) );
```
thomasplevy:
```suggestion
		/**
		 * Filter whether or not sharing is enabled for a certificate
		 * 
		 * @since [version]
		 *
		 * @param boolean               $enabled     Whether or not sharing is enabled.
		 * @param LLMS_User_Certificate $certificate Certificate class instance.
		 */
		return apply_filters( 'llms_certificate_is_sharing_enabled', llms_parse_bool( $this->get( 'allow_sharing' ) ), $this );
```

includes/models/model.llms.user.certificate.php
```diff
@@ -111,4 +111,49 @@ public function get_user_postmeta() {
...
+	 * Can user manage and make some actions on the certificate
+	 *
+	 * @since [version]
+	 *
+	 * @param int|null $user_id Optional. WP User ID (will use get_current_user_id() if none supplied). Default `null`.
+	 * @return bool
+	 */
+	public function can_user_manage( $user_id = null ) {
+
+		$user_id = $user_id ?? get_current_user_id();
```
thomasplevy:
```suggestion
		$user_id = $user_id ? $user_id : get_current_user_id();
```

includes/models/model.llms.user.certificate.php
```diff
@@ -111,4 +111,49 @@ public function get_user_postmeta() {
...
+		$result  = ( $user_id == $this->get_user_id() || llms_can_user_bypass_restrictions( $user_id ) );
+
+		return apply_filters( 'llms_certificate_can_user_manage', $result, $this );
```
thomasplevy:
```suggestion
		/**
		 * Filter whether or not a user can manage a given certificate.
		 *
		 * @since [version]
		 *
		 * @param boolean               $result      Whether or not the user can manage certificate.
		 * @param int                   $user_id     WP_User ID of the user viewing the certificate.
		 * @param LLMS_User_Certificate $certificate Certificate class instance.
		 **/
		return apply_filters( 'llms_certificate_can_user_manage', $result, $user_id, $this );
```

thomasplevy (commented):
Added some parameters to filters, docblocks on the filters, and added a filter on the `is_sharing_allowed()` method.

includes/controllers/class.llms.controller.certificates.php
```diff
@@ -117,10 +117,33 @@ public function maybe_handle_reporting_actions() {
...
 
 	}
 
+	/**
+	 * Change shareable settings of a certificate
+	 *
+	 * @since [version]
+	 *
+	 * @param int  $cert_id WP Post ID of the llms_my_certificate
+	 * @param bool $is_allowed Allow share the certificate or not
```
eri-trabiccolo:
```suggestion
	 * @param int  $cert_id WP Post ID of the llms_my_certificate.
	 * @param bool $is_allowed Allow share the certificate or not.
```
Please bear with us, this will break a rule that is currently disabled:
https://github.com/gocodebox/lifterlms/blob/b681ffbd6b9b36450700383917d0d150982170a8/phpcs.xml#L30
We are slowly fixing these (tracked in #946)

includes/models/model.llms.user.certificate.php
```diff
@@ -111,4 +111,49 @@ public function get_user_postmeta() {
...
+
+		$user_id = $user_id ?? get_current_user_id();
+		$result  = $this->can_user_manage( $user_id ) || $this->is_sharing_allowed();
+
+		return apply_filters( 'llms_certificate_can_user_view', $result, $this );
+
+	}
+
+	/**
+	 * Is sharing settings allowed
```
eri-trabiccolo:
```suggestion
	 * Is sharing allowed
```

templates/content-certificate.php
```diff
@@ -38,19 +37,31 @@
...
+			</button>
 
-		<input type="hidden" name="certificate_id" value="<?php echo get_the_ID(); ?>">
-		<?php wp_nonce_field( 'llms-cert-actions', '_llms_cert_actions_nonce' ); ?>
-	</form>
-</div>
+			<button class="llms-button-secondary" type="submit" name="<?php echo $is_sharing_allowed ? 'llms_disable_cert_sharing' : 'llms_enable_cert_sharing'; ?>">
+		  <?php echo ( $is_sharing_allowed ? _e( 'Disable sharing', 'lifterlms' ) : _e( 'Enable sharing', 'lifterlms' ) ); ?>
+				<i class="fa fa-share-alt" aria-hidden="true"></i>
+			</button>
```
eri-trabiccolo: This template is shared both by `llms_my_certificate` and `llms_certificate` post types.
The first one is the post type to model the certificate that a user, who earned it, can see in his (e.g.) `www.example.what/dashboard/my-certificate/cert-title-{N}`, while the latter is basically used by _admins_ to view the certificate created at `www.example.what/certificate/cert-title`.

This button should be available only when viewing a `llms_my_certificate` post type.

@thomasplevy I was thinking that a new method could be added to the `LLMS_User_Certificate` class, along with the new `is_sharing_allowed()`, that could be called `is_sharing_possible()` and it could look like this:
```php
	/**
	 * Is sharing possible
	 *
	 * @since [version]
	 *
	 * @return bool
	 */
	public function is_sharing_possible() {
		/**
		 * Filter whether or not sharing is possible for a certificate
		 *
		 * @since [version]
		 *
		 * @param boolean               $is_possible Whether or not sharing is possible.
		 * @param LLMS_User_Certificate $certificate Certificate class instance.
		 */
		return apply_filters( 'llms_certificate_is_sharing_possible', get_post_type( $this->get( 'id' ) ) === $this->db_post_type, $this );
	}
```

so that in the template we'll have:
```suggestion
			<?php if ( $cert->is_sharing_possible() ) : ?>
				<button class="llms-button-secondary" type="submit" name="<?php echo $is_sharing_allowed ? 'llms_disable_cert_sharing' : 'llms_enable_cert_sharing'; ?>">
					<?php echo ( $is_sharing_allowed ? _e( 'Disable sharing', 'lifterlms' ) : _e( 'Enable sharing', 'lifterlms' ) ); ?>
					<i class="fa fa-share-alt" aria-hidden="true"></i>
				</button>
			<?php endif; ?>
```

eri-trabiccolo (changes requested):
Thanks @alaa-alshamy, this looks very solid to me too, save the few changes requested by @thomasplevy and my thoughts on adding the new `is_sharing_possible()` method.

includes/models/model.llms.user.certificate.php
alaa-alshamy: But they said:
> The spaceship <=>, null coalesce ?? and null coalesce equals ??= operators can not be used in WordPress Core **until the minimum PHP version has been raised to PHP 7.0**

and i noticed the min php version you support is 7.2, and that's why i used this operator so why not ?

anyway i will make as you want, it's not a big deal

templates/content-certificate.php
alaa-alshamy: @eri-trabiccolo You are right, but i think it's good idea for admin to see this button in preview like other buttons (Save, Print)

but we can make it useless, i mean if he clicked on it, it will not make anything, What do you think @eri-trabiccolo @thomasplevy  ?

eri-trabiccolo: @alaa-alshamy I see what you mean, but honestly don't like the idea of having a button in that view that would do nothing, but let's wait for @thomasplevy opinion.

thomasplevy: A button that doesn't work leads to support requests and angry, frustrated, consfused, etc... users.

I don't think we need another method here though. I think we can just run a check against the post type in the template.

includes/models/model.llms.user.certificate.php
thomasplevy: We adhere (mostly) to the WP core coding standards. The more exceptions we make the more complicated it becomes to maintain our own set of coding standards. We are a small team (unlike the WP core). I don't have the capacity to keep up with all the newest and most "golfy" operators and standards. So I simply adopt what the WP core does (even though we have different minimum version requirements).

Thanks for updating.

templates/content-certificate.php
eri-trabiccolo: You're right, I actually didn't want to create a method - it doesn't.really make a lot of sense having it - although I didn't want to have the logic in the template. But yeah let's go with the check in the template.

thomasplevy: I think this is an acceptable amount of "business logic" in a template.

alaa-alshamy: @thomasplevy @eri-trabiccolo Ok great, i will do that check

alaa-alshamy:
@thomasplevy @eri-trabiccolo Done take a look please

includes/controllers/class.llms.controller.certificates.php
```diff
@@ -117,10 +117,33 @@ public function maybe_handle_reporting_actions() {
...
 		}
 
 	}
 
+	/**
+	 * Change shareable settings of a certificate.
+	 *
+	 * @since [version]
+	 *
+	 * @param int  $cert_id WP Post ID of the llms_my_certificate.
```
eri-trabiccolo:
```suggestion
	 * @param int  $cert_id    WP Post ID of the llms_my_certificate.
```

includes/controllers/class.llms.controller.certificates.php
```diff
@@ -117,10 +117,33 @@ public function maybe_handle_reporting_actions() {
 			$this->download( $cert_id );
 		} elseif ( isset( $_POST['llms_delete_cert'] ) ) {
 			$this->delete( $cert_id );
+		} elseif ( isset( $_POST['llms_enable_cert_sharing'] ) || isset( $_POST['llms_disable_cert_sharing'] ) ) {
+			$this->change_sharing_settings( $cert_id, isset( $_POST['llms_enable_cert_sharing'] ) );
```
eri-trabiccolo:
```suggestion
		} elseif ( isset( $_POST['llms_enable_cert_sharing'] ) ) {
			$this->change_sharing_settings( $cert_id, (bool) $_POST['llms_enable_cert_sharing'] );
		}
```
Since `<button>`s can have a value why don't we use one var only?

templates/content-certificate.php
```diff
@@ -38,19 +37,32 @@
...
-		<button class="llms-button-secondary" type="submit" name="llms_generate_cert">
+		<form action="" method="POST">
+			<button class="llms-button-secondary" type="submit" name="llms_generate_cert">
 			<?php echo _e( 'Save', 'lifterlms' ); ?>
-			<i class="fa fa-cloud-download" aria-hidden="true"></i>
-		</button>
+				<i class="fa fa-cloud-download" aria-hidden="true"></i>
+			</button>
+			<?php if ( get_post_type( $cert->get( 'id' ) ) === $cert->get( 'db_post_type' ) ) : ?>
+				<button class="llms-button-secondary" type="submit" name="<?php echo $is_sharing_enabled ? 'llms_disable_cert_sharing' : 'llms_enable_cert_sharing'; ?>">
```
eri-trabiccolo:
```suggestion
				<button class="llms-button-secondary" type="submit" name="llms_enable_cert_sharing" value="<?php echo ! $is_sharing_allowed; ?>">
```
Since <button>s can have a value why don't we use one var only?

eri-trabiccolo (changes requested):
Minor improvement suggested (and a doc cs fix), what do you think?

templates/content-certificate.php
alaa-alshamy: @eri-trabiccolo Actually nothing, i liked the idea :) 

i will do it and test it

alaa-alshamy:
@eri-trabiccolo Done, take a look again please

eri-trabiccolo (approved):
Good for me @alaa-alshamy, thanks!

thomasplevy (approved):

thomasplevy:
@alaa-alshamy thanks for the contribution and all your hard work on this. We'll be leaving this as pending until either @eri-trabiccolo or myself have time to write tests for this. We're both currently working on different projects that need our attention. Hopefully we'll get tests written in the next few weeks and then merged in.

thomasplevy:
@alaa-alshamy and @eri-trabiccolo 

Question while reviewing this a bit further and writing some tests:

Currently, if an admin (or user with bypass abilities) views a student's earned certificate (not their own) they can enable/disabled sharing *for that user*.

I think this is correct behavior, an admin should be able to do this. Right?

eri-trabiccolo:
@thomasplevy 
Yeah, I think so - admins can do also nasty things.

thomasplevy:
@alaa-alshamy I'm moving this over to #1354 where I'm finishing this up (writing tests and the like).

Thanks so much for your contribution, looking forward to seeing your code again in the future!

PR #1337, opened 2020-09-17
Docblocks

pondermatic:
## Description
Improve DocBlock data types.

1. Use class name instead of "obj".
1. Use "class[]" instead of "array An array of classes."
1. Fix incorrect data types.

## Types of changes
Documentation fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/includes/class-llms-grades.php
+++ b/includes/class-llms-grades.php
@@ -60,8 +60,8 @@ class LLMS_Grades {
 	/**
 	 * Calculates the grades for elements that have a list of children which are averaged / weighted to come up with the total grade
 	 *
-	 * @param    array        $children  list of child objects
-	 * @param    LLMS_Student $student
+	 * @param    array        $children list of child objects
+	 * @param    LLMS_Student $student  A LLMS_Student object.
 	 * @return   float|null
 	 * @since    3.24.0
 	 * @version  3.24.0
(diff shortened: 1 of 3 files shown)


thomasplevy:
@pondermatic as always thanks for your diligence on these kinds of thankless tasks <3

The travis build is failing due to some spacing errors that can be cleaned up with `composer run fix-cs`

And... while it's obvious that using the classnames themselves as opposed to generic objects is favorable the changes made here will break a rule that is currently disabled due to tons of legacy files that need to be updated.

https://github.com/gocodebox/lifterlms/blob/trunk/phpcs.xml#L36

We are slowly fixing these (tracked in #946) but this PR will make that list of todos larger!

 The purpose of having both the type/class hint *and* a parameter description is perhaps redundant at a glance, especially for items like you've changed where the hint changes from `obj` to `LLMS_Course` and then the desrciption changes to "Course object." but for many instances the parameter name and hint does not sufficiently describe the parameter so the comment is essential. For the sake of consistency we don't want a developer to have to determine whether or not a comment is required so the comment is always required (even when redundant).

I know that our codebase breaks a lot of these rules but we're slowly working towards fixing all our own internal rule violations. Despite this, I don't want to accept a PR that fixes an issue and creates a new one that we'll have to solve eventually.

thomasplevy:
Thanks @pondermatic !

pondermatic:
No problem @thomasplevy . I understand and agree with your coding standards.

Issue #1326, opened 2020-09-15
Course importer does not allow usage of reusable blocks


thomasplevy:
### Reproduction Steps

+ Create a course that utilizes a reusable block editor block in course or lesson content
+ Export the course
+ Import the course into a different site

### Expected Behavior

+ The reusable block will be imported into the site with the course import

### Actual Behavior

+ The reusable block will render as "Unavailable" in the new site

### Additional details

A reusable block is stored as a `wp_block` post type, when used in a post's content, the reusable block is stored as a reference that looks like this:

```html
<!-- wp:block {"ref":562} /-->
```

The ref is a reference to the `wp_block` post's post ID.

### Proposed solution

When generating an export, post types that support the block editor should include a property of reusable blocks contained within the post's content That looks something like this:

```json
{
  "id": 123,
  "embeds": {
     "blocks": {
          "456": {
              "title": "Block Title",
              "content": "block content",
          }
      }
 }
}
```

During an import we can read all the reusable blocks included in an export and duplicate them into the import site, and, after importing it, we can update the reference(s) in the post's content.

This issue has be recreated:
+ [x] Locally
+ [x] On a staging site
+ [ ] On a production website
+ [x] With only LifterLMS and a default theme

thomasplevy:
@eri-trabiccolo I came across this bug today and in light of your comments regarding exporting images as an array in favor of parsing images (comments in review on #1324) I'm proposing that we add an "attachments" array member during exports. We can attach "images" here as well as reuseable blocks. Working on a WIP that we'll discuss tomorrow in greater detail.

eri-trabiccolo:
@thomasplevy ok, but I would use a different name since "attachment" has a meaning in wp, what do you think?

thomasplevy:
I was thinking "embeds" but that has a different meaning too. Attachments didn't feel quite right but I didn't spend too much more time thinking about it.

Any suggestions on what it should be called?

eri-trabiccolo:
@thomasplevy 
Mmm maybe we just should keep it simple and consistent, let's remove the 'embeds' and just go with first level:
'blocks'
'attachments'
'images' (which are not attachments, and, in fact I see it as an array of URLs). 
Otherwise, something like 'include/require'?

PR #1341, opened 2020-09-17
WIP: Export additional data (reusable blocks and images)

### LINKED ISSUE #1196: Importer does not import images from the post_content or custom fields (except featured images)

thomasplevy:
## Description

Per #1326 
Fixes #1196 

This PR adds new functionality to the `LLMS_Post_Model::toArray()` method.

During export (or usages of `toArray()`) "extra" data is included that is useful (or required) to fully import data into a different site.

## How has this been tested?

+ New and existing unit tests
+ Manually

## Types of changes

+ Bug fixes

## Checklist:


thomasplevy:
@eri-trabiccolo what do you think about this as an export strategy? If you think this works I'll start working on the import side of it

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -1175,6 +1178,21 @@ public function toArray() {
 			}
 		}
 
+		/**
+		 * Filter whether or not "extra" content should be included in the post array
+		 *
+		 * The dynamic portion of this hook, `$this->model_post_type`, refers to the model's post type. For example "course"m
```
```suggestion
		 * The dynamic portion of this hook, `$this->model_post_type`, refers to the model's post type. For example "course",
```

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -1188,24 +1206,135 @@ public function toArray() {
...
+	 * @return string[] Array of image source URLs.
+	 */
+	protected function to_array_extra_images( $content ) {
+
+		$images = array();
+		$dom    = llms_get_dom_document( $content );
+		if ( is_wp_error( $dom ) ) {
+			return $images;
+		}
+
```
eri-trabiccolo, line 1319: I would add some logic to flag an image as attachment. Why so?
Because I'm thinking about the cloning.
When cloning a lesson with an image in its content we could check if that image already exists, and sideload it only if it's an attachment.

eri-trabiccolo (changes requested):

eri-trabiccolo:
@thomasplevy This is basically what I had mind so, since long time ago I decided I always have to agree with myself I agree with this strategy!

includes/abstracts/abstract.llms.post.model.php
thomasplevy: My plan is to disable sideloading during cloning. EG: We'll only sideload / import blocks during a physical import, during a clone we'll skip the logic.

Do you think that covers it?

I guess there's still a use case where if someone exports and re-imports into their own site they would end up with duplicate images if we aren't running these kinds of checks. What do you think?

eri-trabiccolo: The thing is that I think that when cloning a lesson with an attachment, the new lesson should ideally have an attachment as well, and since attachments cannot be shared another one must be created.
Images that are not attachments can be skipped during a cloning yes.

The latter use case you said above can be solved by checking if the image you're going to import is already present (and in my idea this check would only be run if the image is not an attachment of the post), if so, skip it: this is rather simple because we have the full URL of the image to be imported. Am I missing anything?

thomasplevy: Okay I see what you're saying, yes, okay, you're right.

thomasplevy: @eri-trabiccolo I see a problem with this approach now that I'm working on implementation...

Are you suggesting that we *only import images specifically designated as an attachment to the current post*?

If we do this, then we'll actually miss a large number of images (potentially).

Let's say we create a course (Course A) and we upload an image (Image Z) to use in that course's content. The created attachment will be an attachment to that course.

If we create a second course (Course B) and we add Image Z (the attachment) to this new course. The attachment doesn't get created or updated.

If we export Course B and *only include attachment images* then when we import Course B the image will not be imported into the new site.

thomasplevy: If we clone a lesson and use the current method in this PR then the attachment would not be duplicated but it wouldn't *need to be* because if we use the same image in multiple lessons we don't create multiple attachments unless we physically upload the same image multiple times.

So an attachment isn't "shared" (excactly) but the WP media lib doesn't automatically create multiple attachments when using the same image in multiple posts.

eri-trabiccolo: @thomasplevy 
> Are you suggesting that we only import images specifically designated as an attachment to the current post?

Yes and no. 
I suggest to import (download and upload) only attachment images IF we're cloning or anyways importing from the same website. 
For images which are not attachments we just should check, on import, that they are local (URL is "local") and exist.

Is it more clear what I have in mind?

thomasplevy:
@eri-trabiccolo I think this is ready for a re-review if you have some time

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -1188,24 +1206,131 @@ public function toArray() {
 
 		ksort( $arr ); // Because i'm anal...
 
-		return apply_filters( 'llms_' . $this->model_post_type . '_to_array', $arr, $this );
+		/**
+		 * Filter the final post array created when converting the object to an array
+		 *
+		 * The dynamic portion of this hook, `$this->model_post_type`, refers to the model's post type. For example "course"m
```
eri-trabiccolo, line 1212:
```suggestion
		 * The dynamic portion of this hook, `$this->model_post_type`, refers to the model's post type. For example "course",
```

includes/class.llms.generator.php
```diff
@@ -313,6 +355,7 @@ private function generate_courses() {
 	 * @since 3.3.0
 	 * @since 3.7.3 Unknown.
 	 * @since 4.3.3 Use an empty string in favor of `null` for an empty `post_content` field.
+	 * @since [version] Sideload images attached to the post.
 	 *
 	 * @param array $raw                Raw Access Plan Data
```
eri-trabiccolo, line 360:
```suggestion
	 * @param array $raw                Raw Access Plan Data.
```

includes/class.llms.generator.php
```diff
@@ -647,12 +770,71 @@ private function create_question( $raw, $manager, $author_id ) {
...
+		$block_id = absint( $block_id );
+
+		// If the block already exists, don't create it again.
+		$existing = get_post( $block_id );
+		if ( $existing && 'wp_block' === $existing->post_type && $block['title'] === $existing->post_title && $block['content'] === $existing->post_content ) {
+			return false;
+		}
+
+		// Check if the block was previously imported.
+		$id = empty( $this->reusable_blocks[ $block_id ] ) ? false : $this->reusable_blocks[ $block_id ];
```
eri-trabiccolo, line 817: can't we switch the order of these two checks?
Check whether it was previously imported, if not, check if the block already exists... So we "avoid" a `get_post()` in that rare case the reusable block was already imported during the same importing session. What do you think?

includes/class.llms.generator.php
```diff
@@ -1079,6 +1346,75 @@ protected function is_generator_valid( $generator ) {
...
+		 *
+		 * @param boolean        $enabled   Whether or not sideloading is enabled.
+		 * @param LLMS_Generator $generator Generator instance.
+		 */
+		return apply_filters( 'llms_generator_is_image_sideloading_enabled', true, $this );
+
+	}
+
+	/**
+	 * Determines if reusable block importing is enabled generator
```
eri-trabiccolo, line 1371:
```suggestion
	 * Determines if reusable block importing is enabled for the generator
```

includes/class.llms.generator.php
```diff
@@ -1188,6 +1524,86 @@ public function set_generator( $generator = null ) {
...
+	/**
+	 * Sideload images found in a given post
+	 *
+	 * This attempts to sideload the `src` attribute of every <img> element
+	 * found in the `post_content` of the supplied post.
+	 *
+	 * @since [version]
+	 *
+	 * @param LLMS_Post_Model $post Post object.
+	 * @param array           $raw  Array of raw data
```
eri-trabiccolo, line 1567:
```suggestion
	 * @param array           $raw  Array of raw data.
```

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -1188,24 +1206,131 @@ public function toArray() {
...
+	 * an import these images will be imported into the new site via media sideloading.
+	 *
+	 * @since [version]
+	 *
+	 * @param array $arr Post array from `toArray()`.
+	 * @return array[]
+	 */
+	protected function to_array_extra( $arr ) {
+
+		$arr['_extra'] = array(
```
eri-trabiccolo, line 1257:
```suggestion
		$arr['_extras'] = array(
```
the generators look for '_extras' not for '_extra'

includes/class.llms.generator.php
```diff
@@ -1037,6 +1264,46 @@ private function handle_prerequisites() {
...
+
+				$new_src = $this->create_reusable_block( $block_id, $block );
+				if ( ! is_wp_error( $new_src ) ) {
+					$find[]    = $src;
+					$replace[] = $new_src;
+				}
+			}
+
+			if ( $find && $replace ) {
+				$content = str_replace( $find, $replace, $post->get( 'post_content', true ) );
```
eri-trabiccolo, line 1298:
```suggestion
				$content = str_replace( $find, $replace, $post->get( 'content', true ) );
```

includes/class.llms.generator.php
```diff
@@ -1188,6 +1524,86 @@ public function set_generator( $generator = null ) {
...
+
+				$new_src = $this->sideload_image( $post_id, $src );
+				if ( ! is_wp_error( $new_src ) ) {
+					$find[]    = $src;
+					$replace[] = $new_src;
+				}
+			}
+
+			if ( $find && $replace ) {
+				$content = str_replace( $find, $replace, $post->get( 'post_content', true ) );
```
eri-trabiccolo, line 1598:
```suggestion
				$content = str_replace( $find, $replace, $post->get( 'content', true ) );
```

tests/phpunit/unit-tests/class-llms-test-generator.php
```diff
@@ -8,9 +8,86 @@
...
+
+		$gen = new LLMS_Generator( array() );
+
+		$title   = 'Dupcheck reuse block';
+		$content = 'Block content';
+
+		$dup = $this->factory->post->create( array(
+			'post_type' => 'wp_block',
+			'post_title' => $title,
+			'post_content' => $content,
```
eri-trabiccolo, line 32:
```suggestion
			'post_content' => $content,
			'post_status'  => 'publish',
```
otherwise the block will be created as 'draft' and you'll just reference a block that technically "doesn't exist" (that's how gutenberg sees it).

eri-trabiccolo (changes requested):
There are a few important changes to be made because at this stage this just doesn't work.

includes/class.llms.generator.php
thomasplevy: Makes sense, yes.

thomasplevy:
Closing in favor of (more) complete PR #1386

PR #1353, opened 2020-09-24
Event open sessions handling - performance improvements (WIP)

### LINKED ISSUE #1349: Event session start not recorded on sign-on
### Reproduction Steps

+ Perform a login
+ Check the `wp_lifterlms_events` table in your db.

### Expected Behavior

I would expect to see two new records with approximately the same date, referring to the same user id:
+ a session start event
+ a sign-on event

### Actual Behavior

Only the sign-on event is recorded.

This is because the sign-on event is recorded when the `'wp_login'` action hook is fired, but then current user id is not set yet, so the session start doesn't occur because of this:
https://github.com/gocodebox/lifterlms/blob/4.4.4/includes/class-llms-sessions.php#L415-L418

The solution would be to pass the user_id parameter to `LLMS_Sessions::start()` method, which will continue to fall back on the current user id if the user_id is not provided.

### Error Messages / Logs

```
n/a

```

```

LifterLMS 4.4.4
WordPress 5.5.1
```

This issue has be recreated:
+ [x] Locally
+ [ ] On a staging site
+ [ ] On a production website

n/a

### LINKED ISSUE #1350: Event session end not recorded on sign-out
### Reproduction Steps

+ Perform a login
+ logout
+ Check the `wp_lifterlms_events` table in your db.

### Expected Behavior

+ I would expect to see the last record to be a session end event.

### Actual Behavior

+ No session end event recored

This is simply because we're comparing an `LLMS_Event` with a string here:
https://github.com/gocodebox/lifterlms/blob/4.4.4/includes/class-llms-events.php#L231

(due to the overwriting of `$event` variable:
https://github.com/gocodebox/lifterlms/blob/4.4.4/includes/class-llms-events.php#L221
 that started as string:
https://github.com/gocodebox/lifterlms/blob/4.4.4/includes/class-llms-events.php#L202)

### Error Messages / Logs

```
n/a

```

```

LifterLMS 4.4.4
WordPress 5.5.1
```

This issue has be recreated:
+ [x] Locally
+ [ ] On a staging site
+ [ ] On a production website

n/a

### LINKED ISSUE #1351: Event account sign-out recorded when visiting yoursite.domain/wp-admin/ as logged out user
### Reproduction Steps

+ Visit yoursite.domain/wp-admin/ as logged out user
+ You'll be redirected to yoursite.domain/wp-login.php?redirect_to=https%3A%2F%2Fyoursite.domain%2Fwp-admin%2F&reauth=1
+ Check the wp_lifterlms_events table in your db.

### Expected Behavior

+ I would expect to see no new record logger related to this action

### Actual Behavior

+ an account.signout event record has been logged with `user` 0 and `object_id` 0.

Solution:
check the current user id is not 0 here:
https://github.com/gocodebox/lifterlms/blob/4.4.4/includes/class-llms-events-core.php#L67
before recording.

### Error Messages / Logs

n/a

```
LifterLMS 4.4.4
WordPress 5.5.1
```

This issue has be recreated:
+ [x] Locally
+ [ ] On a staging site
+ [ ] On a production website

n/a

### LINKED ISSUE #1359: account.signon event not recorded when logging in or registering through our login/registration form.
### Reproduction Steps
see https://github.com/gocodebox/lifterlms/pull/1353#pullrequestreview-501469553

+ Log in through our login form
+ Check the db's table wp_lifterlms_events (1)
+ Log out and register as new user through or registration form, or either register directly at checkout buying a course
+ Check the db's table wp_lifterlms_events (2)

### Expected Behavior

+ I'd expect to see a new signon event recorded in the db (along with a session start event) both for cases (1) and (2)

### Actual Behavior

+ No signon events recorded (nor session start events)

The reason for (1) is that when using our login form we basically fire the `wp_login` action at `init|10`:
https://github.com/gocodebox/lifterlms/blob/4.4.4/includes/forms/controllers/class.llms.controller.login.php#L30
and only after that (but still at `init|10`) we register the events that must be recorded:
https://github.com/gocodebox/lifterlms/blob/4.4.4/includes/class-llms-events.php#L63

So the solution would be to register the events at `init|9` (tested it doesn't break the advanced video add-on).

About (2) the issue is that we don't trigger the `wp_login` action when registering:
https://github.com/gocodebox/lifterlms/blob/4.4.4/includes/class.llms.person.handler.php#L761
https://github.com/gocodebox/lifterlms/blob/4.4.4/includes/functions/llms.functions.person.php#L312

Now I'm not sure we should fire a new action hook either in `LLMS_Person_Handler::register()` or in `llms_set_person_auth_cookie()`, I think we could just explicitly record the event in https://github.com/gocodebox/lifterlms/blob/4.4.4/includes/class.llms.person.handler.php#L761

What do you think? 

### Error Messages / Logs

n/a

```
LifterLMS 4.4.4
WordPress 5.4.1
```

This issue has be recreated:
+ [x] Locally
+ [x] On a staging site
+ [x] On a production website

n/a

eri-trabiccolo:
## Description
The main purpose of this PR is to improve the perfomances when getting the open sessions to close (each 5 minutes with cron) on large websites.
The idea is to store the open sessions in a new table: `lifterlms_events_open_sessions`.
Meaning that, when opening a session we store the event, then create a new record with the session’s event `id` in this new table.
When ending a session we record this event in the events table, and remove the open session record from `lifterlms_events_open_sessions`: we always know the event id of the "starting" session:
https://github.com/gocodebox/lifterlms/blob/4.4.4/includes/class-llms-sessions.php#L119
This way the get_open_sessions() https://github.com/gocodebox/lifterlms/blob/4.4.4/includes/class-llms-sessions.php#L284 would just mean getting ids from a pretty small table (`lifterlms_events_open_sessions`).

It also fixes the few bugs I've discovered while working on this:
Fixes #1349 
Fixes #1350 
Fixes #1351
Fixes #1359 

----

This is **still WIP** as it misses the db update routine, but if you get the chance @thomasplevy I'd appreciate having your opinion on the few additions I did, if you think you would like to see additional tests on a specific area and so on.

## How has this been tested?
New and old unit tests.
But I haven't written any new test on the new open sessions handling. I relied on the old tests for the already existing methods which now leverage it.

## Screenshots 
n/a

## Types of changes

## Checklist:


includes/class-llms-sessions.php
```diff
@@ -131,14 +132,28 @@ protected function end( $start ) {
 			)
 		);
 
+		if ( ! is_wp_error( $end ) ) {
+			global $wpdb;
+			$wpdb->query( // phpcs:ignore: WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
```
thomasplevy: Just learned this cool shorthand implemented by the WPCS project to make ignoring these warnings shorter.

`WordPress.DB.DirectDatabaseQuery.DirectQuery` = `db call ok`
`WordPress.DB.DirectDatabaseQuery.NoCaching` = `no-cache ok`

```suggestion
			$wpdb->query( // db call ok; no-cache ok.
```

includes/class-llms-sessions.php
```diff
@@ -276,38 +299,33 @@ protected function get_last_session() {
...
 	 * @param int $limit Number of sessions to return.
-	 * @param int $skip Number of sessions to skip.
+	 * @param int $skip  Number of sessions to skip.
 	 * @return LLMS_Event[]
 	 */
 	protected function get_open_sessions( $limit = 50, $skip = 0 ) {
 
 		global $wpdb;
-		$sessions = $wpdb->get_col(
+		$sessions = $wpdb->get_col( // phpcs:ignore: WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
```
thomasplevy:
```suggestion
		$sessions = $wpdb->get_col( // db call ok; no-cache ok.
```

includes/class-llms-sessions.php
```diff
@@ -407,26 +430,44 @@ protected function get_new_id() {
...
-				'object_id'    => $this->get_new_id(),
+				'object_id'    => $this->get_new_id( $user_id ),
 				'event_type'   => 'session',
 				'event_action' => 'start',
 			)
 		);
 
+		if ( ! is_wp_error( $start ) ) {
+			global $wpdb;
+			$wpdb->query( // phpcs:ignore: WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
```
thomasplevy:
```suggestion
			$wpdb->query( // db call ok; no-cache ok.
```

thomasplevy (commented):
Looks good so far, minor updates related to PHPCS ignores (for your information)

thomasplevy:
@eri-trabiccolo are you ready for a review or is this still a WIP? (Not rushing, just making sure I'm not a roadblock)

Whether or not your asking for feedback, I'd like tests for the update functions. Thank you <3

eri-trabiccolo:
@thomasplevy 
Functionally it only misses the update function tests as you noted, and that's the reason why I hadn't asked for a review - I know they're important. <3

thomasplevy:
@eri-trabiccolo cool cool, no rush, just wanted to make sure I wasn't blocking!

DonavanMartin:
Do this pull request resolve SELECT events performance issues has showing on the following images or I need to open another issue? 
Capture d’écran, le 2020-09-28 à 21 53 27
Capture d’écran, le 2020-09-28 à 21 54 20

thomasplevy:
@DonavanMartin the queries we're reworking in this PR are *more than likely* the queries responsible for the purple spikes seen in these images.

It's hard to be 100% positive as these performance graphs don't exactly report what queries are running slowly. If you want 100% confirmation a slow query log would be more useful. If you have some more concrete data for us to look at, please post it.

We don't have an open issue for this particular PR but you're welcome to open an issue if you'd like us to investigate this further. But please provide some logs in favor of graphs!

Thanks

DonavanMartin:
@thomasplevy This image show the query:

Capture d’écran, le 2020-09-29 à 16 48 57

thomasplevy:
@DonavanMartin this is the query we're working on

Edit: Here's the most relevant part of the diff: https://github.com/gocodebox/lifterlms/pull/1353/files#diff-28c4ad2a7188dac0039c4e8363e4d28bL287-L301

thomasplevy:
@eri-trabiccolo I noticed you chose 4.4.5 for the update function prefixes, lets slate this for version 4.5.0 (barring anything crazy goes on before this is ready). I think this change warrants a minor bump.

eri-trabiccolo:
@thomasplevy when you have the chance can you take a look at this?

tests/phpunit/unit-tests/functions/updates/class-llms-test-functions-updates-450.php
```diff
@@ -0,0 +1,219 @@
...
+	public function test_migrate_events_open_sessions_pagination() {
+
+		// Create open session events.
+		$num_open_sessions = 250;
+		$open_session_ids  = $this->create_open_session_events( $num_open_sessions );
+
+		$loops = 1;
+		// Check how many times a the update function needs to run.
+		// Internally we fetch 200 sessions at time, we expect it to run the following number of times:
+		$expected_loops = (int) ceil( count( $open_session_ids ) / 200 ) + 1;
```
thomasplevy: Code golf:

```suggestion
		$expected_loops = 3;
```

thomasplevy (changes requested):
@eri-trabiccolo one minor (kind of joke) change requested below.

However, I think we need a bit more work here as this does not fix #1349

The tests pass, of course, but I'm not seeing the sign-on event recorded when creating a new account or logging into an existing one.

I think the creation is that we don't use WP core signon during a registration so the hook `wp_login` never fires however I should be seeing a session opened during a regular login. I'm not seeing that happen (still)

The result is that all the tests pass and the code review looks good but when I actually checkout the PR and watch changes to the DB I never see any sessions opened in the new table.

tests/phpunit/unit-tests/functions/updates/class-llms-test-functions-updates-450.php
eri-trabiccolo: Ok, I just wanted to make it parametric... but ok. :)

eri-trabiccolo:
Ohh I see what you mean.
This actually fixes #1349 because it makes sure to record an open session prior to record a sign-on event, if there's no current session open. I also testes it manually - checking the db - loggin in from wp-login.php.
What you're describing here seems to be another bug then, basically autenticating via our registration and our login form doesn't not trigger the hook the wp-login hook. Right?
I'm going to look into this and see what other hook we can use, in place of, or in addition to wp-login.

eri-trabiccolo:
@thomasplevy 
about your latest found, I've opened a new issue https://github.com/gocodebox/lifterlms/issues/1359
but I'm going to include the fix here.

includes/class.llms.person.handler.php
```diff
@@ -21,34 +21,31 @@ class LLMS_Person_Handler {
...
+	 *
+	 * @param string $email User's email address.
+	 * @return string
 	 */
 	public static function generate_username( $email ) {
 
-		/**
-		 * Allow custom username generation
-		 */
+		// Allow custom username generation.
```
thomasplevy:
```suggestion
		/*
		 * Provide a custom username generation function
		 * 
		 * If this filter returns a string, that string will be used for the student's username
		 * rather than the username automatically generated by LifterLMS.
		 *
		 * @since Unknown
		 *
		 * @param null|string $username A `null` value will cause a username to be generated, otherwise the value will be used as the username.
		 * @param string      $email    The user's email address.
		 */	
```

includes/class.llms.person.handler.php
```diff
@@ -759,6 +764,15 @@ public static function register( $data = array(), $screen = 'registration', $sig
...
 				llms_set_person_auth_cookie( $person_id, false );
+				LLMS()->events()->record(
+					array(
+						'actor_id'     => $person_id,
+						'object_type'  => 'user',
+						'object_id'    => $person_id,
+						'event_type'   => 'account',
+						'event_action' => 'signon',
+					)
+				);
```
thomasplevy: I don't like this.

I know this fixes the issue but on reviewing the issue in #1359 it looks to me like there's actually no real reason for `llms_set_person_auth_cookie()` to exist and we could be more generally compatible with the WP core (and other plugins) by rewriting this as:

```suggestion
			// Signon.
			if ( $signon ) {
				wp_signon(
					array(
						'user_login'    => $data['user_login'],
						'user_password' => $data['password'],
					),
					is_ssl()
				);
			}
```

Since we set the `llms_last_login` usermeta data in the `llms_set_person_auth_cookie()` method we lose this data in this change.

However, in looking into this I found that we actually don't set this last login during an actual login, only during a registration.

So we could add a simple function (attached to the `wp_login` action) that will record this metadata and then we will regain this during registration and also add it during an actual login (via WP or LifterLMS or any other plugin that may take over login, say WooCommerce which we actually have an issue with here in that this metadata would never be recorded when using LifterLMS with this plugin). 

We can deprecate a function this way. What do you think?

I ran a quick test of this locally and I don't see any real issues with this change.

thomasplevy (changes requested):

thomasplevy:
@eri-trabiccolo sorry for nitpicking on this one...

includes/class.llms.person.handler.php
eri-trabiccolo: @thomasplevy agreed.
I honestly tend to be conservative, even if I thought to not call `llms_set_person_auth_cookie()` too.
About the `llms_last_login` meta mmm I would just get rid of its setting. Where do we use it? Since we only set it on registration it's not even a last login, but a first login... 
Any statistic based on it would be fake, I think we use it only as fall-back in the last-seen column (which leverages the account.* events)...

thomasplevy: @eri-trabiccolo 

> About the llms_last_login meta mmm I would just get rid of its setting. Where do we use it? Since we only set it on registration it's not even a last login, but a first login...
Any statistic based on it would be fake, I think we use it only as fall-back in the last-seen column (which leverages the account.* events)...

I think that at some point it wasn't fake and got lost in an unrelated fix or refactor at some point.

But yes, I think you're right we can probably drop it all together.

eri-trabiccolo: @thomasplevy 
The `last_login` is used in two places:
- users table
- as fall back for "last seen" in the report's students table 

I think we want to keep this information separated from the "last seen", and that we want to avoid "calculating" it each time. Also, the `last_login` being a user meta, when populating the users table, is retrieved along with the other user metas and cache by wp in a single query.
So I think I'm going to follow your first "advice": attach the usermeta recording (with the current value `current_time('mysql')`) to the `wp_login` action, although I'm not exactly sure where to put this :D

eri-trabiccolo:
@thomasplevy I'm not sure that's a good place where to add the `wp_login` action.
https://github.com/gocodebox/lifterlms/pull/1353/commits/4f6dc4b6816d0fe1637991f3f6029119156c4d7c

let me know

PR #1354, opened 2020-09-28
Add tests for PR #1328

thomasplevy:
Builds upon the work by @alaa-alshamy in #1328 

+ Adds tests
+ Adds / improves / fixes docs and CS in updated files

### DIFF
--- a/includes/class.llms.query.php
+++ b/includes/class.llms.query.php
@@ -273,7 +273,7 @@ class LLMS_Query {
 	 */
 	public function maybe_404_certificate() {
 
-		if ( in_array( get_post_type(), array( 'llms_certificate', 'llms_my_certificate' ), true ) ) {
+		if ( 'llms_my_certificate' === get_post_type() ) {
 			$cert = new LLMS_User_Certificate( get_the_ID() );
 			if ( ! $cert->can_user_view() ) {
 
--- a/includes/class.llms.query.php
+++ b/includes/class.llms.query.php
@@ -7,7 +7,7 @@
  * @package LifterLMS/Classes
  *
  * @since 1.0.0
- * @version 4.0.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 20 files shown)


thomasplevy:
@eri-trabiccolo I've written tests and made a small change regarding how we handle "unviewable" certificates.

Can you have another look before we push to dev and have a careful look at this commit: https://github.com/gocodebox/lifterlms/pull/1354/commits/b7e2c544f7ee222275f01fe955b407523d94ae41

This commit will cause an actual 404 to be served when a certificate cannot be viewed by the current user. This is a change from the existing "error" screen that shows an error message ("Certificate cannot be found") on a plain white background. 

I think a true 404 is favorable to the previous. Thoughts?

includes/class.llms.query.php
```diff
@@ -262,6 +264,29 @@ public function pre_get_posts( $query ) {
...
+	/**
+	 * Serve a 404 for certificates that are not viewable by the current user
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function maybe_404_certificate() {
+
+		if ( in_array( get_post_type(), array( 'llms_certificate', 'llms_my_certificate' ), true ) ) {
```
eri-trabiccolo:
```suggestion
		if ( is_singular( 'llms_my_certificate' ) ) {
```
I think this fits even better, also we don't need a check on the `llms_certificate` because it's not publicly queryable unless you can `manage_lifterlms`.

eri-trabiccolo (changes requested):
Looks good to me, even if technically that's more a 401 rather than a 404 (right?).

includes/class.llms.query.php
thomasplevy: mmm, yes, you are correct, good call.

thomasplevy:
> Looks good to me, even if technically that's more a 401 rather than a 404 (right?).

While yes, it technically is, I think in this scenario we don't want to expose to non-authenticated users that it even can potentially exist. This, I think, is pretty common in web apps when something requires authentication but even exposing that it *might* be accessible is not desired. For example, if you visit a private repository on GitHub which you don't have access to you'll be met with a 404 not a 401.

If you try to access a private users profile on some social media sites you'll see this kind of behavior too.

So yes, it's semantically a 401 but for the sake of privacy I think it should remain a 404

includes/class.llms.query.php
thomasplevy: I'm still going to do a comparison against `get_post_type()` though because it's easier to mock for tests.

PR #1362, opened 2020-10-06
Make sure to remove only custom caps from administrator role when rem…

### LINKED ISSUE #1352: Deleting the plugin when constant LLMS_REMOVE_ALL_DATA set to true results in site.domain/wp-admin/index.php to be not accessible
### Reproduction Steps

+ in your wp-config.php add `define( 'LLMS_REMOVE_ALL_DATA', true );` 
+ as admin disable LifterLMS and then delete it.
+ visit your side.domain/wp-admin/index.php

### Expected Behavior

+ I'd expect to be able to access that page without problems

### Actual Behavior

+ I see a white page with a WordPress message saying I'm not allowed to access that page

The problem should be here:
https://github.com/gocodebox/lifterlms/blob/4.4.4/uninstall.php#L41
so here, where the custom caps for the administrator role should be removed: https://github.com/gocodebox/lifterlms/blob/trunk/includes/class.llms.roles.php#L390
but actually we seem to remove all the role's caps:
https://github.com/gocodebox/lifterlms/blob/trunk/includes/class.llms.roles.php#L405

### Error Messages / Logs

```
"Sorry, you are not allowed to access this page."
```

```
LifterLMS 4.4.4
WordPress 5.5.1
```

This issue has be recreated:
+ [x] Locally
+ [x] On a staging site
+ [ ] On a production website

n/a

### Related User Information
HS-135912

eri-trabiccolo:
## Description
Fixes #1352 

## How has this been tested?
Manually, with existing unit tests, and updating the existing "remove roles" test.

## Types of changes

## Checklist:

### DIFF
--- a/includes/class.llms.roles.php
+++ b/includes/class.llms.roles.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Classes
  *
  * @since 3.13.0
- * @version 3.34.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -16,7 +16,7 @@ defined( 'ABSPATH' ) || exit;
  * @since 3.13.0
  * @since 3.14.0 Add the `lifterlms_instructor` capability.
  * @since 3.34.0 Added the `list_users` capability to instructors.
- *                  Added capabilities for student management.
+ *               Added capabilities for student management.
  */
 class LLMS_Roles {
 
@@ -24,6 +24,7 @@ class LLMS_Roles {
 	 * Retrieve an array of all capabilities for a role
 	 *
 	 * @since 3.13.0
+	 *
 	 * @param string $role Name of the role.
 	 * @return array
 	 */
@@ -387,7 +388,7 @@ class LLMS_Roles {
 		}
 
 		// Remove custom caps from the WP core admin role.
-		self::update_caps( $wp_roles->get_role( 'administrator' ), 'remove' );
+		self::update_caps( $wp_roles->get_role( 'administrator' ), 'remove', array( 'wp' ) );
 
 	}
 
@@ -395,14 +396,23 @@ class LLMS_Roles {
 	 * Update the capabilities for a given role
 	 *
 	 * @since 3.13.0
+	 * @since [version] Added `$exclude_group` parameter that allows excluding groups of caps from the update.
 	 *
-	 * @param WP_Role $role Role object.
-	 * @param string  $type Update type [add|remove].
+	 * @param WP_Role  $role           Role object.
+	 * @param string   $type           Update type [add|remove].
+	 * @param string[] $exclude_groups Array of groups to exclude.
 	 * @return void
 	 */
-	private static function update_caps( $role, $type = 'add' ) {
+	private static function update_caps( $role, $type = 'add', $exclude_groups = array() ) {
 
-		foreach ( self::get_all_caps( $role->name ) as $group => $caps ) {
+		$role_caps = self::get_all_caps( $role->name );
+		if ( ! empty( $exclude_groups ) ) {
+			foreach ( $exclude_groups as $exclude_group ) {
+				unset( $role_caps[ $exclude_group ] );
+			}
+		}
+
+		foreach ( $role_caps as $group => $caps ) {
 
 			foreach ( array_keys( $caps ) as $cap ) {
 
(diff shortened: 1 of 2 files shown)


tests/phpunit/unit-tests/class-llms-test-roles.php
```diff
@@ -65,24 +67,26 @@ public function test_install() {
 
 		$wp_roles = wp_roles();
 
-		// remove first
+		// Remove first.
 		LLMS_Roles::remove_roles();
-
-		// install them
+		add_filter( 'do_print', '__return_true' );
```
thomasplevy: What does this filter do? I cannot find it in the WP code ref or core code base (via manual searching)

includes/class.llms.roles.php
```diff
@@ -387,22 +388,31 @@ public static function remove_roles() {
...
 	 */
-	private static function update_caps( $role, $type = 'add' ) {
+	private static function update_caps( $role, $type = 'add', $exclude_groups = array() ) {
+
+		$role_caps = self::get_all_caps( $role->name );
+		if ( ! empty( $exclude_groups ) ) {
+			foreach ( $exclude_groups as $exclude_group ) {
+				unset( $role_caps[ $exclude_group ] );
+			}
+		}
```
thomasplevy: I think this can be written more succinctly (and by that I mean without having to run (essentially) the same loop twice by using a `continue` in the original loop.

```suggestion
```

includes/class.llms.roles.php
```diff
@@ -387,22 +388,31 @@ public static function remove_roles() {
...
 
-		foreach ( self::get_all_caps( $role->name ) as $group => $caps ) {
+		foreach ( $role_caps as $group => $caps ) {
```
thomasplevy, line 411:
```suggestion
		foreach ( self::get_all_caps( $role->name ) as $group => $caps ) {
		
			if ( in_array( $group, $exclude_groups, true ) ) {
				continue;
			}
		
```

thomasplevy (changes requested):
I think we can remove double looping in the updated `update_caps()` method with the suggestions below

tests/phpunit/unit-tests/class-llms-test-roles.php
eri-trabiccolo: ahah, that was for testing, sorry.

includes/class.llms.roles.php
eri-trabiccolo: well actually it's not running the same loop twice... and the idea was that in most of that cases with my code we make only 1 check, and if true (basically once, only when deleting the plugin), we run a for over 1 element.
With your suggestion below, we make a check for each group, and in most of the cases that check is useless.
So I chose ugliness+less_executions over succinctness.

If we want succinctness what about:
```php
$role_caps = self::get_all_caps( $role->name );
$role_caps = empty( $exclude_groups ) ? $role_caps : array_diff_key( $role_caps, array_flip( $exclude_groups ) );
```

what do you think?

tests/phpunit/unit-tests/class-llms-test-roles.php
eri-trabiccolo: fixed in c661861

includes/class.llms.roles.php
thomasplevy: Sure that works too.

I just don't like the idea of a single function performing two similar loops when it doesn't have to.

eri-trabiccolo: Ok I will probably go with the fancy `array_flip` one because it better feeds my ego.

eri-trabiccolo: solved in 0779fe1

tests/phpunit/unit-tests/class-llms-test-roles.php
```diff
@@ -65,24 +67,26 @@ public function test_install() {
...
 		$wp_roles = wp_roles();
 
-		// remove first
+		// Remove first.
 		LLMS_Roles::remove_roles();
 
-		// install them
+		// Install them.
 		LLMS_Roles::install();
+		remove_filter( 'do_print', '__return_true' );
```
thomasplevy:
```suggestion
		remove_filter( 'do_print', '__return_true' );
```

thomasplevy (changes requested):
See suggestion to remove debug line and then we're good to go

tests/phpunit/unit-tests/class-llms-test-roles.php
eri-trabiccolo: Ok this is stupid (I mean the fact that I left that), sorry.

eri-trabiccolo: fixed in 8ebca20

PR #1364, opened 2020-10-07
WIP: Improve LLMS_Database_Query performance when no pagination information is needed

### LINKED ISSUE #933: Improve LLMS_Database_Query performances when no pagination information is needed
See:
https://github.com/gocodebox/lifterlms/pull/928#discussion_r325180248

Basically we can avoid counting the found rows when executing db queries if we don't need pagination information.

eri-trabiccolo:
## Description
This is a WIP fix for #933

It's might seem a pretty long PR but it actually doesn't contain a lot of relevant changes, although reviewing it might be heavy.
I think it would be better starting reviewing it now, so that we can proceed incrementally.
This is not finished yet, and tested only with existing unit tests.

## How has this been tested?
Only existing unit tests.

## Screenshots 
n/a

## Types of changes
Performance improvements.

## Checklist:

### DIFF
--- a/includes/abstracts/abstract.llms.database.query.php
+++ b/includes/abstracts/abstract.llms.database.query.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Abstracts/Classes
  *
  * @since 3.8.0
- * @version 3.34.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -147,11 +147,12 @@ abstract class LLMS_Database_Query {
 	}
 
 	/**
-	 * Retrieve default arguments for a the query
+	 * Retrieve default arguments for a the query.
 	 *
-	 * @return   array
-	 * @since    3.8.0
-	 * @version  3.8.0
+	 * @since 3.8.0
+	 * @since [version] Added new default arg `no_found_rows` set to false.
+	 *
+	 * @return array
 	 */
 	protected function get_default_args() {
 
@@ -163,6 +164,7 @@ abstract class LLMS_Database_Query {
 				'id' => 'ASC',
 			),
 			'suppress_filters' => false,
+			'no_found_rows'    => false,
 		);
 
 		if ( $this->get( 'suppress_filters' ) ) {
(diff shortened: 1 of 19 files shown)


includes/abstracts/abstract.llms.database.query.php
```diff
@@ -402,6 +405,28 @@ protected function setup_args() {
...
+
+		if ( ! $this->get( 'no_found_rows' ) ) {
+			$select_columns = 'SQL_CALC_FOUND_ROWS ' . $select_columns;
+		}
+
+		if ( $this->get( 'suppress_filters' ) ) {
+			return $select_columns;
+		}
+
+		return apply_filters( $this->get_filter( 'select_columns' ), $select_columns, $this );
```
thomasplevy, line 427: I thought this function was a good idea when I wrote it but years later I realize it makes it hard to read the code when trying to figure out what the filter actually does.

I'd like to remove it eventually.

Knowing this, while working on improving this and related classes if new filters need to be created I would like to not use this function. Please.

thomasplevy:
This looks like a generally good start. I like the approach.

However, I'm very nervous because I can already tell (as you've noted) that this is going to be a very large set of changes.

Furthermore, a great number of the areas affected by this change are currently uncovered. That means that we're going to have to add quite a lot of tests before we can confidently ship this.

includes/abstracts/abstract.llms.database.query.php
eri-trabiccolo: Oh yeah I see, so:
```suggestion
		return apply_filters( "llms_{$this->id}_query_select_columns", $select_columns, $this );
```
I can change the other occurrences as well.

eri-trabiccolo: Oh yeah I see, so:
```suggestion
		return apply_filters( "llms_{$this->id}_query_select_columns", $select_columns, $this );
```
I can change the other occurrences as well.

eri-trabiccolo:
> This looks like a generally good start. I like the approach.

<3

> However, I'm very nervous because I can already tell (as you've noted) that this is going to be a very large set of changes.

Yes, let me add that actually the real changes are not that many and nothing changes by default in the abstract class.
Do you want to follow a different approach? E.g. first a PR that only introduces this feature for the abstract class - by default the extending classes won't be affected by this change.
When it gets merged I will start pushing smaller PRs for each extending class (or event two at a time, depending on the number of affected lines) that will start using that feature, so that will be easier to review/test.
What do you think? @thomasplevy 
 
> Furthermore, a great number of the areas affected by this change are currently uncovered. That means that we're going to have to add quite a lot of tests before we can confidently ship this.

Yes a lot of tests must be added.

thomasplevy:
> > However, I'm very nervous because I can already tell (as you've noted) that this is going to be a very large set of changes.
> 
> Yes, let me add that actually the real changes are not that many and nothing changes by default in the abstract class.
> Do you want to follow a different approach? E.g. first a PR that only introduces this feature for the abstract class - by default the extending classes won't be affected by this change.
> When it gets merged I will start pushing smaller PRs for each extending class (or event two at a time, depending on the number of affected lines) that will start using that feature, so that will be easier to review/test.
> What do you think? @thomasplevy

This feels like a much more manageable approach to me. Even if we have a dozen PRs that get merged into a single release I feel it will be easier to properly review each one in this way. 

I think if we keep each PR with a target of 3-5 affected files that would make it feel easier to manage. 

That would mean that with tests we could update up to two classes per PR.

I know that's a lot more merging and branching and so on but I'll feel more confident in the release and review process in this way.

Thank you

eri-trabiccolo:
This will be split in several PRs, as they'll be more manageable.
Closing this then, thanks for your attention you all! <3

PR #1384, opened 2020-10-15
bionic to focal take2

thomasplevy:
Testing new travis configs


thomasplevy:
nevermind, this is all slower than bionic.

PR #1386, opened 2020-10-15
Import / Generator Refactor: Improvements, Features, and Bug Fixes

### LINKED ISSUE #1196: Importer does not import images from the post_content or custom fields (except featured images)

### LINKED ISSUE #1326: Course importer does not allow usage of reusable blocks

thomasplevy:
## Description

Refactors `LLMS_Generator`, splitting much of the "base" functionality into an abstract and all course-creation related functionality moved into a new class `LLMS_Generator_Courses`.

The original `LLMS_Generator` class becomes a handler used to call "generators".

In the future we will add additional generators, for example membership, student, enrollment data, etc...

This is a "finished" (hopefully?) version of the WIP submitted via #1341 

Fixes #1196 
Fixes #1326 

## How has this been tested?

+ Manually
+ New phpunit tests
+ Existing tests

## Types of changes
+ Bug fixes
+ Refactor
+ New features

## Checklist:


thomasplevy:
@eri-trabiccolo I know you did a thorough review of a large portion of this in #1341 but this is now more finished and includes (I think) most of your notes from that PR.

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -0,0 +1,782 @@
+<?php
+/**
+ * Generate LMS Content from export files or raw arrays of data
+ *
+ * @package LifterLMS/Classes
```
eri-trabiccolo:
```suggestion
 * @package LifterLMS/Abstracts/Classes
```

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -0,0 +1,782 @@
...
+/**
+ * LLMS_Generator class.
+ *
+ * @since 3.3.0
+ * @since 3.30.2 Added hooks and made numerous private functions public to expand extendability.
+ * @since 3.36.3 New method: is_generator_valid()
+ *               Bugfix: Fix return of `set_generator()`.
+ * @since [version] Add sideloading of images found in imported post content.
+ */
+abstract class LLMS_Abstract_Generator_Posts {
```
eri-trabiccolo: I'm not sure, this file didn't exist before, nor the class, and certainly the class summary doesn't fit

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -0,0 +1,782 @@
...
+
+	/**
+	 * Generate a new LLMS_Post_Mdel
+	 *
+	 * @since [version]
+	 *
+	 * @param string $type      The LLMS_Post_Model post type type. For example "course" for an `LLMS_Course` or `membership` for `LLMS_Membership`.
+	 * @param array  $raw       Array of raw, used to create the post.
+	 * @param int    $author_id Fallback author ID, used when now author data can be found in `$raw`.
+	 * @throws Exception When the class identified by `$type` is not found or when an error is encountered during post creation.
```
eri-trabiccolo: is this the right position for this tag?
I think we don't have a rule for that, but I would prefer it somewhere else, like
```
@since

@throw

@param
@return
```
looks more readable to me...

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -0,0 +1,782 @@
...
+		$block_id = absint( $block_id );
+
+		// If the block already exists, don't create it again.
+		$existing = get_post( $block_id );
+		if ( $existing && 'wp_block' === $existing->post_type && $block['title'] === $existing->post_title && $block['content'] === $existing->post_content ) {
+			return false;
+		}
+
+		// Check if the block was previously imported.
+		$id = empty( $this->reusable_blocks[ $block_id ] ) ? false : $this->reusable_blocks[ $block_id ];
```
eri-trabiccolo: can't we switch the order of these two checks?
Check whether it was previously imported, if not, check if the block already exists... So we "avoid" a get_post() in that rare case the reusable block was already imported during the same importing session. What do you think?
(https://github.com/gocodebox/lifterlms/pull/1341#discussion_r496700942)

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -0,0 +1,782 @@
...
+	/**
+	 * Add custom data to a post based on the 'custom' array
+	 *
+	 * @since 3.16.11
+	 * @since 3.28.3 Add extra slashes around JSON strings.
+	 * @since 3.30.2 Skip JSON evaluation for non-string values; make publicly accessible.
+	 * @since [version] Moved from `LLMS_Generator`.
+	 *
+	 * @param int   $post_id WP Post ID.
+	 * @param array $raw raw data.
```
eri-trabiccolo:
```suggestion
	 * @param array $raw     Raw data.
```

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -0,0 +1,782 @@
...
+		 * @param array $raw       Original raw author data.
+		 */
+		return apply_filters( 'llms_generator_get_author_id', $author_id, $raw );
+
+	}
+
+	/**
+	 * Receives a raw array of course, plan, section, lesson, etc data and gets an author id
+	 * falls back to optionally supplied fallback id
+	 * falls back to current user id
```
eri-trabiccolo:
```suggestion
	 * Receives a raw array of course, plan, section, lesson, etc data and gets an author id
	 *
	 * Falls back to optionally supplied fallback id.
	 * Falls back to current user id.
```

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -0,0 +1,782 @@
...
+	 *
+	 * @since 3.3.0
+	 * @since 3.30.2 Made publicly accessible.
+	 * @since [version] Moved from `LLMS_Generators`.
+	 *
+	 * @param array $raw raw data
+	 * @param int   $fallback_author_id WP User ID
```
eri-trabiccolo:
```suggestion
	 * @param array $raw                Raw data.
	 * @param int   $fallback_author_id Optional. WP User ID. Default is `null`.
	 *                                  If not supplied, if no author is set, the current user ID will be used.
```

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -0,0 +1,782 @@
...
+	/**
+	 * Saves an image (from URL) to the media library and sets it as the featured image for a given post
+	 *
+	 * @since 3.3.0
+	 * @since [version] Moved from `LLMS_Generator` and made `protected` instead of `private`.
+	 *               Add a return instead of `void`; Don't import if sideloading is disabled; Use `$this->sideload_image()` sideloading.
+	 *
+	 * @param string $url_or_raw Array of raw data or URL to an image.
+	 * @param int    $post_id    WP Post ID.
+	 * @return null|false|int Returns `null` if sideloading is disabled, WP Post ID of the attachement on success, `false` on error.
```
eri-trabiccolo:
```suggestion
	 * @return null|false|int Returns `null` if sideloading is disabled, WP Post ID of the attachment on success, `false` on error.
```

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -0,0 +1,782 @@
...
+	}
+
+	/**
+	 * Sideload an image from a url
+	 *
+	 * @since [version]
+	 *
+	 * @param int    $post_id WP_Post ID of the post where the image will be attached.
+	 * @param string $url     The image's URL.
+	 * @return string|int|WP_Error Returns a WP_Error on failure, the image's new URL when `$return` is "src", otherwise returns the image's attachement ID.
```
eri-trabiccolo:
```suggestion
	 * @return string|int|WP_Error Returns a WP_Error on failure, the image's new URL when `$return` is "src", otherwise returns the image's attachment ID.
```

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -0,0 +1,782 @@
...
+	/**
+	 * Sideload images found in a given post
+	 *
+	 * This attempts to sideload the `src` attribute of every <img> element
+	 * found in the `post_content` of the supplied post.
+	 *
+	 * @since [version]
+	 *
+	 * @param LLMS_Post_Model $post Post object.
+	 * @param array           $raw  Array of raw data
```
eri-trabiccolo:
```suggestion
	 * @param array           $raw  Array of raw data.
```

includes/class-llms-generator-courses.php
```diff
@@ -0,0 +1,633 @@
...
+ * @package LifterLMS/Classes
+ *
+ * @since [version]
+ * @version [version]
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * LLMS_Generator class.
```
eri-trabiccolo:
```suggestion
 * LLMS_Generator_Courses class
```

includes/class-llms-generator-courses.php
```diff
@@ -0,0 +1,633 @@
...
+
+	/**
+	 * Generator called for single course imports
+	 *
+	 * Converts the single course into a format that can be handled by the bulk courses generator
+	 * and invokes that generator.
+	 *
+	 * @since 3.3.0
+	 * @since [version] Moved from `LLMS_Generator` and made `public` instead of `private`.
+	 *              Returns an int on success
```
eri-trabiccolo:
```suggestion
	 *              Returns an int on success.
```

includes/class-llms-generator-courses.php
```diff
@@ -0,0 +1,633 @@
...
+	 * Create a new access plan
+	 *
+	 * @since 3.3.0
+	 * @since 3.7.3 Unknown.
+	 * @since 4.3.3 Use an empty string in favor of `null` for an empty `post_content` field.
+	 * @since [version] Sideload images attached to the post, use `create_post()` from abstract, add hooks.
+	 *
+	 * @param array $raw                Raw Access Plan Data
+	 * @param int   $course_id          WP Post ID of a LLMS Course to assign the access plan to
+	 * @param int   $fallback_author_id WP User ID to use for the access plan author if no author is supplied in the raw data
```
eri-trabiccolo:
```suggestion
	 * @param array $raw                Raw Access Plan Data.
	 * @param int   $course_id          WP Post ID of a LLMS Course to assign the access plan to.
	 * @param int   $fallback_author_id Optional. WP User ID to use for the access plan author if no author is supplied in the raw data. Default is `null`.
	 *                                  When not supplied the fall back will be on the current user ID.
```

includes/class-llms-generator-courses.php
```diff
@@ -0,0 +1,633 @@
...
+		 *
+		 * @param array          $raw       Raw course data array.
+		 * @param LLMS_Generator $generator Generator instance.
+		 */
+		$raw = apply_filters( 'llms_generator_before_new_access_plan', $raw, $this );
+
+		// Force course relationship.
+		$raw['product_id'] = $course_id;
+
+		$plan = $this->create_post( 'access_plan', $raw, get_current_user_id() );
```
eri-trabiccolo:
```suggestion
		$plan = $this->create_post( 'access_plan', $raw, $fallback_author_id );
```

includes/class-llms-generator-courses.php
```diff
@@ -0,0 +1,633 @@
...
+	 * @since 3.3.0
+	 * @since 3.30.2 Added hooks.
+	 * @since 4.3.3 Use an empty string in favor of `null` for empty `post_content` and `post_excerpt` fields.
+	 * @since [version] Import images and reusable blocks found in the post's content and use `create_post()` from abstract.
+	 *
+	 * @param array $raw                Raw lesson data.
+	 * @param int   $order              Lesson order within the section (starts at 1).
+	 * @param int   $section_id         WP Post ID of the lesson's parent section.
+	 * @param int   $course_id          WP Post ID of the lesson's parent course.
+	 * @param int   $fallback_author_id Optional author ID to use as a fallback if no raw author data supplied for the lesson.
```
eri-trabiccolo:
```suggestion
	 * @param int   $fallback_author_id Optional. Author ID to use as a fallback if no raw author data supplied for the lesson. Default is `null`.
	 *                                  When not supplied the fall back will be on the current user ID.

```

includes/class-llms-generator-courses.php
```diff
@@ -0,0 +1,633 @@
...
+	 * Creates a new quiz
+	 * Creates all questions within the quiz as well
+	 *
+	 * @since 3.3.0
+	 * @since 3.30.2 Added hooks.
+	 * @since 4.3.3 Use an empty string in favor of `null` for an empty `post_content` field.
+	 * @since [version] Sideload images attached to the post  and use `create_post()` from abstract.
+	 *
+	 * @param array $raw                Raw quiz data.
+	 * @param int   $fallback_author_id Optional author ID to use as a fallback if no raw author data supplied for the lesson.
```
eri-trabiccolo:
```suggestion
	 * @param int   $fallback_author_id Optional. Author ID to use as a fallback if no raw author data supplied for the quiz. Default is `null`.
	 *                                  When not supplied the fall back will be on the current user ID.

```

includes/class-llms-generator-courses.php
```diff
@@ -0,0 +1,633 @@
...
+	/**
+	 * Creates a new question
+	 *
+	 * @since 3.3.0
+	 * @since 3.30.2 Added hooks.
+	 * @since [version] Attempt to sideload images found in the imported post's content and image choices.
+	 *
+	 * @param array $raw       Raw question data.
+	 * @param obj   $manager   Question manager instance.
+	 * @param int   $author_id Optional author ID to use as a fallback if no raw author data supplied for the lesson.
```
eri-trabiccolo:
```suggestion
	 * @param int   $author_id Optional. Author ID to use as a fallback if no raw author data supplied for the question. Default is `null`.
	 *                         When not supplied the fall back will be on the current user ID.
```

includes/class-llms-generator-courses.php
```diff
@@ -0,0 +1,633 @@
...
+	 * Creates all lessons within the section data.
+	 *
+	 * @since 3.3.0
+	 * @since 3.30.2 Added hooks.
+	 * @since [version] Use `create_post()` from abstract.
+	 *
+	 * @param array $raw                Raw section data.
+	 * @param int   $order              Order within the course (starts at 1).
+	 * @param int   $course_id          WP Post ID of the parent course.
+	 * @param int   $fallback_author_id Optional author ID to use as a fallback if no raw author data supplied for the lesson.
```
eri-trabiccolo:
```suggestion
	 * @param int   $fallback_author_id Optional. Author ID to use as a fallback if no raw author data supplied for the section.
	 *                                  When not supplied the fall back will be on the current user ID.

```

includes/class-llms-generator-courses.php
```diff
@@ -0,0 +1,633 @@
...
+
+		/**
+		 * Filter raw section import data prior to generation
+		 *
+		 * @since 3.30.2
+		 *
+		 * @param array          $raw                Raw quiz data array.
+		 * @param int            $order              Order within the course (starts at 1).
+		 * @param int            $course_id          WP Post ID of the parent course.
+		 * @param int            $fallback_author_id Optional author ID to use as a fallback if no raw author data supplied for the lesson.
```
eri-trabiccolo:
```suggestion
		 * @param int            $fallback_author_id Optional author ID to use as a fallback if no raw author data supplied for the section.
```

includes/class-llms-generator-courses.php
```diff
@@ -0,0 +1,633 @@
...
+	protected function create_question( $raw, $manager, $author_id ) {
+
+		/**
+		 * Filter raw question import data prior to generation
+		 *
+		 * @since 3.30.2
+		 *
+		 * @param array          $raw       Raw quiz data array.
+		 * @param obj            $manager   Question manager instance.
+		 * @param int            $author_id Optional author ID to use as a fallback if no raw author data supplied for the lesson.
```
eri-trabiccolo:
```suggestion
		 * @param int            $author_id Optional author ID to use as a fallback if no raw author data supplied for the question.
```

includes/class-llms-generator-courses.php
```diff
@@ -0,0 +1,633 @@
...
+	 */
+	protected function create_quiz( $raw, $fallback_author_id = null ) {
+
+		/**
+		 * Filter raw quiz import data prior to generation
+		 *
+		 * @since 3.30.2
+		 *
+		 * @param array          $raw                Raw quiz data array.
+		 * @param int            $fallback_author_id Optional author ID to use as a fallback if no raw author data supplied for the lesson.
```
eri-trabiccolo:
```suggestion
		 * @param int            $fallback_author_id Optional author ID to use as a fallback if no raw author data supplied for the quiz.
```

includes/class-llms-loader.php
```diff
@@ -96,6 +96,7 @@ public function includes() {
 		require_once LLMS_PLUGIN_DIR . 'includes/abstracts/abstract.llms.database.query.php';
```
eri-trabiccolo, line 96: missing method's changelog
and version change in the file header.

includes/class.llms.generator.php
```diff
@@ -1080,69 +308,93 @@ protected function is_generator_valid( $generator ) {
...
+		switch ( current_action() ) {
+
+			case 'llms_generator_new_access_plan':
+			case 'llms_generator_new_course':
+			case 'llms_generator_new_section':
+			case 'llms_generator_new_lesson':
+			case 'llms_generator_new_quiz':
+			case 'llms_generator_new_question':
+				$this->record_generation( $object->get( 'id' ), $object->get( 'type' ) );
+				break;
```
eri-trabiccolo, line 329: Can't we turn this into a `default` ?
I see that in this case we're sure we don't record the generation of something no explicitly specified in the case statements above but..

includes/class.llms.generator.php
```diff
@@ -1080,69 +308,93 @@ protected function is_generator_valid( $generator ) {
...
 	 *
 	 * @since 3.7.3
+	 * @since [version] Call `set_default_post_status()` from the configured generator.
 	 *
 	 * @param string $status Any valid WP Post Status.
 	 * @return void
 	 */
 	public function set_default_post_status( $status ) {
-		$this->default_post_status = $status;
+		call_user_func( array( $this->generator[0], 'set_default_post_status' ), $status );
```
eri-trabiccolo, line 397:
```suggestion
		$this->generator[0]->set_default_post_status( $status );
```
We can, right?

includes/class.llms.generator.php
```diff
@@ -1189,30 +441,96 @@ public function set_generator( $generator = null ) {
...
+	 * @since 3.16.11
+	 * @since 3.28.3 Add extra slashes around JSON strings.
+	 * @since 3.30.2 Skip JSON evaluation for non-string values; make publicly accessible.
+	 * @deprecated [version] `LLMS_Generator::add_custom_values()` is deprecated in favor of `LLMS_Generator_Courses::add_custom_values`.
 	 *
-	 * @param array $raw Array of raw data.
-	 * @param obj   $obj The LLMS Post Object generated from the raw data.
-	 * @return int|false Raw id when present or `false` if no raw id was found.
+	 * @param int   $post_id WP Post ID.
+	 * @param array $raw raw data.
```
eri-trabiccolo:
```suggestion
	 * @param int   $post_id WP Post ID.
	 * @param array $raw     Raw data.
```

includes/class.llms.generator.php
```diff
@@ -1189,30 +441,96 @@ public function set_generator( $generator = null ) {
...
+		llms_deprecated_function( 'LLMS_Generator::format_date()', '[version]', 'LLMS_Generator_Courses::format_date()' );
+		return $this->courses_generator->format_date( $raw_date );
+	}
 
-			return $raw['id'];
+	/**
+	 * Receives a raw array of course, plan, section, lesson, etc data and gets an author id
+	 * falls back to optionally supplied fallback id
+	 * falls back to current user id
+	 *
```
eri-trabiccolo:
```suggestion
	 *
	 * Falls back to optionally supplied fallback id.
	 * Falls back to current user id.
	 *
```

includes/class.llms.generator.php
```diff
@@ -21,892 +21,211 @@
...
-	 * @return string
-	 */
-	public function format_date( $raw_date = null ) {
-
-		if ( ! $raw_date ) {
-			return current_time( 'mysql' );
-		} else {
-			return date( 'Y-m-d H:i:s', strtotime( $raw_date ) );
+		try {
+			call_user_func( $this->generator, $this->raw );
```
eri-trabiccolo, line 131:
```suggestion
			( $this->generator )( $this->raw );
```
why don't we use just this? ^

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -1188,24 +1206,131 @@ public function toArray() {
...
+	 * an import these images will be imported into the new site via media sideloading.
+	 *
+	 * @since [version]
+	 *
+	 * @param array $arr Post array from `toArray()`.
+	 * @return array[]
+	 */
+	protected function to_array_extra( $arr ) {
+
+		$arr['_extra'] = array(
```
```suggestion
		$arr['_extras'] = array(
```
the generators look for '_extras' not for '_extra'

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -0,0 +1,782 @@
...
+
+		// Check if the block was previously imported.
+		$id = empty( $this->reusable_blocks[ $block_id ] ) ? false : $this->reusable_blocks[ $block_id ];
+		if ( ! $id ) {
+
+			$id = wp_insert_post(
+				array(
+					'post_content' => $block['content'],
+					'post_title'   => $block['title'],
+					'post_type'    => 'wp_block',
```
eri-trabiccolo, line 233:
```suggestion
					'post_type'    => 'wp_block',
					'post_status'  => 'publish',
```
otherwise the block will be created as 'draft' and you'll just reference a block that technically "doesn't exist" (that's how gutenberg sees it).

eri-trabiccolo (changes requested):
Some changes requested.
I've also found, when cloning only that:
`[19-Oct-2020 14:56:51 UTC] PHP Deprecated:  LLMS_Generator::get_generated_posts() is <strong>deprecated</strong> since version [version]! Use LLMS_Generator::get_generated_content() instead. in /srv/www/wordpress-two/public_html/wp-includes/functions.php on line 4773
`
then when exporting from a local site A (after fixing the bugs) and importing on a local site B that images were not sideloaded because of this condition met:
https://github.com/WordPress/WordPress/blob/5.5.1/wp-includes/http.php#L554-L575

includes/abstracts/llms-abstract-generator-posts.php
thomasplevy: We don't have a rule but we should.

My thinking was that `@throws` is *kind of * like a return so it should go at the end with the return

Looking through the WP core they have no defined rule that I can find about where this tag should go and they also have it in an inconsistent pattern throughout the core when it is used.

It looks likes its mostly used by the core within included libraries.

Let's just make a rule...

While I think what you have above is better than what I committed I do feel that `@throws` belongs with the return value to a certain extent. I'd like to put it at the end after the return tag, so:

```
@param
@return

@throws
```

~I could also be persuaded to add a space between return and throws if you think that would make it more readable~ I persuaded myself.

includes/abstracts/llms-abstract-generator-posts.php
thomasplevy: good call

includes/class.llms.generator.php
thomasplevy: Why?

thomasplevy:
> hen when exporting from a local site A (after fixing the bugs) and importing on a local site B that images were not sideloaded because of this condition met:
> https://github.com/WordPress/WordPress/blob/5.5.1/wp-includes/http.php#L554-L575

In order to make this scenario work we'd need to do more logic than I think it would be worth, right? If we allow requests to local to run we need to do so conditionally to make sure that the current host is also local otherwise the requests will outright fail.

The only ones that are going to ever encounter this are devs and I think a dev will understand why this is happening (even if we have to explain) and they'd probably say "Okay" and carry on. I'd hope at least.

Right? Is this worth adding *more* logic?


I think at least, am I misreading this?

includes/abstracts/llms-abstract-generator-posts.php
thomasplevy: @eri-trabiccolo moving this discussion to #1397

includes/class.llms.generator.php
eri-trabiccolo: I should give you links about benchmarks of that function, I know, although we already know that function calls have a cost in PHP, so if we could simply avoid it, why not?

eri-trabiccolo:
> > hen when exporting from a local site A (after fixing the bugs) and importing on a local site B that images were not sideloaded because of this condition met:
> > https://github.com/WordPress/WordPress/blob/5.5.1/wp-includes/http.php#L554-L575
> 
> In order to make this scenario work we'd need to do more logic than I think it would be worth, right? If we allow requests to local to run we need to do so conditionally to make sure that the current host is also local otherwise the requests will outright fail.
> 
> The only ones that are going to ever encounter this are devs and I think a dev will understand why this is happening (even if we have to explain) and they'd probably say "Okay" and carry on. I'd hope at least.
> 
> Right? Is this worth adding _more_ logic?
> 
> I think at least, am I misreading this?

I think you're right... But I think we can note it somewhere in the code maybe, so that either a dev or us in 2years from now, when someone will report this, we just have to read why it's happening. What do you think?

I reported this for the only reason that, when testing, I found something that wasn't working as expected and that sharing my findings was worthwhile.

eri-trabiccolo:
@thomasplevy 
Forgot to review the test jsons, they have "_extra" fields while they are supposed to have "_extras". <3

thomasplevy:
> I think you're right... But I think we can note it somewhere in the code maybe, so that either a dev or us in 2years from now, when someone will report this, we just have to read why it's happening. What do you think?
> 
> I reported this for the only reason that, when testing, I found something that wasn't working as expected and that sharing my findings was worthwhile.

okay, we can make a note of this in the codebase but I'm not sure where a note like this should go... I suppose in the docblock on the `LLMS_Abstract_Generator_Posts::side_load_image()` class we can make a note that it's not possible to sideload from local urls?
I

eri-trabiccolo:
> > I think you're right... But I think we can note it somewhere in the code maybe, so that either a dev or us in 2years from now, when someone will report this, we just have to read why it's happening. What do you think?
> > I reported this for the only reason that, when testing, I found something that wasn't working as expected and that sharing my findings was worthwhile.
> 
> okay, we can make a note of this in the codebase but I'm not sure where a note like this should go... I suppose in the docblock on the `LLMS_Abstract_Generator_Posts::side_load_image()` class we can make a note that it's not possible to sideload from local urls?
> I

Yeah `LLMS_Abstract_Generator_Posts::sideload_image()` _that must be the place_ 🎶 , I agree.

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -306,8 +306,10 @@ private function create( $title = '' ) {
 	/**
 	 * Clones the Post if the post is cloneable
 	 *
-	 * @return   WP_Error|int|null WP_Error, WP Post ID of the clone (new) post, or null if post is not cloneable.
-	 * @since    3.3.0
+	 * @since 3.3.0
+	 * @since [version] Use `LLMS_Generator::get_generated_content()` in favor of deprecated `LLMS_Generator::get_generated_posts()`..
```
```suggestion
	 * @since [version] Use `LLMS_Generator::get_generated_content()` in favor of deprecated `LLMS_Generator::get_generated_posts()`.
```

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -1188,24 +1208,131 @@ public function toArray() {
 
 		ksort( $arr ); // Because i'm anal...
 
-		return apply_filters( 'llms_' . $this->model_post_type . '_to_array', $arr, $this );
+		/**
+		 * Filter the final post array created when converting the object to an array
+		 *
+		 * The dynamic portion of this hook, `$this->model_post_type`, refers to the model's post type. For example "course"m
```
```suggestion
		 * The dynamic portion of this hook, `$this->model_post_type`, refers to the model's post type. For example "course",
```

eri-trabiccolo (changes requested):
minor doc stuff, I don't even know if you should fix them...
I made some tests and looks like it works fine.

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -0,0 +1,787 @@
...
+	public function set_default_post_status( $status ) {
+		$this->default_post_status = $status;
+	}
+
+	/**
+	 * Sideload an image from a url
+	 *
+	 * @since [version]
+	 *
+	 * @param int    $post_id WP_Post ID of the post where the image will be attached.
```
thomasplevy:
```suggestion
	 * @since [version]
	 *
	 * @link https://developer.wordpress.org/reference/hooks/http_request_host_is_external/ If exporting from a local site and importing into another local site, images *will not* be side loaded as a result of this condition in the WP Core
	 *
	 * @param int    $post_id WP_Post ID of the post where the image will be attached.
```

thomasplevy (commented):
@eri-trabiccolo how's this?

includes/abstracts/llms-abstract-generator-posts.php
eri-trabiccolo: fine for me.

PR #1387, opened 2020-10-16
UX Improvements to shareable certificates

### LINKED ISSUE #1379: Certificates: New sharing button creates confusing user experience when button is next to the "Save" button
### Reproduction Steps

+ View an earned certificate as a student
+ Click "Enable Sharing" to allow a certificate to be shared

### Expected Behavior

+ Users are expecting that a link will be provided to them which they can use to share the certificate
+ Some users expect that they must also "Save" the certificate (using the "Save" button) after enabling/disabling sharing

### Actual Behavior

+ Clicking "save" is not necessary (as this is actually an export button)
+ There is no user feedback indicating what URL should be used to "Share" the certificate after sharing is enabled.

### Proposal

+ The language of the "Save" button should be changed to "Download" or "Export"
+ When sharing is enabled, a readonly text input should be output with the certificate's URL populated in it. This will help users identify what URL should be used to share the certificate.

imknight:
## Description

Fixes #1379

## Screenshots 

Captura de pantalla 2024-06-04 a las 11 15 40 a  m
ld cause existing functionality to not work as expected) -->

## Checklist:


templates/content-certificate.php
```diff
@@ -54,6 +54,13 @@
 				</button>
 			<?php endif; ?>
 
+			<?php if ( $is_sharing_enabled ) : ?>
+			<span class="llms-button-secondary">
```
thomasplevy: I was envisioning a read-only text input in favor of a button that doesn't do anything when clicked.

thomasplevy (changes requested):

templates/content-certificate.php
imknight: Hi, I thought the proposal mention "When sharing is enabled, a read-only text input should be output with the certificate's URL populated in it. "?

eri-trabiccolo: Hi @imknight ,
yes so in place of this `<span>` that looks like a button, @thomasplevy was envisioning a read-only text input, as mentioned in the proposal.
Is that anything I'm missing?!

imknight: oh got it, finally, I understand, will update the code.
I can change it into input easily, but need some tweak on styles, shall it combine within this change request?
can't really get the gulp working, any documentation that I can read about workflow to tweak the style?

thomasplevy: Have you had a look at our Installing for Development Docs: https://github.com/gocodebox/lifterlms/blob/trunk/docs/installing.md

thomasplevy: The CSS is probably less important than making sure this fundamentally works. Although now that I'm looking at this again we're going to need to add some UX to the field too...

I'm sure anyone who sees a field is also going to expect the following:

1) When you click the field it should highlight everything
2) There should be a "Copy" button next to the field

So we'll need to add some JS to this to power the UX

imknight: I got it for 2 so that if a user clicks on the copy button, it will copy the permalink to the clipboard, but what is 1 about? highlight everything?

thomasplevy: Instead of having to double click (or single click and then ctrl/cmd + a) to highlight the whole link, it would be *simpler* for the end user for them to click once in the field and have the entire contents of the field automatically selected/highlighted.

Here's a quick gif of how google drive handles this (you maybe can't tell but I'm just clicking the field and it automatically highlights the whole contents)

Peek 2020-10-20 09-25

I find that most apps that let you copy a link have this behavior automatically so if we *don't* do it we'll get support requests about how it doesn't work right, almost guaranteed.

Thanks!

templates/content-certificate.php
```diff
@@ -54,6 +54,10 @@
 				</button>
 			<?php endif; ?>
 
+			<?php if ( $is_sharing_enabled ) : ?>
+				<input disabled="disabled" name="llms_sharing_permalink" value="<?php echo esc_url( get_permalink( get_the_ID() ) ); ?>">
```
thomasplevy: Readonly in favor of disabled

A readonly field can still be tabbed into as opposed to a disabled field that cannot

Field name isn't needed since it won't be read by form handlers

Using an ID will make js targeting faster / simpler

```suggestion
				<input readonly="readonly" id="llms-sharing-permalink" value="<?php echo esc_url( get_permalink( get_the_ID() ) ); ?>">
```

thomasplevy (changes requested):

templates/content-certificate.php
thomasplevy: Also, just to test the concept, it can be done pretty simply with some JS like this:

```js
document.getELementById( 'test-input' ).addEventListener( 'focus', e => {
     e.target.select();
} );

**Review by thomasplevy — CHANGES_REQUESTED** [review]
Sorry @imknight and thanks for your patience on this one.

I'm not a huge fan of doing an inline script like this on a client-facing page. We make concessions sometimes but this doesn't feel like a good place to do it.

Especially because on a closer inspection I think we can do it with a oneliner on the input itself...

`onfocus="this.select()"` I think does it, right?

Then we don't have to have an inline `<script>` tag and we don't have to add a JS asset specifically for this purpose.

Script event handler attributes aren't that attractive and I also like to avoid them but in this case I think there's a good argument for using them.

What do you think?

And, again, thanks for your patience!

**Comment by imknight** [comment]
ok, will try to work on oneliner script, I thought for a template is good to use the inline script.

File: templates/content-certificate.php [hunk 3]
```diff
@@ -54,6 +54,10 @@
 				</button>
 			<?php endif; ?>
 
+			<?php if ( $is_sharing_enabled ) : ?>
+				<input readonly="readonly" id="llms_sharing_permalink" onClick="this.select();" value="<?php echo esc_url( get_permalink( get_the_ID() ) ); ?>">
```
thomasplevy: Ensure keyboard users can still benefit. Although it seems in practice that it's still triggered with `onClick` I like semantics of `onfocus`.

```suggestion
				<input readonly="readonly" id="llms_sharing_permalink" onfocus="this.select();" value="<?php echo esc_url( get_permalink( get_the_ID() ) ); ?>">
```

thomasplevy (commented):
I've been holding on this because now that the initial work is complete I think additional UI work is needed before this can be shipped.

1. A copy button is still needed. This adds some complexity due to the nature of copying to the clipboard via browser interactions. It will require a JS file to be included on the certificate to power the UI.
2. Looking at line of buttons makes me realize that we need to change our strategy on how this content is output below the certificate. I don't have a good solution in mind so I don't know how to push this forward. What I'm seeing today doesn't look right with the input (and then later a new button from 1 above) but I don't have any alternatives in mind.

Thank for your work thus far @imknight, we'll get this merged in at some point but at this moment this isn't a high-enough priority item for us to take over internally and since I don't know exactly what direction to move in I don't know how to properly relay this to you.

I need to spend some design time here and I don't have that capacity right now.

Thanks,

imknight:
noted and thank

brianhogg:
We may want to add a "Copy to Clipboard" link/button (see Advanced Coupons UI for accessible suggestions on how to do this), but otherwise making it clear that they need to share the current URL looks good.

Test to make sure UI looks reasonable, potentially add copy to clipboard, then merge.

brianhogg:
Added screenshot. Awaiting approval or if we want to re-work the design a bit.

brianhogg (approved):

brianhogg:
After discussion, switched text input to a button that, when clicked, copies the URL to the clipboard.

imknight:
lol from 2020 until now, finally see it merge!

PR #1388, opened 2020-10-16
Bug fix 1239

imknight:
## Description

Fixes #

## Checklist:


thomasplevy (changes requested):
While this does *work* it feels like adding more logic to the HTML template isn't the *best* way to do this.

We have a method that prepares fields with default data and we can fix this by adding some logic to the existing method instead of adding this condition to the templated.

Adding a case for `switch-number` here would be my preferred solution: https://github.com/gocodebox/lifterlms/blob/153304618c97da71b4743c2b78e55e0731a8ca3f/assets/js/builder/Views/SettingsFields.js#L340

thomasplevy:
@imknight thanks for your contributions to our project. If you could please help us save some effort by, in addition to writing code, also reviewing our contributors guidelines and paying primary attention to our changelog guidelines. I don't like to beat up new contributors about documentation but if you're planning to become a regular contributor I'd love it if you could contribute within our guidelines (which are annoying, I know, but they help us as maintainers a great deal!). Thanks!

Here's the part I'm most interested in: https://github.com/gocodebox/lifterlms/blob/trunk/docs/documentation-standards.md#changelogs

Additionally, when submitting a PR, please fill out at least *some* of the template instead of just submitting a blank one!

imknight:
noted on the preferred way, I think will close this pull request and create a new one ( or shall I just change my pull request? ). Will try to pick up the changelog documentation, let me know if I m doing it wrong.

imknight:
closing this since I working on wrong direction.

PR #1389, opened 2020-10-16
Add an admin tool to handle rescheduling recurring payments

thomasplevy:
## Description
This PR adds an admin tool which will identify orders with an *active* status that do not have an action scheduler recurring payment action associated with it.

Orders which may have encountered errors during creation might end up in this state and while it is possible to manually fix these through the admin panel UI, it's hard to identify orders in this state.

This tool makes bulk updating affected orders trivial.

This tool displays conditionally. It will not show on the tools list if the query does not detect any affected orders.

## How has this been tested?

+ Manual (details below)
+ New unit tests

## Screenshots 

Screenshot_2020-10-16_17-01-57

## Types of changes
+ New feature
+ Minor changes to the `LLMS_Order` class

## Checklist:

## Manual testing

To test this manually:
+ Create one or more orders with recurring payments
+ Navigate to Status -> Scheduled Actions
+ Cancel 1 or more recurring payment actions (observer their scheduled date)
+ Navigate to the Status -> Tools & Utilities
+ Click the new button
+ Check scheduled actions again, these orders should have their actions rescheduled

HS-138030

### DIFF
--- a/includes/admin/tools/class-llms-admin-tool-recurring-payment-rescheduler.php
+++ b/includes/admin/tools/class-llms-admin-tool-recurring-payment-rescheduler.php
@@ -113,7 +113,7 @@ class LLMS_Admin_Tool_Recurring_Payment_Rescheduler extends LLMS_Abstract_Admin_
 
 		foreach ( $this->get_orders() as $id ) {
 			$order = llms_get_post( $id );
- 			$order->maybe_schedule_payment( false );
+			$order->maybe_schedule_payment( false );
 			$order->maybe_schedule_expiration();
 			if ( $order->get_next_scheduled_action_time( 'llms_charge_recurring_payment' ) ) {
 				$orders[] = $id;
@@ -137,7 +137,7 @@ class LLMS_Admin_Tool_Recurring_Payment_Rescheduler extends LLMS_Abstract_Admin_
 
 		global $wpdb;
 
-		return $wpdb->get_results(
+		return $wpdb->get_results( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.NoCaching -- Caching implemented in `get_orders()`.
 			"SELECT p.ID
 			   FROM {$wpdb->posts} AS p
 	      LEFT JOIN {$wpdb->prefix}actionscheduler_actions AS a
--- a/includes/admin/tools/class-llms-admin-tool-recurring-payment-rescheduler.php
+++ b/includes/admin/tools/class-llms-admin-tool-recurring-payment-rescheduler.php
@@ -1,0 +1,172 @@
+<?php
+/**
+ * Admin tool used to autmoatically reschedule recurring orders missing a pending scheduled payment action
+ *
+ * @package LifterLMS/Admin/Tools/Classes
+ *
+ * @since [version]
+ * @version [version]
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * LLMS_Admin_Tool_Batch_Eraser
+ *
+ * @since [version]
+ */
+class LLMS_Admin_Tool_Recurring_Payment_Rescheduler extends LLMS_Abstract_Admin_Tool {
+
+	/**
+	 * Tool ID.
+	 *
+	 * @var string
+	 */
+	protected $id = 'recurring-payment-rescheduler';
+
+	/**
+	 * Retrieve a description of the tool
+	 *
+	 * This is displayed on the right side of the tool's list before the button.
+	 *
+	 * @since [version]
+	 *
+	 * @return string
+	 */
+	protected function get_description() {
+
+		$count = count( $this->get_orders() );
+
+		$desc  = __( 'Schedule recurring payment actions for orders which have not had a recurring payment action scheduled.', 'lifterlms' );
+		$desc .= ' ';
+		// Translators: %d = the number of pending batches.
+		$desc .= sprintf(
+			_n(
+				'There is currently %d order that will have a payment rescheduled.',
+				'There are currently %d orders that will have their payments rescheduled.',
+				$count,
+				'lifterlms'
+			),
+			$count
+		);
+
+		return $desc;
+
+	}
+
+	/**
+	 * Retrieve the tool's label
+	 *
+	 * The label is the tool's title. It's displayed in the left column on the tool's list.
+	 *
+	 * @since [version]
+	 *
+	 * @return string
+	 */
+	protected function get_label() {
+		return __( 'Reschedule Recurring Payments', 'lifterlms' );
+	}
+
+	/**
+	 * Retrieve the tool's button text
+	 *
+	 * @since [version]
+	 *
+	 * @return string
+	 */
+	protected function get_text() {
+		return __( 'Reschedule Payments', 'lifterlms' );
+	}
+
+	/**
+	 * Retrieve a list of orders
+	 *
+	 * @since [version]
+	 *
+	 * @return int[]
+	 */
+	protected function get_orders() {
+
+		$orders = wp_cache_get( $this->id, 'llms_tool_data' );
+		if ( ! $orders ) {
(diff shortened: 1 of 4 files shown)


thomasplevy:
@eri-trabiccolo I know we've got a big PR for your review in #1386 but please prioritize this ahead of that. I'd like to get this shipped Monday if possible, thanks

includes/admin/tools/class-llms-admin-tool-recurring-payment-rescheduler.php
```diff
@@ -0,0 +1,172 @@
...
+ * @package LifterLMS/Admin/Tools/Classes
+ *
+ * @since [version]
+ * @version [version]
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * LLMS_Admin_Tool_Batch_Eraser
```
eri-trabiccolo:
```suggestion
 * LLMS_Admin_Tool_Recurring_Payment_Rescheduler class
```

includes/admin/tools/class-llms-admin-tool-recurring-payment-rescheduler.php
```diff
@@ -0,0 +1,172 @@
...
+
+		return $orders;
+
+	}
+
+	/**
+	 * Schedules payments and expiration for an order
+	 *
+	 * Retrieves orders from the `get_orders()` method and schedules a recurring payment
+	 * and expiration action based on it's existing calculated order data.
```
eri-trabiccolo:
```suggestion
	 * and expiration action based on its existing calculated order data.
```

includes/admin/tools/class-llms-admin-tool-recurring-payment-rescheduler.php
```diff
@@ -0,0 +1,172 @@
...
+
+		return $wpdb->get_results( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.NoCaching -- Caching implemented in `get_orders()`.
+			"SELECT p.ID
+			   FROM {$wpdb->posts} AS p
+	      LEFT JOIN {$wpdb->prefix}actionscheduler_actions AS a
+			     ON a.args   = CONCAT( '{\"order_id\":', p.ID, '}' )
+			    AND a.hook   = 'llms_charge_recurring_payment'
+			    AND a.status = 'pending'
+			  WHERE 1
+			    AND p.post_type   = 'llms_order'
```
eri-trabiccolo, line 148: this just for a matter of readability, right?!

includes/models/model.llms.order.php
```diff
@@ -806,6 +806,18 @@ public function get_next_payment_due_date( $format = 'Y-m-d H:i:s' ) {
 
 	}
 
+	/**
+	 * Retrieve the timestamp of the next scheduled event for a given action
+	 *
+	 * @since [version]
+	 *
+	 * @param string $action Action hook ID. Core actions are "llms_charge_recurring_payment" and "llms_access_plan_expiration".
+	 * @return int|false Returns the timestamp of the next action as an integer or `false` when no action exist.
```
eri-trabiccolo, line 815:
```suggestion
	 * @return int|bool The timestamp for the next occurrence of a pending scheduled action, `true` for an async or in-progress action or `false` if there is no matching action.
```
looks like `as_next_scheduled_action` can return `true` for an async or in-progress action

https://github.com/woocommerce/action-scheduler/blob/v3.1.5/functions.php#L150

includes/admin/tools/class-llms-admin-tool-recurring-payment-rescheduler.php
```diff
@@ -0,0 +1,172 @@
...
+			    AND p.post_status = 'llms-active'
+			    AND a.action_id IS NULL
+			  LIMIT 100
+			;"
+		);
```
eri-trabiccolo:
```suggestion
		); // no-cache ok -- Caching implemented in `get_orders()`.
```
And you've not added the ignore for the direct call as we're about to get rid of that rule, right?

includes/admin/tools/class-llms-admin-tool-recurring-payment-rescheduler.php
```diff
@@ -0,0 +1,172 @@
...
+	 *
+	 * @since [version]
+	 *
+	 * @return object[]
+	 */
+	protected function query_orders() {
+
+		global $wpdb;
+
+		return $wpdb->get_results( // phpcs:ignore WordPress.DB.DirectDatabaseQuery.NoCaching -- Caching implemented in `get_orders()`.
```
eri-trabiccolo:
```suggestion
		return $wpdb->get_results(
```

eri-trabiccolo (changes requested):
Aside for minor doc changes this looks good, BUT if we assume that the reason of the issue we're fixing here was due to the log files I think we should also run a `do_action( 'llms_backup_logs' );` at the beginning of the `handle()` method, what do you think?

tests/phpunit/unit-tests/admin/tools/class-llms-test-admin-tool-recurring-payment-rescheduler.php
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	/**
+	 * Test should_load()
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	protected function test_should_load() {
```
eri-trabiccolo:
```suggestion
	public function test_should_load() {
```

eri-trabiccolo (changes requested):
Forgot this, `protected` visibility makes the tests fail.

includes/admin/tools/class-llms-admin-tool-recurring-payment-rescheduler.php
thomasplevy: yes, I like being able to align query parts like this. I picked this up from the WP core a while ago and I like it. To my knowledge there's no performance impact but you know I might be wrong about this.

I'm not very attached to it so if you think we should remove I can

includes/models/model.llms.order.php
thomasplevy: didn't realize that. Thanks

includes/admin/tools/class-llms-admin-tool-recurring-payment-rescheduler.php
thomasplevy: Right.

I for the life of me couldn't figure out why the shorthand wasn't working. It has to be on the END! Thanks for teaching me this.

tests/phpunit/unit-tests/admin/tools/class-llms-test-admin-tool-recurring-payment-rescheduler.php
thomasplevy: I pushed real quick at the end of the day and must have missed the warning locally, sorry.

thomasplevy:
> ...BUT if we assume that the reason of the issue we're fixing here was due to the log files I think we should also run a `do_action( 'llms_backup_logs' );` at the beginning of the `handle()` method, what do you think?

I don't think it's necessary, the cron runs daily and based on the way the crons work if someone is affected it'll either have already run the backup cron or it will run shortly after upgrading.

includes/admin/tools/class-llms-admin-tool-recurring-payment-rescheduler.php
eri-trabiccolo: I don't think it has any performance drawbacks either, and I understand your point + I remember I saw this in core too.

includes/admin/tools/class-llms-admin-tool-recurring-payment-rescheduler.php
eri-trabiccolo: I learned it last week during the doc party ahah.

PR #1393, opened 2020-10-18
Add support for multiple roles in LLMS_User_Permissions::editable_roles()

### LINKED ISSUE #1356: Multiple Role affecting Admin Permissions - take 2
### Reproduction Steps
similar to https://github.com/gocodebox/lifterlms/issues/1183

+ Install any plugin that lets you have multiple roles
+ On an admin account (not your main one) add a secondary role of Instructor
+ Jump into that account
+ Go back to the Users page and try to edit a user
+ Look at the Role setting

### Expected Behavior

+ The Administrator role should allow you to change the user role to whatever
+ The select should show the current user's role as selected option

### Actual Behavior

+ You can only change the user role to Instructor Assistant
+ The select shows "No role for this site" as selected option

**The workaround** would just be to not add the Instructor role to an administrator (or llms manager), or to remove it if accidentally you added it - to do the latest you need another administrator of course.

### Error Messages / Logs

n/a

LifterLMS 4.4.4
WordPress 5.5.1

+ [x] Locally
+ [x] On a staging site
+ [x] On a production website

n/a

### Related User Information
HS-136130

daniel-shuy:
## Description

Fixes #1356 

Whitelists editable roles using addition instead of subtraction.

## Checklist:

### DIFF
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -47,6 +47,7 @@ class LLMS_User_Permissions {
 	 * @since 3.13.0
 	 * @since 3.34.0 Moved the `llms_editable_roles` filter to the class method get_editable_roles().
 	 * @since 3.37.14 Use strict comparison.
+	 * @since [version] Better handling of users with multiple roles.
 	 *
 	 * @link https://codex.wordpress.org/Plugin_API/Filter_Reference/editable_roles
 	 *
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -60,17 +60,17 @@ class LLMS_User_Permissions {
 			return $all_roles;
 		}
 
-		$roles = array();
+		$user       = wp_get_current_user();
+		$user_roles = $user->roles;
 
-		$user = wp_get_current_user();
+		if ( in_array( 'administrator', $user_roles, true ) ) {
+			return $all_roles;
+		}
 
 		$editable_roles = self::get_editable_roles();
 
-		foreach ( $user->roles as $user_role ) {
-			if ( 'administrator' === $user_role ) {
-				return $all_roles;
-			}
-
+		$roles = array();
+		foreach ( $user_roles as $user_role ) {
 			if ( isset( $editable_roles[ $user_role ] ) ) {
 				$roles = array_merge( $roles, $editable_roles[ $user_role ] );
 			}
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -69,6 +69,10 @@ class LLMS_User_Permissions {
 
 		$editable_roles = self::get_editable_roles();
 
+		if ( empty( array_intersect( $user_roles, array_keys( $editable_roles ) ) ) ) {
+			return $all_roles;
+		}
+
 		$roles = array();
 		foreach ( $user_roles as $user_role ) {
 			if ( isset( $editable_roles[ $user_role ] ) ) {
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -72,7 +72,7 @@ class LLMS_User_Permissions {
 
 		$roles = array_unique( $roles );
 
-		foreach ( $all_roles as $role ) {
+		foreach ( array_keys( $all_roles ) as $role ) {
 			if ( ! in_array( $role, $roles, true ) ) {
 				unset( $all_roles[ $role ] );
 			}
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -72,7 +72,7 @@ class LLMS_User_Permissions {
 		$roles = array_unique( $roles );
 
 		foreach ( $all_roles as $role ) {
-			if ( ! in_array( $role, $roles ) ) {
+			if ( ! in_array( $role, $roles, true ) ) {
 				unset( $all_roles[ $role ] );
 			}
 		}
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -55,6 +55,8 @@ class LLMS_User_Permissions {
 	 */
 	public function editable_roles( $all_roles ) {
 
+		$roles = array();
+
 		$user = wp_get_current_user();
 
 		$lms_roles = self::get_editable_roles();
@@ -63,11 +65,15 @@ class LLMS_User_Permissions {
 
 			if ( in_array( $role, $user->roles, true ) ) {
 
-				foreach ( $all_roles as $the_role => $caps ) {
-					if ( ! in_array( $the_role, $allowed_roles, true ) ) {
-						unset( $all_roles[ $the_role ] );
-					}
-				}
+				$roles = array_merge( $roles, $allowed_roles );
+			}
+		}
+
+		$roles = array_unique( $roles );
+
+		foreach ( $all_roles as $role ) {
+			if ( ! in_array( $role, $roles, true ) ) {
+				unset( $all_roles[ $role ] );
 			}
 		}
 
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -56,6 +56,10 @@ class LLMS_User_Permissions {
 	 */
 	public function editable_roles( $all_roles ) {
 
+		if ( is_multisite() && is_super_admin() ) {
+			return $all_roles;
+		}
+
 		$roles = array();
 
 		$user = wp_get_current_user();
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -62,10 +62,14 @@ class LLMS_User_Permissions {
 
 		$lms_roles = self::get_editable_roles();
 
-		foreach ( $lms_roles as $role => $allowed_roles ) {
+		foreach ( $user->roles as $user_role ) {
+			if ( 'administrator' === $user_role ) {
+				return array_keys( $all_roles );
+			}
 
-			if ( in_array( $role, $user->roles, true ) ) {
+			$allowed_roles = $lms_roles[ $user_role ];
 
+			if ( ! is_null( $allowed_roles ) ) {
 				$roles = array_merge( $roles, $allowed_roles );
 			}
 		}
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -55,6 +55,8 @@ class LLMS_User_Permissions {
 	 */
 	public function editable_roles( $all_roles ) {
 
+		$roles = array();
+
 		$user = wp_get_current_user();
 
 		$lms_roles = self::get_editable_roles();
@@ -63,11 +65,15 @@ class LLMS_User_Permissions {
 
 			if ( in_array( $role, $user->roles, true ) ) {
 
-				foreach ( $all_roles as $the_role => $caps ) {
-					if ( ! in_array( $the_role, $allowed_roles, true ) ) {
-						unset( $all_roles[ $the_role ] );
-					}
-				}
+				$roles = array_merge( $roles, $allowed_roles );
+			}
+		}
+
+		$roles = array_unique( $roles );
+
+		foreach ( $all_roles as $role ) {
+			if ( ! in_array( $role, $roles ) ) {
+				unset( $all_roles[ $role ] );
 			}
 		}
 
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -64,7 +64,7 @@ class LLMS_User_Permissions {
 
 		foreach ( $user->roles as $user_role ) {
 			if ( 'administrator' === $user_role ) {
-				return array_keys( $all_roles );
+				return $all_roles;
 			}
 
 			if ( isset( $lms_roles[ $user_role ] ) ) {
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Classes
  *
  * @since 3.13.0
- * @version 3.41.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -81,7 +81,7 @@ class LLMS_User_Permissions {
 		$llms_roles = array_keys( LLMS_Roles::get_roles() );
 
 		foreach ( array_keys( $all_roles ) as $role ) {
-			// only filter out LLMS roles.
+			// Only filter out LLMS roles.
 			if ( ! in_array( $role, $roles, true ) && in_array( $role, $llms_roles, true ) ) {
 				unset( $all_roles[ $role ] );
 			}
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -82,11 +82,8 @@ class LLMS_User_Permissions {
 
 		$roles = array_unique( $roles );
 
-		$llms_roles = array_keys( LLMS_Roles::get_roles() );
-
 		foreach ( array_keys( $all_roles ) as $role ) {
-			// Only filter out LLMS roles.
-			if ( ! in_array( $role, $roles, true ) && in_array( $role, $llms_roles, true ) ) {
+			if ( ! in_array( $role, $roles, true ) ) {
 				unset( $all_roles[ $role ] );
 			}
 		}
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -60,22 +60,25 @@ class LLMS_User_Permissions {
 
 		$user = wp_get_current_user();
 
-		$lms_roles = self::get_editable_roles();
+		$editable_roles = self::get_editable_roles();
 
 		foreach ( $user->roles as $user_role ) {
 			if ( 'administrator' === $user_role ) {
 				return $all_roles;
 			}
 
-			if ( isset( $lms_roles[ $user_role ] ) ) {
-				$roles = array_merge( $roles, $lms_roles[ $user_role ] );
+			if ( isset( $editable_roles[ $user_role ] ) ) {
+				$roles = array_merge( $roles, $editable_roles[ $user_role ] );
 			}
 		}
 
 		$roles = array_unique( $roles );
 
+		$llms_roles = array_keys( LLMS_Roles::get_roles() );
+
 		foreach ( array_keys( $all_roles ) as $role ) {
-			if ( ! in_array( $role, $roles, true ) ) {
+			// only filter out LLMS roles.
+			if ( ! in_array( $role, $roles, true ) && in_array( $role, $llms_roles, true ) ) {
 				unset( $all_roles[ $role ] );
 			}
 		}
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -67,10 +67,8 @@ class LLMS_User_Permissions {
 				return array_keys( $all_roles );
 			}
 
-			$allowed_roles = $lms_roles[ $user_role ];
-
-			if ( ! is_null( $allowed_roles ) ) {
-				$roles = array_merge( $roles, $allowed_roles );
+			if ( isset( $lms_roles[ $user_role ] ) ) {
+				$roles = array_merge( $roles, $lms_roles[ $user_role ] );
 			}
 		}
 
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -55,8 +55,6 @@ class LLMS_User_Permissions {
 	 */
 	public function editable_roles( $all_roles ) {
 
-		$roles = array();
-
 		$user = wp_get_current_user();
 
 		$lms_roles = self::get_editable_roles();
@@ -65,15 +63,11 @@ class LLMS_User_Permissions {
 
 			if ( in_array( $role, $user->roles, true ) ) {
 
-				$roles = array_merge( $roles, $allowed_roles );
-			}
-		}
-
-		$roles = array_unique( $roles );
-
-		foreach ( $all_roles as $role ) {
-			if ( ! in_array( $role, $roles, true ) ) {
-				unset( $all_roles[ $role ] );
+				foreach ( $all_roles as $the_role => $caps ) {
+					if ( ! in_array( $the_role, $allowed_roles, true ) ) {
+						unset( $all_roles[ $the_role ] );
+					}
+				}
 			}
 		}
 
--- a/tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
+++ b/tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
@@ -178,13 +178,13 @@ class LLMS_Test_User_Permissions extends LLMS_UnitTestCase {
 		$all_roles = wp_roles()->roles;
 
 		wp_set_current_user( $users['assistant'] );
-		$assistant_editable_roles = LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) );
+		$assistant_editable_roles = array_keys ( LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) ) );
 
 		// assert that assistants cannot edit any roles
 		$this->assertEmpty( $assistant_editable_roles );
 
 		wp_set_current_user( $users['admin'] );
-		$administrator_editable_roles = LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) );
+		$administrator_editable_roles = array_keys ( LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) ) );
 
 		// assert that administrator can edit all roles
 		foreach ( array_keys( $all_roles ) as $role ) {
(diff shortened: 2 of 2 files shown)


thomasplevy:
@daniel-shuy 

Thanks for your contribution!

This looks quite good but before proceeding we'll need to add some tests to make sure this doesn't end up with a future regression. You can see a test for the "take 1" version of this issue here: https://github.com/gocodebox/lifterlms/blob/153304618c97da71b4743c2b78e55e0731a8ca3f/tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php#L143-L165

This will need something like that to proceed.

I never like to beat external contributors up about adding test cases but I can't ship anything until we're internally confident and testing is part of that.

I'll leave this open for now and if you'd like to add tests we'd greatly appreciate it otherwise either @eri-trabiccolo or myself will get to it when we're freed up from our current projects.

Thanks again!

daniel-shuy:
@thomasplevy Done! I've added a unit test, and to prove that it works I even reverted my changes to verify that the test fails without the fix.

thomasplevy:
@daniel-shuy beautiful thanks so much!

We'll have a closer look at this soon

includes/class.llms.user.permissions.php
```diff
@@ -63,11 +66,15 @@ public function editable_roles( $all_roles ) {
...
-						unset( $all_roles[ $the_role ] );
-					}
-				}
+				$roles = array_merge( $roles, $allowed_roles );
+			}
+		}
+
+		$roles = array_unique( $roles );
+
+		foreach ( $all_roles as $role ) {
```
eri-trabiccolo: `$all_roles` is of the type:
```
[
'role_1' => [ 'name' => 'Some name', 'cababilities' => [...]],
...
]
```

so here you'd want something like this:
```suggestion
		foreach ( array_keys( $all_roles ) as $role ) {
```
otherwise this method it's just doing nothing... :(

Additionally, if you visit the user's page (wp-admin/users.php) as an administrator for example, and you have the debug logging enabled, you'll see a bunch of warnings of this type.
`[21-Oct-2020 14:12:55 UTC] PHP Warning:  Illegal offset type in unset in /srv/www/wordpress-one/public_html/wp-content/plugins/lifterlms/includes/class.llms.user.permissions.php on line 77`

The new added tests pass only because a "wrong" `$all_roles` is passed to this method. :D

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
```diff
@@ -164,6 +164,40 @@ public function test_user_can_manage_user_multiple_roles() {
...
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function test_editable_roles_multiple_roles() {
+
+		$users = $this->create_mock_users();
+
+		$all_roles = LLMS_Roles::get_roles();
```
eri-trabiccolo: The array of roles to feed the method `editable_roles` must be similar to the one passed by the `'editable_roles'` filter: https://developer.wordpress.org/reference/hooks/editable_roles/
so:

```suggestion
		$all_roles = wp_roles()->roles;
```

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
```diff
@@ -164,6 +164,40 @@ public function test_user_can_manage_user_multiple_roles() {
...
+		$editable_roles = LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) );
+
+		// assert that lms_manager with instructor role has editable roles from both roles
+		foreach ( $lms_manager_editable_roles as $lms_manager_editable_role ) {
+			$this->assertContains( $lms_manager_editable_role, $editable_roles );
+		}
+		foreach ( $instructor_editable_roles as $instructor_editable_role ) {
+			$this->assertContains( $instructor_editable_role, $editable_roles );
+		}
+
```
eri-trabiccolo, line 247: this is not exactly what we want to fix here.
We want that, for example, users with both the roles `'administrator'` and `'instructor'` can still edit all the roles an `'administrator'` initially could...

eri-trabiccolo (changes requested):
Aside from the specific code changes requested, this PR, as it is - but fixing the bit I highlighted `foreach ( array_keys( $all_roles ) as $role ) {` - doesn't fix the issue it aims to fix, plus it introduces a regression, as `administrator`s will not be able to alter the role of any user.

includes/class.llms.user.permissions.php
daniel-shuy: Oops, I can't believe I made such a silly mistake, sorry and thanks

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
daniel-shuy: This implementation works for any 2 roles, except `administrator`, because `administrator` is not configured in `get_editable_roles()`. I think the best way might be to specifically check for `administrator`, because I don't think we want to list all the roles the `administrator` can edit in `get_editable_roles()`, or it would require updating that method everytime a new role is added

daniel-shuy: I've catered for the `administrator` role, and I've added regression tests

includes/class.llms.user.permissions.php
```diff
@@ -55,19 +56,27 @@ public function __construct() {
...
+		$roles = array();
+
 		$user = wp_get_current_user();
 
 		$lms_roles = self::get_editable_roles();
 
-		foreach ( $lms_roles as $role => $allowed_roles ) {
+		foreach ( $user->roles as $user_role ) {
+			if ( 'administrator' === $user_role ) {
+				return array_keys( $all_roles );
```
eri-trabiccolo:
```suggestion
				return $all_roles;
```
why do return the keys of the input array?
if you visit, as administrator, the wp-admin/users.php you're flooded (depending on the number of users of course) of php notices like:
`PHP Warning:  Illegal string offset 'name' in /srv/www/wordpress-one/public_html/wp-admin/includes/template.php on line 938`
and the roles select in the user edit screen looks like:
Schermata 2020-10-23 alle 16 44 47

eri-trabiccolo (changes requested):
While this might even work - after the bug I highlighted is fixed - fixing #1356 I'm not sure this is the way to go... Probably we just cannot fix #1356 for all the (multiple) roles.

Unfortunately the WP ecosystem doesn't allow us to assume that only the `administrator` role must be skipped here, there can be other plugins adding custom roles and wanting them to have full control on other roles, and we cannot prevent this - this PR potentially prevents it. 

Additionally, this code introduces a regression in multisites, because it will prevent network admins to edit roles on sites for which they are not administrators (although this can be easily fixed by adding checks on if it's multisite etc.). LifterLMS is not officially compatible with multisites but I don't think we should introduce such a regression either.

I think the "allow list" is not the best approach here, the WP ecosystem is too vary.
We need to move carefully and be sure we only alter our roles.
So my idea is that we have two scenarios here:
1) the user we're parsing has only one role, and its role is one of 'lms_manager', 'instructor' (meaning the keys of `LLMS_User_Permissions::get_editable_roles()`), so this user will be able to edit the roles mapped by `LLMS_User_Permissions::get_editable_roles()`.
2) the user we're parsing has more than one role and one of them is one of 'lms_manager', 'instructor', in this case we basically have no real way to know what role this user should be able to edit... I see to approaches here:
2a) our role ('lms_manager', 'instructor') wins and determines what other role the user can edit.
2b) the other role wins and we just do nothing, at most we can add our custom editable roles in case the `$all_roles` passed to the filter was already modified by a plugin and didn't contain our allowed ones.

in 2a) we can safely exclude administrators and super admins in multisite - it'll be incomplete but better than nothing.

what do you think @thomasplevy ?

includes/class.llms.user.permissions.php
daniel-shuy: Ah, sorry, fixed it

daniel-shuy:
@eri-trabiccolo What if the filter only filters out LLMS roles? Will this break regression for single internal roles (Scenario 1)?

If the filter only filters out LLMS roles, then it will still return roles for other plugins, which should then be filtered out by `editable_roles` filters from the other respective plugins.

eri-trabiccolo:
@daniel-shuy 
this way you're preventing other roles (custom or anyways with administrative privileges) to assign a LifterLMS role to a user.
e.g. a network administrator who's not an administrator of the site A, cannot turn the role of a Subscriber of the site A into a Student (which is a LifterLMS role).

daniel-shuy:
> this way you're preventing other roles (custom or anyways with administrative privileges) to assign a LifterLMS role to a user.
e.g. a network administrator who's not an administrator of the site A, cannot turn the role of a Subscriber of the site A into a Student (which is a LifterLMS role).

@eri-trabiccolo I see. Hmm, do you know how other WordPress plugins handle this issue?

daniel-shuy:
@eri-trabiccolo when you say network administrator, do you mean Multisite Super Admins? If so, maybe we can check if the user has the `manage_network_users` capability? (according to https://wordpress.org/support/article/roles-and-capabilities/#super-admin, only Super Admins have the `manage_network_users` capability by default)

eri-trabiccolo:
@daniel-shuy yeah I meant that, sorry.
Yeah we can also use this other kind of check
`	if ( is_multisite() && is_super_admin() ) `

daniel-shuy:
@eri-trabiccolo I've made the changes, thanks!

includes/class.llms.user.permissions.php
```diff
@@ -55,19 +56,34 @@ public function __construct() {
...
+			}
+		}
+
+		$roles = array_unique( $roles );
+
+		$llms_roles = array_keys( LLMS_Roles::get_roles() );
+
+		foreach ( array_keys( $all_roles ) as $role ) {
+			// only filter out LLMS roles.
+			if ( ! in_array( $role, $roles, true ) && in_array( $role, $llms_roles, true ) ) {
```
eri-trabiccolo: I don't understand the purpose of this condition.
Say I'm an user with a custom role who can edit roles, and this role is 'my_custom_role' ok?
For me the variable `$roles` at line 79 above will be an empty array.
So: `! in_array( $role, $roles, true )` will be always `true`, because `$roles` is empty, then `in_array( $role, $llms_roles, true )` will be true only for the LifterLMS defined roles (`LLMS_Roles::get_roles()`). 
So this means that you're preventing 'my_custom_role' from changing the role of a Subscriber to Student (which is an LLMS_Role), for example.
We just don't need this behavior, why we should? Is there anything I'm missing?
What's your purpose here?

eri-trabiccolo (changes requested):
I think I'm missing the purpose of a piece of code...

includes/class.llms.user.permissions.php
daniel-shuy: This is to handle the case where the user has multiple roles that are defined in `LLMS_User_Permissions::get_editable_roles()`. Previously, if the user has both `lms_manager` and `instructor`, even though the `lms_manager` role should allow the user to edit `instructor`, `instructors_assistant`, `lms_manager`, `student`, the user will only be allowed to edit `instructors_assistant`, because it would remove all roles that are not mapped to `instructor` in `LLMS_User_Permissions::get_editable_roles()`.

With this change it would instead remove all LLMS roles that are not mapped to any of the roles that the user has that are defined in `LLMS_User_Permissions::get_editable_roles()`. This means that any roles that need to be able to edit LLMS roles (except Super Admin and `administrator`) must be configured in `LLMS_User_Permissions::get_editable_roles()`, eg. if `my_custom_role` can edit `student`:
```php
		$roles = apply_filters(
			'llms_editable_roles',
			array(
				'lms_manager' => array( 'instructor', 'instructors_assistant', 'lms_manager', 'student' ),
				'instructor'  => array( 'instructors_assistant' ),
                                'my_custom_role' => array( 'student' ),
			)
		);
```

eri-trabiccolo: Ah I see, so I was missing the "configuration" idea.
Mmm I think that would be fine for users with multiple roles, when at least one of them is a key of the `LLMS_User_Permissions::get_editable_roles()` array.
I still don't think that if I'm a user with only one plenipotentiary role, my_custom_role, LifterLMS should forbid me to promote a Subscriber to Student unless I use the filter above.

So I think that if
`array_intersect_key( array_flip( $user->roles ), LLMS_User_Permissions::get_editable_roles() )` is `empty` we should return `$all_roles` very early.

includes/class.llms.user.permissions.php
```diff
@@ -55,19 +56,34 @@ public function __construct() {
...
+				$roles = array_merge( $roles, $editable_roles[ $user_role ] );
+			}
+		}
+
+		$roles = array_unique( $roles );
+
+		$llms_roles = array_keys( LLMS_Roles::get_roles() );
+
+		foreach ( array_keys( $all_roles ) as $role ) {
+			// only filter out LLMS roles.
```
eri-trabiccolo:
```suggestion
			// Only filter out LLMS roles.
```

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
```diff
@@ -164,6 +164,83 @@ public function test_user_can_manage_user_multiple_roles() {
...
+		$users = $this->create_mock_users();
+
+		$all_roles = wp_roles()->roles;
+
+		$llms_roles = LLMS_Roles::get_roles();
+
+		wp_set_current_user( $users['assistant'] );
+		$assistant_editable_roles = array_keys ( LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) ) );
+
+		// assert that assistants cannot edit any LLMS roles
```
eri-trabiccolo:
```suggestion
		// Assert that assistants cannot edit any LLMS roles.
```

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
```diff
@@ -164,6 +164,83 @@ public function test_user_can_manage_user_multiple_roles() {
...
+		foreach ( array_keys( $llms_roles ) as $llms_role ) {
+			$this->assertNotContains( $llms_role, $assistant_editable_roles );
+		}
+		// assert that assistants can edit external roles
```
eri-trabiccolo:
```suggestion
		// Assert that assistants can edit external roles.
```

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
```diff
@@ -164,6 +164,83 @@ public function test_user_can_manage_user_multiple_roles() {
...
+		$this->assertNotEmpty( $assistant_editable_roles );
+
+		wp_set_current_user( $users['admin'] );
+		$administrator_editable_roles = array_keys ( LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) ) );
+
+		// assert that administrator can edit all roles
```
eri-trabiccolo:
```suggestion
		// Assert that administrator can edit all roles.
```

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
```diff
@@ -164,6 +164,83 @@ public function test_user_can_manage_user_multiple_roles() {
...
+
+		wp_set_current_user( $users['instructor'] );
+		$instructor_editable_roles = array_keys ( LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) ) );
+
+		wp_set_current_user( $users['lms_manager'] );
+		$user = wp_get_current_user();
+		$user->add_role( 'instructor' );
+		$lms_manager_instructor_editable_roles = array_keys ( LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) ) );
+
+		// assert that lms_manager with instructor role has editable roles from both roles
```
eri-trabiccolo:
```suggestion
		// Assert that lms_manager with instructor role has editable roles from both roles.
```

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
```diff
@@ -164,6 +164,83 @@ public function test_user_can_manage_user_multiple_roles() {
...
+		foreach ( $instructor_editable_roles as $instructor_editable_role ) {
+			$this->assertContains( $instructor_editable_role, $lms_manager_instructor_editable_roles );
+		}
+
+		wp_set_current_user( $users['admin'] );
+		$user = wp_get_current_user();
+		$user->add_role( 'instructor' );
+		$administrator_instructor_editable_roles = array_keys ( LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) ) );
+
+		// assert that administrator with instructor role can edit all roles
```
eri-trabiccolo:
```suggestion
		// Assert that administrator with instructor role can edit all roles.
```

eri-trabiccolo (changes requested):
Only minor doc cs changes plus what i said here (with additional tests):
https://github.com/gocodebox/lifterlms/pull/1393/files/c5d43ba16adab022577dc95cfc5b8d85a851465c#r514975272

includes/class.llms.user.permissions.php
daniel-shuy: >So I think that if
array_intersect_key( array_flip( $user->roles ), LLMS_User_Permissions::get_editable_roles() ) is empty we should return $all_roles very early

Wouldn't that allow for example `student`s to edit all LLMS roles?

eri-trabiccolo: I think that the best thing is that you test the code so that you're not forced to trust my words.
`student`s cannot edit users at all so they are not impacted by this piece of code.

Also, that method, as you can see, doesn't add anything to `$all_roles`, at most it removes some role so...

daniel-shuy: Ok, so we allow all unmapped roles to edit all LLMS roles, and we rely on the edit user check to make sure that users with specific capabilities can edit users

daniel-shuy:
Rebased onto `dev` (a test was failing)

includes/class.llms.user.permissions.php
eri-trabiccolo: yes, that's how it is in the current stable, and as far as I know, there's no need to alter this behavior.

> we rely on the edit user check to make sure that users with specific capabilities can edit users

For users who cannot edit other users the filter hook `'editable_role'` will just not be fired so...

daniel-shuy: I've made the changes accordingly and updated the tests

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
```diff
@@ -164,6 +164,97 @@ public function test_user_can_manage_user_multiple_roles() {
...
+		$users = $this->create_mock_users();
+
+		$all_roles = wp_roles()->roles;
+
+		$editable_roles = LLMS_Unit_Test_Util::call_method( $this->obj, 'get_editable_roles');
+
+		wp_set_current_user( $users['lms_manager'] );
+		$lms_manager_editable_roles = array_keys ( LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) ) );
+
+		// Assert that lms_managers can edit mapped roles
```
eri-trabiccolo:
```suggestion
		// Assert that lms_managers can edit mapped roles.
```
I think my previous review included both the first word capitalization and the ending dot...
am I wrong?

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
```diff
@@ -164,6 +164,97 @@ public function test_user_can_manage_user_multiple_roles() {
...
+		foreach ( $editable_roles['lms_manager'] as $editable_role ) {
+			$this->assertContains( $editable_role, $lms_manager_editable_roles );
+		}
+
+		wp_set_current_user( $users['instructor'] );
+		$instructor_editable_roles = array_keys ( LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) ) );
+
+		// Assert that instructor can edit mapped roles
```
eri-trabiccolo:
```suggestion
		// Assert that instructor can edit mapped roles.
```

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
```diff
@@ -164,6 +164,97 @@ public function test_user_can_manage_user_multiple_roles() {
...
+		foreach ( $editable_roles['instructor'] as $editable_role ) {
+			$this->assertContains( $editable_role, $instructor_editable_roles );
+		}
+
+		wp_set_current_user( $users['assistant'] );
+		$assistant_editable_roles = array_keys ( LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) ) );
+
+		// Assert that assistants can edit all roles
```
eri-trabiccolo:
```suggestion
		// Assert that assistants can edit all roles.
```

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
```diff
@@ -164,6 +164,97 @@ public function test_user_can_manage_user_multiple_roles() {
...
+		foreach ( array_keys( $all_roles ) as $role ) {
+			$this->assertContains( $role, $assistant_editable_roles );
+		}
+
+		wp_set_current_user( $users['admin'] );
+		$administrator_editable_roles = array_keys ( LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) ) );
+
+		// Assert that administrator can edit all roles
```
eri-trabiccolo:
```suggestion
		// Assert that administrator can edit all roles.
```

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
```diff
@@ -164,6 +164,97 @@ public function test_user_can_manage_user_multiple_roles() {
...
+
+		wp_set_current_user( $users['instructor'] );
+		$instructor_editable_roles = array_keys ( LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) ) );
+
+		wp_set_current_user( $users['lms_manager'] );
+		$user = wp_get_current_user();
+		$user->add_role( 'instructor' );
+		$lms_manager_instructor_editable_roles = array_keys ( LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) ) );
+
+		// Assert that lms_manager with instructor role has editable roles from both roles
```
eri-trabiccolo:
```suggestion
		// Assert that lms_manager with instructor role has editable roles from both roles.
```

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
```diff
@@ -164,6 +164,97 @@ public function test_user_can_manage_user_multiple_roles() {
...
+		foreach ( $instructor_editable_roles as $instructor_editable_role ) {
+			$this->assertContains( $instructor_editable_role, $lms_manager_instructor_editable_roles );
+		}
+
+		wp_set_current_user( $users['admin'] );
+		$user = wp_get_current_user();
+		$user->add_role( 'instructor' );
+		$administrator_instructor_editable_roles = array_keys ( LLMS_Unit_Test_Util::call_method( $this->obj, 'editable_roles', array( $all_roles ) ) );
+
+		// Assert that administrator with instructor role can edit all roles
```
eri-trabiccolo:
```suggestion
		// Assert that administrator with instructor role can edit all roles.
```

eri-trabiccolo (changes requested):
I think there are missing trailing dots in the test's comments.

Thanks, I'm going to test this PR code a little bit!

tests/phpunit/unit-tests/user/class-llms-test-user-permissions.php
daniel-shuy: Oh sorry, didn't notice, will fix it

eri-trabiccolo:
This PR introduces (at least) this regression:
LMS_Manager and Instructor roles now can assign whatever role to an user, while before they could only assign the roles mapped in:
`LLMS_User_Permissions::get_editable_roles()`.

Which is the scenario 1. I described here: https://github.com/gocodebox/lifterlms/pull/1393#pullrequestreview-515736698
> So my idea is that we have two scenarios here:
>
> 1) the user we're parsing has only one role, and its role is one of 'lms_manager', 'instructor' (meaning the keys of LLMS_User_Permissions::get_editable_roles()), so this user will be able to edit the roles mapped by LLMS_User_Permissions::get_editable_roles().
> 2) the user we're parsing has more than one role and one of them is one of 'lms_manager', 'instructor', in this case we basically have no real way to know what role this user should be able to edit... I see to approaches here:
> 2a) our role ('lms_manager', 'instructor') wins and determines what other role the user can edit.
> 2b) the other role wins and we just do nothing, at most we can add our custom editable roles in case the $all_roles passed to the filter was already modified by a plugin and didn't contain our allowed ones.

Actually 1. should be more generalized.
If the current user roles are completely mapped by `LLMS_User_Permissions::get_editable_roles()` (which means:
a) the user's role is 1 and it's LMS_Manager
b) the user's role is 1 and it's Instructor
c) the user's roles are 2 and they are LMS_Manager and Instructor)
then the roles the user can manage are defined by `LLMS_User_Permissions::get_editable_roles()`.

eri-trabiccolo (changes requested):
See 
https://github.com/gocodebox/lifterlms/pull/1393#issuecomment-719574004

daniel-shuy:
> LMS_Manager and Instructor roles now can assign whatever role to an user, while before they could only assign the roles mapped in:
`LLMS_User_Permissions::get_editable_roles()`

you mean external (non-LLMS) roles?

If I restrict roles that are mapped in `LLMS_User_Permissions::get_editable_roles()` to not be able to edit external (non-LLMS) roles, then we run into the dilemma of the 2nd scenario that you mentioned:
> the user we're parsing has more than one role and one of them is one of 'lms_manager', 'instructor', in this case we basically have no real way to know what role this user should be able to edit... I see to approaches here:
2a) our role ('lms_manager', 'instructor') wins and determines what other role the user can edit.
2b) the other role wins and we just do nothing, at most we can add our custom editable roles in case the $all_roles passed to the filter was already modified by a plugin and didn't contain our allowed ones.

So should it be `2a` or `2b`?

daniel-shuy:
> in 2a) we can safely exclude administrators and super admins in multisite

Now that administrators and multisite super admins are excluded, looks like its `2a` then. I'll go with that then.

eri-trabiccolo:
Hi @daniel-shuy sorry I've been pretty busy this week.
Will take a look at this hopefully by the beginning of the next week.
Thanks for your contribution!
Best

daniel-shuy:
Hopefully this can still make it for the LifterLMS Hacktoberfest event :P

thomasplevy:
Yes it does, we haven't been able to pull it in but we're going to count this as a contribution.

I'll be tallying contributions and we'll be drawing a winner soon. I'll be posting the winner to #development on our slack channel and notifying the winner as a comment on one of their PRs here on GitHub.

Thanks again

eri-trabiccolo (changes requested):
Hi @daniel-shuy sorry for the late review.
This looks good to me, it fixes #1356 and doesn't seem to introduce any drawbacks!
Many thanks.
May I ask you to just change the `version` tag here:
https://github.com/gocodebox/lifterlms/blob/d7810d44d3a7277ceb0dd607723fa742e20c6708/includes/class.llms.user.permissions.php#L8
according to our [doc standards](https://github.com/gocodebox/lifterlms/blob/trunk/docs/documentation-standards.md#file-headers) it should be changed to:
` * @version [version]`

then I think we're ready to go :D 
Thanks!

eri-trabiccolo (approved):
@daniel-shuy 
looks good to me, thanks! <3

PR #1398, opened 2020-10-19
Add docs for exceptions via the 'throws' tag

### LINKED ISSUE #1397: Documentation standards for @throws tag
is this the right position for this tag?
I think we don't have a rule for that, but I would prefer it somewhere else, like
```
@since

@throw

@param
@return
```
looks more readable to me...

_Originally posted by @eri-trabiccolo in https://github.com/gocodebox/lifterlms/pull/1386#discussion_r506545444_

thomasplevy:
Proposed documentation standards updates, closes #1397 

(Intentionally opening this PR directly against trunk)


eri-trabiccolo (approved):
Fine for me.

PR #1402, opened 2020-10-20
Improve code quality of recent refactor

thomasplevy:
+ Minor refactor to reduce code quality issues reported by codeclimate on PR #1400
+ Fixed a bug not caught by initial developer or code reviewer

### DIFF
--- a/includes/abstracts/llms-abstract-generator-posts.php
+++ b/includes/abstracts/llms-abstract-generator-posts.php
@@ -678,7 +678,7 @@ abstract class LLMS_Abstract_Generator_Posts {
 
 		$image_url = ( is_array( $url_or_raw ) && ! empty( $url_or_raw['featured_image'] ) ) ? $url_or_raw['featured_image'] : $url_or_raw;
 
-		if ( is_string( $image_url ) ) {
+		if ( $image_url && is_string( $image_url ) ) {
 
 			$id = $this->sideload_image( $post_id, $image_url, 'id' );
 			if ( ! is_wp_error( $id ) ) {
--- a/includes/abstracts/llms-abstract-generator-posts.php
+++ b/includes/abstracts/llms-abstract-generator-posts.php
@@ -528,7 +528,10 @@ abstract class LLMS_Abstract_Generator_Posts {
 		// Importing blocks is disabled.
 		if ( ! $this->is_reusable_block_importing_enabled() ) {
 			return null;
-		} elseif ( empty( $raw['_extras']['blocks'] ) ) {
+		}
+
+		// No blocks to import.
+		if ( empty( $raw['_extras']['blocks'] ) ) {
 			return false;
 		}
 
@@ -775,7 +778,10 @@ abstract class LLMS_Abstract_Generator_Posts {
 		// Sideloading is disabled.
 		if ( ! $this->is_image_sideloading_enabled() ) {
 			return null;
-		} elseif ( empty( $raw['_extras']['images'] ) ) {
+		}
+
+		// No images to sideload.
+		if ( empty( $raw['_extras']['images'] ) ) {
 			return false;
 		}
 
--- a/includes/abstracts/llms-abstract-generator-posts.php
+++ b/includes/abstracts/llms-abstract-generator-posts.php
@@ -783,12 +783,12 @@ abstract class LLMS_Abstract_Generator_Posts {
 
 		$find     = array();
 		$replace  = array();
-		$curr_url = get_site_url();
+		$curr_url = parse_url( get_site_url(), PHP_URL_HOST );
 		$post_id  = $post->get( 'id' );
 		foreach ( $raw['_extras']['images'] as $src ) {
 
 			// Don't sideload images from this site.
-			if ( parse_url( $src, PHP_URL_HOST ) === parse_url( $src, PHP_URL_HOST ) ) {
+			if ( parse_url( $src, PHP_URL_HOST ) === $curr_url ) {
 				continue;
 			}
 
--- a/includes/abstracts/llms-abstract-generator-posts.php
+++ b/includes/abstracts/llms-abstract-generator-posts.php
@@ -108,7 +108,7 @@ abstract class LLMS_Abstract_Generator_Posts {
 	 * @since 3.16.11
 	 * @since 3.28.3 Add extra slashes around JSON strings.
 	 * @since 3.30.2 Skip JSON evaluation for non-string values; make publicly accessible.
-	 * @since [version] Moved from `LLMS_Generator`,
+	 * @since [version] Moved from `LLMS_Generator`.
 	 *
 	 * @param int   $post_id WP Post ID.
 	 * @param array $raw     Raw data.
@@ -140,6 +140,7 @@ abstract class LLMS_Abstract_Generator_Posts {
 	 * @param int    $post_id WP_Post ID.
 	 * @param string $key     Meta key.
 	 * @param mixed  $val     Meta value.
+	 * @return void
 	 */
 	protected function add_custom_value( $post_id, $key, $val ) {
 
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
Ohh, gave a rapid look was the if on `parse_url` the bug?
Anyways I spotted a couple of `elseif` after a return statement that can be turned into `if` only... Well going to add them in my review tomorrow.

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -108,26 +108,50 @@ public function __construct() {
...
+	 * Automatically slashes JSON data when supplied.
+	 *
+	 * Automatically unserializes serialized data so `add_post_meta()` can re-serialize.
+	 *
+	 * @since [version]
+	 *
+	 * @param int    $post_id WP_Post ID.
+	 * @param string $key     Meta key.
+	 * @param mixed  $val     Meta value.
+	 */
```
eri-trabiccolo, line 144:
```suggestion
	 * @return void
	 */
```

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -542,35 +528,79 @@ protected function handle_reusable_blocks( $post, $raw ) {
 		// Importing blocks is disabled.
 		if ( ! $this->is_reusable_block_importing_enabled() ) {
 			return null;
+		} elseif ( empty( $raw['_extras']['blocks'] ) ) {
```
eri-trabiccolo:
```suggestion
		} 
		if ( empty( $raw['_extras']['blocks'] ) ) {
```

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -542,35 +528,79 @@ protected function handle_reusable_blocks( $post, $raw ) {
...
+		$id = wp_insert_post(
+			array(
+				'post_content' => $block['content'],
+				'post_title'   => $block['title'],
+				'post_type'    => 'wp_block',
+				'post_status'  => 'publish',
+			)
+		);
+
+		if ( ! is_wp_error( $id ) ) {
```
eri-trabiccolo: This only makes sense if we insert the post asking for returning a wp error on failure. 
Meaning that the previous instruction must be:
```php
		$id = wp_insert_post(
			array(
				'post_content' => $block['content'],
				'post_title'   => $block['title'],
				'post_type'    => 'wp_block',
				'post_status'  => 'publish',
			),
			true
		);
```
otherwise, this must be just
```suggestion
		if ( $id ) {
```

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -649,7 +679,7 @@ protected function set_featured_image( $url_or_raw, $post_id ) {
 
 		$image_url = ( is_array( $url_or_raw ) && ! empty( $url_or_raw['featured_image'] ) ) ? $url_or_raw['featured_image'] : $url_or_raw;
 
-		if ( $image_url ) {
+		if ( is_string( $image_url ) ) {
```
eri-trabiccolo:
```suggestion
		if ( is_string( $image_url ) && $image_url ) {
```
right? or we're just assuming that variable will never be empty?

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -674,6 +704,29 @@ public function set_default_post_status( $status ) {
 		$this->default_post_status = $status;
 	}
 
+	/**
+	 * Set all metadata for a give post object
```
eri-trabiccolo:
```suggestion
	 * Set all metadata for a given post object
```

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -674,6 +704,29 @@ public function set_default_post_status( $status ) {
...
+	 *
+	 * This method will only set metdata for registered LLMS_Post_Model properties.
```
eri-trabiccolo:
```suggestion
	 * This method will only set metadata for registered LLMS_Post_Model properties.
```

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -674,6 +704,29 @@ public function set_default_post_status( $status ) {
...
+	 *
+	 * @since [version]
+	 *
+	 * @param LLMS_Post_Model $post An LLMS post object
+	 * @param array           $raw  Array of raw data
+	 *
+	 * @return void
```
eri-trabiccolo:
```suggestion
	 * @param LLMS_Post_Model $post An LLMS post object.
	 * @param array           $raw  Array of raw data.
	 * @return void
```

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -724,34 +777,33 @@ protected function sideload_images( $post, $raw ) {
 		// Sideloading is disabled.
 		if ( ! $this->is_image_sideloading_enabled() ) {
 			return null;
+		} elseif ( empty( $raw['_extras']['images'] ) ) {
```
eri-trabiccolo:
```suggestion
		} 
		if ( empty( $raw['_extras']['images'] ) ) {
```

includes/abstracts/llms-abstract-generator-posts.php
```diff
@@ -108,26 +108,50 @@ public function __construct() {
 	 * @since 3.16.11
 	 * @since 3.28.3 Add extra slashes around JSON strings.
 	 * @since 3.30.2 Skip JSON evaluation for non-string values; make publicly accessible.
-	 * @since [version] Moved from `LLMS_Generator`.
+	 * @since [version] Moved from `LLMS_Generator`,
```
eri-trabiccolo:
```suggestion
	 * @since [version] Moved from `LLMS_Generator`.
```

eri-trabiccolo (changes requested):
minor changes requested

includes/abstracts/llms-abstract-generator-posts.php
thomasplevy: What's the reasoning behind this change?

eri-trabiccolo: `else` after a return statement doesn't make sense.
It's not wrong, but doesn't make sense, right?

thomasplevy: `else` by itself doesn't make sense but an `elseif` I think does.

If the first condition isn't met, there won't be a return and then the second condition might be met.

includes/abstracts/llms-abstract-generator-posts.php
thomasplevy: mmm silly me. Thanks

includes/abstracts/llms-abstract-generator-posts.php
thomasplevy: I think it could actually be an empty string, which we wouldn't want to pass in so you're right, but I'm going to reverse the checks so it's `$image_url && is_string(...)`

thomasplevy:
@eri-trabiccolo one more check before I lay this whole refactor to rest (I hope), specifically here:

https://github.com/gocodebox/lifterlms/pull/1402/commits/c36c0a570deb0e36cb474107e75ac1f56d01f513

I've decided that we're *not* going to import course samples and templates from an API and instead are going to include the JSON files in the `sample-data` directory (the way we do currently) with the added change that we're going to include additionally include any images contained in any course templates.

This means that during an import of sample data we need to sideload images from that directory into the media library but that also means that we need a way to filter whether or not sideloading is allowed from the same site.

I'll be working on the process to include the imports and images and make it all work in a separate PR but this final change is "necessary" to fulfill those plans.

includes/abstracts/llms-abstract-generator-posts.php
eri-trabiccolo: the only relevant part of an `elseif` after a return statement is the `if` which I keep.
`elseif` is equivalent to `else { if (something) { do(); } }` so if you had something like:
```php
if ( a ) {
    return;
} else {
    if ( b ) {
        do_something_else();
   }
}
```
I'm sure you turned it into:
```php
if ( a ) {
    return;
} 
if ( b ) {
    do_something_else();
}
```
so that's what I'm proposing here :D

PR #1403, opened 2020-10-21
Improve the Importer success message

thomasplevy:
## Description

When doing final testing on generator improvements merged in from #1386 I noticed that as a result of adding new keys to the `LLMS_Generator` generated content property we now have a success message with duplicates since it lists the counts for *all* stats in the return.

I've made some changes and now, instead of listing our all of the generated content with numbers, the success message only contains two items:

1. Courses imported
2. Users imported

And instead of listing just a count of how many of each item were imported it creates a comma-separated list of anchors to the edit post/user admin page with the text of the anchor being the course title or user's display name.

This makes the success message, I hope, generally more usable instead of just relaying information.

I've never received any feedback about this but I would imagine that the first step after importing something is to have a look at it. This reduces the number of steps needed to perform that action.

We *lose* the counts but I don't think this really matters.

## How has this been tested?

+ Added New tests
+ Existing tests pass
+ Manually

## Screenshots 

Screenshot_2020-10-21_16-03-28

New message ^

## Types of changes

+ Updates

## Checklist:

### DIFF
--- a/includes/admin/class.llms.admin.import.php
+++ b/includes/admin/class.llms.admin.import.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Admin/Classes
  *
  * @since 3.3.0
- * @version 3.37.3
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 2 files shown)


thomasplevy:
@eri-trabiccolo I'd like to say this is the last one for this bit of refactoring but I have another one prepped too...

Thanks

thomasplevy:
@chrisbadgett can you have a look at the description and the screenshot and give me a yay or nay too please!

includes/admin/class.llms.admin.import.php
```diff
@@ -36,55 +36,60 @@ public function __construct() {
...
-	 * Localize statistic information for display on success.
+	 * Convert an array of generated content IDs to a list of anchor tags to edit the generated content
 	 *
-	 * @since 3.35.0
+	 * @since [version]
 	 *
-	 * @param string $stat Statistic key name.
-	 * @return string
+	 * @param int[]  $ids  Array of object IDs. Either WP_Post IDs or WP_User IDs.
+	 * @param string $type Object types. Either "post" or "user".
```
eri-trabiccolo:
```suggestion
	 * @param string $type Object's type. Either "post" or "user".
```

includes/admin/class.llms.admin.import.php
```diff
@@ -36,55 +36,60 @@ public function __construct() {
...
 
-			case 'terms':
-				$name = __( 'Terms', 'lifterlms' );
-				break;
+		$generated = $generator->get_generated_content();
 
-			default:
-				$name = $stat;
+		if ( ! empty( $generated['course'] ) ) {
+			$msg .= '<li>' . sprintf( __( 'Imported courses: %s', 'lifterlms' ), $this->get_generated_content_list( $generated['course'], 'post' ) ) . '</li>';
```
eri-trabiccolo, line 85:
```suggestion
			// Translators: %s = comma-separated list of anchors to the imported courses.
			$msg .= '<li>' . sprintf( __( 'Imported courses: %s', 'lifterlms' ), $this->get_generated_content_list( $generated['course'], 'post' ) ) . '</li>';
```

includes/admin/class.llms.admin.import.php
```diff
@@ -36,55 +36,60 @@ public function __construct() {
...
+		}
+		if ( ! empty( $generated['user'] ) ) {
+			$msg .= '<li>' . sprintf( __( 'Imported users: %s', 'lifterlms' ), $this->get_generated_content_list( $generated['user'], 'user' ) ) . '</li>';
```
eri-trabiccolo, line 89:
```suggestion
			// Translators: %s = comma-separated list of anchors to the imported users.
			$msg .= '<li>' . sprintf( __( 'Imported users: %s', 'lifterlms' ), $this->get_generated_content_list( $generated['user'], 'user' ) ) . '</li>';
```

tests/phpunit/unit-tests/admin/class-llms-test-admin-import.php
```diff
@@ -9,6 +9,7 @@
  *
  * @since 3.35.0
  * @since 3.37.8 Update path to assets directory.
+ * @since [version] Test success message genration.
```
eri-trabiccolo:
```suggestion
 * @since [version] Test success message generation.
```

eri-trabiccolo (changes requested):
I'm not against this change but... mmm
don't you think that the number of lessons/question/quizzes - especially when importing one course at a time - gave the user a nice feeling that everything went through? dunno - just a question.

thomasplevy:
> I think I initially built it with the numbers so that *I could feel confident* that it worked but, lets say you've never used it before and you don't see all those numbers and you just see my screenshot, would you imagine feeling confident?

I think I would be satisfied yeah

> I am sure that some experienced user will complain about the change but I'm not sure there's a really valid complaint there.

I think you're right.

Issue #1278, opened 2020-08-06
Persistent Payment Gateway Upsell Alert Issue


chrisbadgett:
Users are indicating persistent payment gateway alert notices on every admin page load and would like the option to close alert forever.

Proposed solution: 

Add grey link under or beside or below "Remind me later" button that reads: "Remind me never"

PR #1405, opened 2020-10-23
Add an option to dismiss admin notifications forever

### LINKED ISSUE #1360: Admin Notices should be able to be dismissed forever
Currently, the `LLMS_Admin_Notices::add_notice()` will redisplay any dismissed notice 7 days after it was dismissed. This default value of 7 days can be overridden with a custom value.

When we wish for something to be permanently dismissed, we add a large value (730 days or 2 years) to make it seem like it's dismissed forever, but this isn't actually what happens.

See related issue #1278 

Ideally the `add_notice()` method should have an option to "dismiss_forever". When a notice is dismissed forever we should record that it has been permantently dismissed the `add_notice()` method should prevent the addition of the notice.

imknight:
## Description
added dismiss_forever option for notice to dismiss and remove

Fixes #1360 

## How has this been tested?
set a notice which will dismiss forever, once hide the notice, check the option has never added back to the option table.

## Types of changes
## Checklist:
- [X ] My code has been tested.
- [ X] My code passes all existing automated tests
- [X ] My code follows the LifterLMS Coding & Documentation Standards.

### DIFF
--- a/includes/admin/class.llms.admin.notices.php
+++ b/includes/admin/class.llms.admin.notices.php
@@ -73,6 +73,7 @@ class LLMS_Admin_Notices {
 	 * @return   void
 	 * @since    3.0.0
 	 * @version  3.3.0 - added "flash" option
+	 * @version  [version] - added "dismiss_forever" option
 	 */
 	public static function add_notice( $notice_id, $html_or_options = '', $options = array() ) {
 
@@ -95,6 +96,7 @@ class LLMS_Admin_Notices {
 			array(
 				'dismissible'      => true,
 				'dismiss_for_days' => 7,
+				'dismiss_forever'  => false, // If true, will delete the notice after hide it.
 				'flash'            => false, // If true, will delete the notice after displaying it.
 				'html'             => '',
 				'remind_in_days'   => 7,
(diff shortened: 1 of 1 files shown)


thomasplevy (commented):
@imknight I haven't pulled down to test this but I'm pretty sure that this will result in the notice returning again on the next page load in favor of actually dismissing it forever.

Can you confirm that you've tested this with existing notifications and that it works as expected? By that I mean that the "forever" notice never actually returns.

It might be helpful to look at the PR related to #1278 (and the notice in this issue) to see if adding this new option to that value actually has the desired effect.

imknight:
I thought that simply prevent the options to add back will stop the notification appear again.
what's a good way to for me to call out that payment gateway notice?

thomasplevy:
@imknight 

I was on vacation, thanks for your patience!

1) Disable, deactivate, uninstall any and all LifterLMS payment gateways (deactivating might be the only necessary step here, I don't quite remember without digging into it and reading the code)
2) Head into the db and from the `wp_options` table, delete the option with the `option_name` of `llms_notices` (it might actually be `lifterlms_notices` -- We used to not enforce a convention on how things were prefixed and I can't remember off the top of my head which is which.


Best,

thomasplevy:
@imknight closing this since it doesn't currently work. If you wish to continue working on this please reopen or submit an updated PR when you're ready.

Thanks <3

imknight:
ya sorry for that, been really busy with work recently, will relook into this after I clear my work

PR #1408, opened 2020-10-29
Stop scrubbing the post name because it may has url encoded characters

alaa-alshamy:
## Description

This PR, fix a bug in using the post name when it has a url encoded characters for example when it has an arabic characters
And the current scrub uses the "sanitize_text_field" function which is remove the url encoded characters, in this block:
```
#/wp-includes/formatting.php 5401

while ( preg_match( '/%[a-f0-9]{2}/i', $filtered, $match ) ) {
	$filtered = str_replace( $match[0], '', $filtered );
	$found    = true;
}
```

Fixes #

## Steps to reproduce

- Create a course with an arabic characters in it's slug
- Enroll in it
- Open the "My Grades" page
- Click on the link of the course

You will see a blank page because the url is wrong

## How has this been tested?

- Manually

## Screenshots 
Screenshot from 2020-10-29 19-32-54
Screenshot from 2020-10-29 19-30-39

## Types of changes

- Bug fix

## Checklist:

### DIFF
--- a/includes/abstracts/abstract.llms.post.model.php
+++ b/includes/abstracts/abstract.llms.post.model.php
@@ -446,7 +446,7 @@ abstract class LLMS_Post_Model implements JsonSerializable {
 		// if we found a valid, apply default llms get get filter and return the value.
 		if ( isset( $val ) ) {
 
-			if ( ! $raw && 'content' !== $key ) {
+			if ( ! $raw && 'content' !== $key && 'name' !== $key ) {
 				$val = $this->scrub( $key, $val );
 			}
 
--- a/includes/abstracts/abstract.llms.post.model.php
+++ b/includes/abstracts/abstract.llms.post.model.php
@@ -446,7 +446,19 @@ abstract class LLMS_Post_Model implements JsonSerializable {
 		// if we found a valid, apply default llms get get filter and return the value.
 		if ( isset( $val ) ) {
 
-			if ( ! $raw && 'content' !== $key && 'name' !== $key ) {
+			/**
+			 * Filters the list of properties which should be excluded from scrubbing during a property read.
+			 *
+			 * The dynamic portion of this hook, `{$this->model_post_type}`, refers to the post's model type,
+			 * for example "course" for an `LLMS_Course`, "membership" for an `LLMS_Membership`, etc...
+			 *
+			 * @since [version]
+			 *
+			 * @param string[]        $props An array of property keys to be excluded from scrubbing.
+			 * @param LLMS_Post_Model $this  Instance of the post object.
+			 */
+			$exclude = apply_filters( "llms_get_{$this->model_post_type}_no_scrub_props", array( 'content', 'name' ), $this );
+			if ( ! $raw && ! in_array( $key, $exclude, true ) ) {
 				$val = $this->scrub( $key, $val );
 			}
 


eri-trabiccolo:
Thanks for reporting this, that's interesting @alaa-alshamy 
Although I think this needs some other thoughts...

I think this can be a good occasion to think about whether or not we need to scrub the "post properties" prior to insert them into the db (`wp_insert_post` should already do that for us).
You're probably right and the post_name (aka the slug) should never be scrubbed, when getting it (as you propose) but also when setting it.

While we think about this, if you need a rapid fix, I suggest you to override this file in your child-theme (or custom plugin):
https://github.com/gocodebox/lifterlms/blob/6feb59190fc9c8873524caadeb14ab019cc87ecf/templates/myaccount/my-grades.php

And change this line:
https://github.com/gocodebox/lifterlms/blob/6feb59190fc9c8873524caadeb14ab019cc87ecf/templates/myaccount/my-grades.php#L28
this way:
```php
<td><a href="<?php echo llms_get_endpoint_url( 'my-grades', $course->get( 'name', true ) ); ?>"><?php echo $course->get( 'title' ); ?></a></td>
```

That allows you to get the property skipping the scrubbing :D

alaa-alshamy:
@eri-trabiccolo Thanks for your suggestion, yes i noticed that but i preferred to make my change to fix it generally because i think there's another places have this issue

but do you think my change will has some side effects ?

eri-trabiccolo:
Hi @alaa-alshamy 
at a first glance no, I don't think your change will have negative side effects, but we need tests, and as said above, maybe take this occasion for a more in depth change that involves other "post" properties, and both the `get` and `set` methods.
You PR is good to me but only deals with the specific case of the post name (slug) get, this means that using the `LLMS_Post_Model` api to set a post slug will still improperly scrub it.

So, as said above, while we think a little bit more about this you might want to fix your very issue in a way - overriding templates - that will ensure you don't lose your changes at the next plugin update.

:D

alaa-alshamy:
@eri-trabiccolo Mmmm I got your point, i hope you finish thinking about this issue soon :) Thanks

thomasplevy:
Thanks for raising this concern @alaa-alshamy 

@eri-trabiccolo While you raise another good point about whether or not we should be scrubbing during inserts (my gut tells me probably not, but I need to dig into underlying core code and existing functionality and look at tests and so on before committing) I don't think we need to expand this particular PR to include a discussion about that.

I think it would be best to open a new issue about your concerns here @eri-trabiccolo and we can move forward with this PR fixing the bug that @alaa-alshamy has raised.

HOWEVER, since we now have two properties in question `post_content` and `post_name` I think it best to change the strategy with which we're excluding properties from scrubbing.

I think it would be better to use a list (with a filter). See my review suggestion @alaa-alshamy

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -446,7 +446,7 @@ private function ___get( $key, $raw = false ) {
 		// if we found a valid, apply default llms get get filter and return the value.
 		if ( isset( $val ) ) {
 
-			if ( ! $raw && 'content' !== $key ) {
+			if ( ! $raw && 'content' !== $key && 'name' !== $key ) {
```
thomasplevy: In order to future proof this on the assumption that at some point in the future we decide or need to add additional properties to this list, this will keep the logical expression a little simpler (instead of just adding an additional condition) and will allow 3rd parties to extend this list as needed without having to modify core code.

```suggestion
			/**
			 * Filters the list of properties which should be excluded from scrubbing during a property read.
			 *
			 * The dynamic portion of this hook, `{$this->model_post_type}`, refers to the post's model type,
			 * for example "course" for an `LLMS_Course`, "membership" for an `LLMS_Membership`, etc...
			 *
			 * @since [version]
			 *
			 * @param string[]        $props An array of property keys to be excluded from scrubbing.
			 * @param LLMS_Post_Model $this  Instance of the post object.
			 */
			$exclude = apply_filters( "llms_get_{$this->model_post_type}_no_scrub_props", array( 'content', 'name' ), $this );
			if ( ! $raw && ! in_array( $key, $exclude, true ) ) {
```

thomasplevy (changes requested):

alaa-alshamy (commented):
Improve fixing the post slug bug

alaa-alshamy:
@thomasplevy Big thanks for your improvement, i like it especially the part of changing the excluded props from outside :+1:

thomasplevy:
Basic tests to written to cover the scenario identified by the bug report in this pr at https://github.com/gocodebox/lifterlms/pull/1444/commits/d71fa7322917081cdd4c162a520004b3120d28d3

PR #1416, opened 2020-11-03
PHP 8 Compat & Testing

### LINKED ISSUE #1414: PHP 8.0 Testing and Compatibility
PHP 8.0 set for release on 11/26/2020 - https://wiki.php.net/todo/php80

+ [x] Reading and research
+ [x] Testing - manual -- Initial testing shows that LifterLMS is functional on PHP 8, haven't spotted any issues so far. 
+ [x] Testing automated (phpunit & e2e tests)
+ [ ] ~Add-on testing~ Moved to add-on repos
+ [x] Add to travis test matrix

Yoast compat report: https://developer.yoast.com/blog/the-2020-wordpress-and-php-8-compatibility-report/

Current errors as reported by phpunit integration test suite: [output.txt](https://github.com/gocodebox/lifterlms/files/5537907/output.txt)

thomasplevy:
+ Fixes php8 compatibility issues identified by phpunit integration tests
+ Improves travis configs to allow simpler testing against php8 by add-ons
+ Adds an automatic travis config for add-ons

Closes #1414


thomasplevy:
@eri-trabiccolo there's quite a lot of changes here mostly related to build steps, mostly I need your review on the code changes but if you could have a glance through everything and let me know if you spot any issues I'd appreciate it.

The changes here won't have any negative impact in php7 environments so the sooner we can get this out the better, I think.

composer.json
```diff
@@ -89,21 +90,25 @@
...
+      "@tests-remove",
+      "@post-update-install-cmd"
     ],
-    "tests-run": "\"vendor/bin/phpunit\"",
-    "phpcs": [
-      "\n\nWARNING: This command is deprecated, use `composer run check-cs` instead!\n\n",
-      "composer run check-cs"
-    ]
+    "post-update-install-cmd": [
+      "@tests-install",
```
eri-trabiccolo, line 99: wait... I just went with a `composer install` after pulling this and got an error because the db was already installed.
I don't event think we should install tests on composer install to be honest, but if we want to do it then let's go with `@tests-reinstall`...
(so the same in the php8/composer.json above)

phpcs.xml
```diff
@@ -4,6 +4,13 @@
 
 	<file>.</file>
 
+	<!-- Exclude project directories -->
```
eri-trabiccolo:
```suggestion
    <!-- Exclude project directories -->
```

eri-trabiccolo (changes requested):
I also had issues running tests with php8, but I think due to wp, probably I didn't get what I have to do :D

composer.json
thomasplevy: Okay... this gets hairy...

In order to include the "patched" versions of files included in the `autoload-dev` in the composer config file we need to have the wp tests library installed.

If we don't install the test library then the `config-cs` command fails because the autoloader is attempting to load files that don't exist (yet)

eri-trabiccolo: Ok that's fine, I see, so either we go with the reinstall or we add a check in our tests-install to skip the db creation if it already exists, but this is not critical I guess. <3

thomasplevy: I'll update the test install to skip if the db already exists, that seems better I think, right?

thomasplevy: Okay, I've made a bunch of changes based on your feedback:

Relevant updates from lifterlms-tests: https://github.com/gocodebox/lifterlms-tests/commit/0dc32075fae66ade4ea5d576ef39256a22f1facd

See recent commits for the rest

1. I added new composer scripts to make installing on php8 simpler
2. I've removed the tests installation from the post-install command
3. I've removed the necessity of having the class map in the autoload dev of the composer.json file

Now, during tests installation, if there are patched items in the wp tests library directory, they'll be copied directly into the phpunit directory, backups of the original versions will be made and automatically restored during test removal

This will simplify the steps and make switching PHP versions pretty simple, I hope.

eri-trabiccolo: Works fine :D <3

eri-trabiccolo (approved):

PR #1422, opened 2020-11-05
Enhancements to Setup Wizard & Importer

### LINKED ISSUE #1305: Upgrade LifterLMS Sample Course Import Process
+ [x] Adjust current last step to setup wizard
+ [x] Add new Finish step to setup wizard
+ [x] ~Upgrade import file process to bring course and lesson images in addition to featured images to destination instead of just referencing~ See #1196 
+ [x] Get Import files and add 3 new courses to course import process
+ [x] Strip author information from files
+ [x] ~Replace file in core plugin~ See #1276  

### Adjust current last step

‹_A_LifterLMS_Sandbox_—_WordPress-2

### Prepare Courses
All 3 courses for import are here: https://free-course-template.myliftersite.com/courses/

### Wizard New Screen Concept

IMG_1336

### Copy for Wizard

Install Sample Courses!

Accelerate your progress by installing a quick LifterLMS training course and useful course templates.

~

LifterLMS Quickstart Course

Learn how to use the most important parts of LifterLMS in less than 30 minutes

~

Free Course Lead Magnet Template

A proven mini course template to help grow your email list and increase enrollments 

~

Course Template

A proven course sales page template and lesson structure for high value online courses

thomasplevy:
## Description

Closes #1305 

+ Replaces the final step of the setup wizard with an option to import 3 separate courses from the export api at academy.lifterlms.com
+ Updates the LifterLMS -> Import screen to:
  A. Allow "cloud" import from the above API
  B. Moves the "Upload" functionality to a secondary action, similar to the WP Core plugins screen

## How has this been tested?

+ Manually
+ New & existing phpunit/integration tests
+ ~**E2E tests are going to fail as a result of setup wizard changes. These need to be reworked and rewritten. I'm submitting this now to start code review while I work on the e2e tests**~

## Types of changes

+ Updates and improvements
+ Minor deprecations in the public API

## Checklist:


includes/admin/class-llms-export-api.php
```diff
@@ -0,0 +1,84 @@
+<?php
+/**
+ * Manage imports from lifterlms.com export API
+ *
+ * File description.
+ *
```
eri-trabiccolo:
```suggestion
```

includes/admin/class-llms-export-api.php
```diff
@@ -0,0 +1,84 @@
...
+ * @package LifterLMS/Classes
```
eri-trabiccolo:
```suggestion
 * @package LifterLMS/Admin/Classes
```

includes/admin/class-llms-export-api.php
```diff
@@ -0,0 +1,84 @@
...
+
+	}
+
+
+	/**
+	 * Retrieve a list of available exports
+	 *
+	 * @since [version]
+	 *
+	 * @param int $page     Results page
```
eri-trabiccolo:
```suggestion
	 * @param int $page     Results page.
```

includes/admin/class.llms.admin.setup.wizard.php
```diff
@@ -123,90 +146,114 @@ public function generator_course_status( $status ) {
...
+		$i    = array_search( $step, $keys, true );
+
+		// Next step doesn't exist or the next step would be greater than the index of the last step.
+		if ( false === $i || $i++ >= count( $keys ) - 1 ) {
 			return false;
-		} else {
-			return $keys[ $i++ ];
 		}
+
+		return $keys[ $i++ ];
```
eri-trabiccolo: golf
```suggestion
		return $keys[ $i ];
```
because: http://sandbox.onlinephpfunctions.com/code/3ddf7183a2773398e4d7ea2661fd60c00d011d4d

includes/admin/class.llms.admin.setup.wizard.php
```diff
@@ -260,130 +320,206 @@ public function output() {
...
+			return null;
 		}
 
-		if ( ! current_user_can( 'manage_lifterlms' ) ) {
-			return;
+		$res = new WP_Error( 'llms-setup-save-invalid', __( 'There was an error saving your data, please try again.', 'lifterlms' ) );
+
+		$step = llms_filter_input( INPUT_POST, 'llms_setup_save', FILTER_SANITIZE_STRING );
+		if ( method_exists( $this, 'save_' . $step ) ) {
+			$res = call_user_func( array( $this, 'save_' . $step ) );
```
eri-trabiccolo, line 407: I continue on my golf (and perf) theory :P 
```suggestion
			$res = $this->{"save_$step"}();
```
look at that beauty!

includes/admin/views/importable-courses.php
```diff
@@ -0,0 +1,39 @@
...
+	 *
+	 * @since [version]
+	 *
+	 * @param array[] $courses List of importable course data.
+	 */
+	do_action( 'llms_before_importable_courses', $courses );
+
+foreach ( $courses as $course ) {
+	include LLMS_PLUGIN_DIR . 'includes/admin/views/importable-course.php';
+}
```
eri-trabiccolo:
```suggestion
	foreach ( $courses as $course ) {
		include LLMS_PLUGIN_DIR . 'includes/admin/views/importable-course.php';
	}
```

eri-trabiccolo (changes requested):
Minor stuff. Need to test it to better debug it.

includes/admin/class.llms.admin.setup.wizard.php
thomasplevy: FASCINATING. I didn't know this.

However, I think now that I understand it the code feels more readable as:

```php

		if ( false === $i || $i + 1 >= count( $keys ) ) {
			return false;
		}

		return $keys[ ++$i ];

```

Would you agree?

includes/admin/class.llms.admin.setup.wizard.php
thomasplevy: I still haven't seen any metrics on your performance theory though.... wink

includes/admin/class.llms.admin.setup.wizard.php
eri-trabiccolo: I agree <3

includes/admin/class.llms.admin.setup.wizard.php
eri-trabiccolo: a metric that certifies that function calls are expensive in php in general
or that despite we know they're expensive in php in general `call_user_func( array( $this, 'save_' . $step ) );` is still less expensive than `$this->{"save_$step"}();`? :D

eri-trabiccolo: Still no comment on the beauty though...

eri-trabiccolo:
I ran the wizard and imported the 3 course, it went fine. The import itself was a little bit slow though - probably because of the images?!
Have to say that the UI of the template selection is not 100% clear, meaning that the checkmark is not 100% visible to me.
Might look stupid but maybe a message that says "You're about to import 3 courses", or "You selected 3 courses to be imported", might help.
I'm also not completely sure the 3 of them must be selected by default... 

Finally, as I think we already discussed, in the future we might think about doing the import with a bg process, plus a bar showing the progress. as of now we could even just start a spinner, I think it gives more the feeling that something is happening...

What do you think?

thomasplevy:
Fixed e2e test util packages so e2e tests are now passing

And in doing so added e2e coverage for new UI elements. I did not write e2e tests to use the cloud importer on the import screen but I don't think it's necessary because that area is well covered in php

thomasplevy:
> I ran the wizard and imported the 3 course, it went fine. The import itself was a little bit slow though - probably because of the images?!

Yes, that's why it's slow. I don't know if there's really any way to speed this up unless we decide to only include templates that don't have images.

> Have to say that the UI of the template selection is not 100% clear, meaning that the checkmark is not 100% visible to me.

What about something like this instead?

Screenshot_2020-11-06_14-27-53


> Might look stupid but maybe a message that says "You're about to import 3 courses", or "You selected 3 courses to be imported", might help.

I like this idea. WC does this in their wizard when they install plugins during one of the steps. I'll work on this.

> I'm also not completely sure the 3 of them must be selected by default...

@chrisbadgett can you weigh in on this and tie break? I'm not sure we should have all 3 checked by default also but I'm not terribly sure we shouldn't. It's *possible* that we could improve the UI overall by not having *anything* checked by default. This would force users to notice

> Finally, as I think we already discussed, in the future we might think about doing the import with a bg process, plus a bar showing the progress. as of now we could even just start a spinner, I think it gives more the feeling that something is happening...
> 
> What do you think?

Yes, for sure, I think moving to a background process with a loading bar makes sense but this does already include a spinner:

Peek 2020-11-06 14-02

thomasplevy:
> Have to say that the UI of the template selection is not 100% clear, meaning that the checkmark is not 100% visible to me.
> Might look stupid but maybe a message that says "You're about to import 3 courses", or "You selected 3 courses to be imported", might help.

Here we go, this feels a lot better, thoughts @eri-trabiccolo & @chrisbadgett 

Peek 2020-11-06 15-06

chrisbadgett:
The latest list of course imports with the toggle on by default looks clear to me. Very clean and more readable than the 3 column. 

I believe the message at the bottom "the 3 selected courses will be downloaded and imported into this site from LifterLMS.com"  is no longer true since the courses are imported from inside the plugin itself?

thomasplevy:
@chrisbadgett I worked hard to include everything in the plugin but with images the size of the plugin tripled as a result. From 3.5mb to 9.6mb.

I felt that was an unacceptable increase in size, so I rolled back my decision and have stuck with importing via an API.

There's problems with doing that. There's problems with not importing from an API too.

eri-trabiccolo:
@thomasplevy 
I like the new list + gutenchecks the better as well.

I like the new message, but I have to be frank:
1) it's below and on the opposite side (left) of the import button (right) which - being blue - attracts my attention
2) in smaller and cursive font
3) in an area with a grey background while all the "action" happens in an area with white background

These 3 things will, 99%, make it drop off my radar, I wouldn't even notice it. But this is how my brain works.
So I would prefer it above the buttons, right aligned, so that it sits (at least in desktops) above the primary button. IMHO

thomasplevy:
@eri-trabiccolo 

I only disagree with the usage of a smaller size and italics. When moved above the button the layout shifts and I think it's very noticeable despite a smaller size. I think using a smaller font (with italics) helps retain the desired emphasis but still allows it to be ignored since it is a small (but important) note.

Hows this look:

Peek 2020-11-09 08-53

eri-trabiccolo:
> @eri-trabiccolo
> 
> I only disagree with the usage of a smaller size and italics. When moved above the button the layout shifts and I think it's very noticeable despite a smaller size. I think using a smaller font (with italics) helps retain the desired emphasis but still allows it to be ignored since it is a small (but important) note.
> 
> Hows this look:
> 

Oh yeah I like it, and I already agreed with you, but I expressed myself badly.
What I wanted to say is that the combination of those 3 bullets made that message drop off my radar, but the only actual change I would have done was what you did, because:
> So I would prefer it above the buttons, right aligned, so that it sits (at least in desktops) above the primary button. IMHO

So above the buttons, right aligned, perfect :D

eri-trabiccolo (approved):

thomasplevy:
I broke the e2e tests with my last commit, need to fix that and then I'll merge

PR #1428, opened 2020-11-10
Add post model extra data fields only when exporting

### LINKED ISSUE #1427: Avoid adding 'extras' to the post toArray when parsing  data to be used in the course builder.
### Reproduction Steps

see #1426 

### Expected Behavior

+ I'd expect 'extras' to not be added to the array model

### Actual Behavior

+ 'extras' are added

I wonder if by default we should just set this filter to `false`, and only enable it on export (we don't need it on clonation at the moment, right?):
https://github.com/gocodebox/lifterlms/blob/4.7.1/includes/abstracts/abstract.llms.post.model.php#L1194

### Error Messages / Logs

n/a

LifterLMS 4.7.1

+ [x] Locally
+ [x] On a staging site
+ [x] On a production website

n/a

eri-trabiccolo:
## Description
Fixes #1427 

I've just added a new filter hook that we can leverage to allow adding extra data fields when exporting (false by default).
I thought a little bit at the _better_ way to solve this here, I think this is the one that gives us less headaches, and, since the new filter doesn't depend on the model post type doesn't force us to add new filters in the add-ons to "unlock" them.

## How has this been tested?
Manual export

## Types of changes
Enhancement.

## Checklist:

### DIFF
--- a/includes/abstracts/abstract.llms.post.model.php
+++ b/includes/abstracts/abstract.llms.post.model.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Abstracts/Classes
  *
  * @since 3.0.0
- * @version 4.7.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -53,7 +53,7 @@ abstract class LLMS_Post_Model implements JsonSerializable {
 	 * ie: "llms_order" for the "llms_order" post type
 	 *
 	 * @var string
-	 * @since  3.0.0
+	 * @since 3.0.0
 	 */
 	protected $db_post_type;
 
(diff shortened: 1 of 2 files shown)


thomasplevy (changes requested):
Thanks for taking the time to fix all the docs in this massive file.

I like the strategy but I think we should add the filters to force "extras" in the `LLMS_Post_Model::export()` method in favor of doing it in the controller of the post table.

This will ensure that even if the `export()` method is called from some unexpected context (you know, by a 3rd party, for example) the export file will have all the expected content.

Additionally, I believe a note should be included in the docs of the `llms_post_model_to_array_add_extras` filter noting that `__return_true` is used to force the filter on during exports to hopefully save you and I and other poor souls the time of digging for potential problems in the inevitable future.

Edit: Open to an argument why my suggested location for the filter is incorrect.

PR #1429, opened 2020-11-10
Remove reliace on mb_string functions

### LINKED ISSUE #1118: Fatal when downloading a certificate if mbstring PHP module not installed/enabled
The mbstring PHP module is optional, so I think we need a fall back. 

### Reproduction Steps

+ On a server where the mbstring PHP module is not installed/enabled download/save a certificate
+ Check the error log

### Expected Behavior

+ I expect to download the certificate without any issues

### Actual Behavior

+ Cannot download the certificate

### Error Messages / Logs

```
Fatal error: Uncaught Error: Call to undefined function mb_convert_encoding() in /home/***/public_html/wp-content/plugins/lifterlms/includes/class.llms.certificates.php:228 Stack trace: #0 /home/***/public_html/wp-content/plugins/lifterlms/includes/class.llms.certificates.php(188): LLMS_Certificates->modify_dom(‘<!DOCTYPE html>…’) #1 /home/***/public_html/wp-content/plugins/lifterlms/includes/class.llms.certificates.php(105): LLMS_Certificates->get_export_html(‘2064’) #2 /home/***/public_html/wp-content/plugins/lifterlms/includes/class.llms.certificates.php(161): LLMS_Certificates->generate_export(‘/home/***/…’, ‘2064’) #3 /home/***/public_html/wp-content/plugins/lifterlms/includes/controllers/class.llms.controller.certificates.php(149): LLMS_Certificates->get_export(‘2064’) #4 /home/***/public_html/wp-content/plugins/lifterlms/includes/controllers/class.llms.controller.certificates.php(111): LLMS_Controller_Certificates->download(‘2064’) #5 /home/***/public_html/wp-includes/cla in /home/***/public_html/wp-content/plugins/lifterlms/includes/class.llms.certificates.php on line 228
```

### System Information
(redacted)
<details>
  
```
WordPress

Home Url: https://---.com.au
Site Url: https://---.com.au
Login Url: https://---.com.au/wp-login.php
Version: 5.4
Debug Mode: No
Debug Log: No
Debug Display: Yes
Locale: en_US
Multisite: No
Page For Posts: Not Set
Page On Front: Home (#12) [https://---.com.au/]
Show On Front: page
Wp Cron: Yes
Server

Mysql Version: 5.7.29
Php Curl: Yes
Php Default Timezone: UTC
Php Fsockopen: Yes
Php Max Input Vars: 1000
Php Max Upload Size: 8 MB
Php Memory Limit: 256M
Php Post Max Size: 8M
Php Soap: No
Php Suhosin: No
Php Time Limt: 30
Php Version: 7.2.27
Software: Apache
Wp Memory Limit: 40M
Theme

Name: Astra
Version: 2.4.2
Themeuri: https://wpastra.com/
Authoruri: https://wpastra.com/about/
Template: 
Child Theme: No
Llms Support: Yes

Plugins

Astra Hooks: 1.0.2
Astra Pro: 2.4.0
Classic Editor: 1.5
Contact Form 7: 5.1.7
Elementor: 2.9.7
Elementor - Header, Footer & Blocks: 1.4.1
LifterLMS: 3.37.18
Really Simple SSL: 3.3.1
Starter Templates: 2.2.1
Ultimate Addons for Gutenberg: 1.14.8
Under Construction: 3.75
WPForms Lite: 1.5.9.5
Yoast SEO: 13.5

Integrations

BbPress: No
BuddyPress: No

Template Overrides
```

</details>

n/a

### Related User Information

https://wordpress.org/support/topic/certificate-download-error/

---

I couldn't find an equivalent "non-mb" for `mb_convert_encoding()`, but from several sources, e.g. https://stackoverflow.com/a/11978382
looks like 
`htmlspecialchars_decode(utf8_decode(htmlentities($htmls, ENT_COMPAT, 'UTF-8', false)));`
is a good alternative (fall back in our case) for:
`mb_convert_encoding( $html, 'HTML-ENTITIES', 'UTF-8' );` that we use here:
https://github.com/gocodebox/lifterlms/blob/master/includes/class.llms.certificates.php#L228

### LINKED ISSUE #1426: Possible fatal when loading the course builder - when no PHP mbstring module support
### Reproduction Steps

+ make sure your php doesn't have the mbstring module enabled
+ create a course with some text in its content
+ open it in the course builder

### Expected Behavior

+ I'd expect the builder working fine

(moreover we should not even parse the extra when loading a post in the builder..., I'll open a new issue about this: see https://github.com/gocodebox/lifterlms/issues/1427)

### Actual Behavior

+ The builder shows a blank page

### Error Messages / Logs

```
PHP Fatal error:  Uncaught Error: Call to undefined function mb_convert_encoding() in /home/{removed}/public_html/{removed}/wp-content/plugins/lifterlms/includes/llms.functions.core.php:362
```

```
LifterLMS 4.7.1

```

+ [ ] Locally
+ [ ] On a staging site
+ [x] On a production website

n/a

### Related User Information

HS-140566

see also #1118 
we should probably go with the polyfill.

thomasplevy:
## Description

+ Fixes #1426
+ Fixes #1118 

## How has this been tested?

+ Manually
+ New tests

## Types of changes

+ Bug fix
+ Drops dependency on php `mb_string` module

## Checklist:

### DIFF
--- a/includes/llms.functions.core.php
+++ b/includes/llms.functions.core.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Functions
  *
  * @since 1.0.0
- * @version 4.7.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -344,6 +344,7 @@ function llms_get_date_diff( $time1, $time2, $precision = 2 ) {
  * loading a partial string or an HTML string with errors.
  *
  * @since 4.7.0
+ * @since [version] Remove reliance on `mb_convert_encoding()`.
  *
  * @param string $string An HTML string, either a full HTML document or a partial string.
  * @return DOMDocument|WP_Error Returns an instance of DOMDocument with `$string` loaded into it
@@ -359,7 +360,7 @@ function llms_get_dom_document( $string ) {
 	$libxml_state = libxml_use_internal_errors( true );
 
 	$dom = new DOMDocument();
-	if ( ! $dom->loadHTML( mb_convert_encoding( $string, 'HTML-ENTITIES', 'UTF-8' ) ) ) {
+	if ( ! $dom->loadHTML( '<meta http-equiv="Content-Type" content="text/html; charset=utf-8">' . $string ) ) {
 		$dom = new WP_Error( 'llms-dom-document-error', __( 'DOMDocument XML Error encountered.', 'lifterlms' ), libxml_get_errors() );
 	}
 
--- a/includes/llms.functions.core.php
+++ b/includes/llms.functions.core.php
@@ -370,7 +370,7 @@ function llms_get_dom_document( $string ) {
 	// Remove the fixer meta element, if it's not removed it creates invalid HTML5 Markup.
 	$meta = $dom->getElementById( 'llms-get-dom-doc-utf-fixer' );
 	if ( $dom ) {
-		$meta->parentNode->removeChild( $meta );
+		$meta->parentNode->removeChild( $meta ); // phpcs:ignore: WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
 	}
 
 	// Clear and restore errors.
--- a/includes/llms.functions.core.php
+++ b/includes/llms.functions.core.php
@@ -359,9 +359,18 @@ function llms_get_dom_document( $string ) {
 	// Don't throw or log warnings.
 	$libxml_state = libxml_use_internal_errors( true );
 
+	// This forces DOMDocument to convert non-utf8 characters into HTML entities and without relying on `mb_convert_encoding()`.
+	$utf8_fixer = '<meta id="llms-get-dom-doc-utf-fixer" http-equiv="Content-Type" content="text/html; charset=utf-8">';
+
 	$dom = new DOMDocument();
-	if ( ! $dom->loadHTML( '<meta http-equiv="Content-Type" content="text/html; charset=utf-8">' . $string ) ) {
+	if ( ! $dom->loadHTML( $utf8_fixer . $string ) ) {
 		$dom = new WP_Error( 'llms-dom-document-error', __( 'DOMDocument XML Error encountered.', 'lifterlms' ), libxml_get_errors() );
+	}
+
+	// Remove the fixer meta element, if it's not removed it creates invalid HTML5 Markup.
+	$meta = $dom->getElementById( 'llms-get-dom-doc-utf-fixer' );
+	if ( $dom ) {
+		$meta->parentNode->removeChild( $meta );
 	}
 
 	// Clear and restore errors.
--- a/includes/class.llms.certificates.php
+++ b/includes/class.llms.certificates.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Classes
  *
  * @since 1.0.0
- * @version 4.3.1
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -178,6 +178,7 @@ class LLMS_Certificates {
 	 * @since 3.24.3 Unknown.
 	 * @since 3.37.3 Refactored method into multiple functions.
 	 * @since 4.3.1 If `$this->scrape_certificate()` generates a `WP_Error` early return it.
+	 * @since [version] Remove redundant check for the presence of `DOMDocument`.
 	 *
 	 * @param int $certificate_id WP_Post ID of the earned certificate.
 	 * @return WP_Error|string HTML of the certificate on success, otherwise an error object.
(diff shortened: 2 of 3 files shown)


eri-trabiccolo (commented):
I haven't had the chance to fully test this yet... I suppose we can rely on the fact that the blog charset is always UTF-8 (although that's an option that can be _potentially_, forced to something else) - because I don't know what would happen with two different meta charset set, which will win?.

Btw,
I just ran a test on a certificate save and it produced the following html (with 2021 theme):
```
<!DOCTYPE html>
<html><head><meta http-equiv="Content-Type" content="text/html; charset=utf-8"><meta charset="UTF-8"><meta name="viewport" content="width=device-width, initial-scale=1"><title>Certificate – WP</title><meta name="robots" content="noindex,nofollow"><style>
img.wp-smiley,
img.emoji {
	display: inline !important;
	border: none !important;
	box-shadow: none !important;
	height: 1em !important;
	width: 1em !important;
	margin: 0 .07em !important;
	vertical-align: -0.1em !important;
	background: none !important;
	padding: 0 !important;
}
</style><style>/*! This file is auto-generated */
[and so on....]
```

mmm, needs further investigations on my side, sorry for being late.

thomasplevy:
> I haven't had the chance to fully test this yet... I suppose we can rely on the fact that the blog charset is always UTF-8 (although that's an option that can be potentially, forced to something else) - because I don't know what would happen with two different meta charset set, which will win?.

I am nearly positive that this new method would not create a regression with a blog forced into a non utf-8 charset. This PR attempts to remove functions from the `mb_string` module. It's not attempting to address whatever issue might occur with non-utf8 blogs. But maybe I'm still not fully understanding the code. To be very frank I simply don't really have my head wrapped around use cases for different charsets or how to force a site into a different charset. If you understand this a bit can you augment these tests to create these scenarios and see if we are indeed introducing a regression? I decided that I did not want to drop in the polyfill because I don't understand this and I don't want to introduce a new dependency as a result of a lack of understanding. I did a lot of digging and tried several different methods and understand what's going on a bit better but this comment introduces new doubts. I need some concrete examples please.

And What's the issue with the HTML you're citing here? Is something broken?

eri-trabiccolo:
I see the reason behind this change. My comment was just to say that despite it looks good to me, since I found that, I wanted to make some deeper tests on whether we're introducing some regression. Buy I realize that I haven't said it properly, so maybe next time would be better I submit an actual review rather than casting possible time-consuming doubts. Sorry. <3

includes/llms.functions.core.php
```diff
@@ -359,7 +360,7 @@ function llms_get_dom_document( $string ) {
 	$libxml_state = libxml_use_internal_errors( true );
 
 	$dom = new DOMDocument();
-	if ( ! $dom->loadHTML( mb_convert_encoding( $string, 'HTML-ENTITIES', 'UTF-8' ) ) ) {
+	if ( ! $dom->loadHTML( '<meta http-equiv="Content-Type" content="text/html; charset=utf-8">' . $string ) ) {
```
eri-trabiccolo: can we turn this into:
```suggestion
    $string = $skip_utf8_encoding_conversion ? $string : string . '<meta http-equiv="Content-Type" content="text/html; charset=utf-8">' . $string;
	if ( ! $dom->loadHTML( $string ) ) {
```

where `$skip_utf8_encoding_conversion` is an additional parameter of this function set to `false` by default, so that when certificates calls this function it passes `true` as second parameter.
Otherwise we have an invalid html code:
`Error: A document must not include both a meta element with an http-equiv attribute whose value is content-type, and a meta element with a charset attribute.`
source https://validator.w3.org/nu/#textarea when pasting there the result of a certificate export.

This because when exporting a certificate we export the whole html which already contains a `<meta charset="UTF-8">`.

eri-trabiccolo (changes requested):
Cannot spot any real issue, can you just let me know what you think about and maybe apply my suggestion?!

includes/llms.functions.core.php
eri-trabiccolo: I was wrong on this, as this is actually what does the magic:
`'<meta http-equiv="Content-Type" content="text/html; charset=utf-8">'`

thomasplevy: hmmm hmmm mmm hmmm....

puzzled. Ill take another stab in a bit

thomasplevy: @eri-trabiccolo I was able to fix the markup error by removing the meta element afterwards.

It looks like as long as this is present when the HTML is loaded, it doesn't matter if it's removed when `saveHTML()` is called (which is where we'll see that invalid markup).

The tests were failing when you were working on this because with the removal it actually works the way it would have when using `mb_convert_encoding()` -- the tests were setup to check against the actual characters but with this method now it converts the characters to html entity codes.

The tests have been adjusted to accommodate and I think this is now better than what I initially submitted.

eri-trabiccolo (approved):

PR #1430, opened 2020-11-16
Fix "my courses" page results to only include the enrollment courses

alaa-alshamy:
## Description

Fix "my courses" page results to only include the enrollment courses, before it includes the cancelled courses also

Fixes #

## How has this been tested?

- Manually

## Types of changes

- Bug fix

## Checklist:


thomasplevy:
@alaa-alshamy this is expected behavior, not a bug.

If you'd like us to consider changing this behavior pleas open an issue outlining the bug and we can discuss ways to change the behavior in a way that doesn't cause users who *do expect this expected behavior* to experience a new bug or regression as a result of the changes made.

PR #1442, opened 2020-11-25
Feature/skip email

### LINKED ISSUE #1411: Filter `$skip_email` in class.llms.person.handler.php
I need to be able to skip the `email_exists()` check within class.llms.person.handler.php on line 977.

I have a need to be able to register users with the same email address. And am able to do so with filtering core WordPress functionality, but I can't filter the `$skip_email` value in LifterLMS or the `email_exists()` function in WordPress. And I realize this is a fringe use case.

A filter for editing the `$skip_email = false;` value on line 967 would work.

As an alternative I can edit the plugin directly, but then would no longer be able to receive auto updates.

jasonyingling:
Meant to get to this sooner but things delayed me.

## Description
Added a filter and docblock to modify the value of the `$skip_email` variable.

Fixes #1411 

## How has this been tested?
Tested in by changing the filter to `true` with the following in functions.php on a local site running WordPress 5.5.3.

Also ran the `composer run-script tests-run -- --group LLMS_Person_Handler` test script.

```
function cbt_skip_email( $skip_email ) {
	return true;
}
add_filter( 'llms_skip_email', 'cbt_skip_email', 10, 1 );
```

## Types of changes

## Checklist:

### DIFF
--- a/includes/class.llms.person.handler.php
+++ b/includes/class.llms.person.handler.php
@@ -971,7 +971,7 @@ class LLMS_Person_Handler {
 				 *
 				 * @param bool $skip_email A value of true will skip checking if the email exists.
 				 */
-				$skip_email = apply_filters( 'llms_skip_email', false );
+				$skip_email = apply_filters( 'llms_skip_new_user_email_exists_check', false );
 
 				// Only run this check when we're trying to change the email address for an account update.
 				if ( 'account' === $screen ) {
--- a/includes/class.llms.person.handler.php
+++ b/includes/class.llms.person.handler.php
@@ -965,7 +965,7 @@ class LLMS_Person_Handler {
 			if ( 'email_address' === $name ) {
 
 				/**
-				 * Allow for skipping of the emxail_exists check.
+				 * Allow for skipping of the email_exists check.
 				 *
 				 * @since [version]
 				 *
--- a/includes/class.llms.person.handler.php
+++ b/includes/class.llms.person.handler.php
@@ -967,7 +967,7 @@ class LLMS_Person_Handler {
 				/**
 				 * Allow for skipping of the emxail_exists check.
 				 *
-				 * @since x.x.x
+				 * @since [version]
 				 *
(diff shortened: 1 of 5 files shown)


includes/class.llms.person.handler.php
```diff
@@ -964,7 +964,14 @@ public static function validate_fields( $data, $screen = 'registration' ) {
 			// Check email field for uniqueness.
 			if ( 'email_address' === $name ) {
 
-				$skip_email = false;
+				/**
+				 * Allow for skipping of the emxail_exists check.
+				 *
+				 * @since x.x.x
```
thomasplevy: Per doc standards: https://github.com/gocodebox/lifterlms/blob/trunk/docs/documentation-standards.md#using-placeholders

```suggestion
				 * @since [version]
```

includes/class.llms.person.handler.php
```diff
@@ -964,7 +964,14 @@ public static function validate_fields( $data, $screen = 'registration' ) {
...
+				 *
+				 * @param bool $skip_email A value of true will skip checking if the email exists.
+				 */
+				$skip_email = apply_filters( 'llms_skip_email', false );
```
thomasplevy: As submitted I feel the hook name is too generic, something like this feels better to me:

```suggestion
				$skip_email = apply_filters( 'llms_skip_new_user_email_exists_check', false );
```

thomasplevy (changes requested):

thomasplevy:
@jasonyingling 

Thanks for your contribution here. 

A minor documentation issue that I'd have ignored had I not also wanted to request a change to the hook itself.

My thinking is that `llms_skip_email` is way too generic for what we're actually doing here with this variable, I've suggested a change above, thoughts?

jasonyingling:
@thomasplevy That's a much better name for it. I agree that was too generic. I was spinning wheels trying to best place it, but I think what you suggested works great.

thomasplevy:
Thanks @jasonyingling !

PR #1452, opened 2020-12-02
Use an HR in favor of a hidden h1 on add-ons admin screen

thomasplevy:
## Description

While testing the helper plugin on php8 I noticed that admin notices were getting "caught" between the inline title and the "My Keys" button.

This PR fixes this by using the `<hr>` element used by the WP core to signify the "end of the header" and it looks like this element "catches" admin notices in favor of putting them below the first `<h1>` tag.

Note that the issue is only apparent when the helper is activated and the change has no noticeable regressive impact when the helper is not installed.

## How has this been tested?

+ Manually

## Screenshots 

Without fix:
Screenshot_2020-12-02_14-34-35

With fix:
Screenshot_2020-12-02_14-33-16

Without helper (and fix):
Screenshot_2020-12-02_14-36-59

## Checklist:

### DIFF
--- a/includes/admin/class.llms.admin.addons.php
+++ b/includes/admin/class.llms.admin.addons.php
@@ -271,7 +271,7 @@ class LLMS_Admin_AddOns {
 	 * Output HTML for the current screen
 	 *
 	 * @since 3.5.0
-	 * @since 3.28.0 Unknown
+	 * @since 3.28.0 Unknown.
 	 * @since [version] Use `hr.wp-header-end` in favor of a second (hidden) <h1> to "catch" admin notices.
 	 *
 	 * @return void
--- a/includes/admin/class.llms.admin.addons.php
+++ b/includes/admin/class.llms.admin.addons.php
@@ -7,7 +7,7 @@
  * @package LifterLMS/Admin/Classes
  *
  * @since 3.5.0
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
@thomasplevy assigned to me? Or did you mean to just ask for my review?

thomasplevy:
YOU KNOW WHAT I MEANT.

eri-trabiccolo:
Not really man, I thought that maybe you just were tired and assigned it to me to finish it :D

eri-trabiccolo (approved):
Ah nice, didn't know about this js trick...

Who knows why we had that h1 duplication (although the second was just a screen-reader text).

thomasplevy:
> Who knows why we had that h1 duplication (although the second was just a screen-reader text).

The screen reader `<h1>` was intended to do what the `<hr>` is doing in this PR. I haven't traced this to a change in the WP core but I'm thinking that at some point they added this `<hr>` and in doing so modified the way the notices are "caught" with JS and that created the bug I've found.

I don't know this for sure, it's an assumption based on my own faulty memory that this didn't always work like this and that hidden `<h1>` was a hack to do this same thing.

/shrug

eri-trabiccolo:
thanks Thomas <3

PR #1466, opened 2020-12-22
Change the title of the instructors block to be suitable for course and membership cases

### LINKED ISSUE #1464: Some js texts are missing from translation
### Reproduction Steps

+ Open "Course builder" of any course
+ Edit any lesson
+ Hover over the tool tip of "Free lesson" for example
+ The text of the tool tip is not translatable ( because it's not exist in translation template(pot) to be translated )

see this screenshot:
Screenshot from 2020-12-21 10-57-00

### Technical description

After some investigation i figured that those texts are not added to the POT file because they are not added to the "get_strings" function in this line:
https://github.com/gocodebox/lifterlms/blob/trunk/includes/class.llms.l10n.js.php#L38

and this file actually compiled automatically via the gulp as described in the head

```
<?php
/**
 * Localize JS strings
 * This file should not be edited directly
 * It is compiled automatically via the gulp task `pot-js`
 * See the lifterlms-lib-tasks package for more information
 *
 * @package  LifterLMS/Classes/Localization
 * @since    3.17.8
 * @version  4.7.0
 */
```
And i think the gulp script not adding them because those texts used in double quote in this line:
https://github.com/gocodebox/lifterlms/blob/trunk/assets/js/builder/Schemas/Lesson.js#L39

and the gulp script expect them to be in single quote in this line:
https://github.com/gocodebox/lifterlms-lib-tasks/blob/master/tasks/pot-js.js#L28

So if you searched about `LLMS.l10n.translate( "` you will find all those texts i mean

### Expected Behavior

+ It should be added to the POT file to be translated

### Actual Behavior

+ it's not exist in the POT file

### Error Messages / Logs

```

```

<details>

```


```

</details>

This issue has be recreated:
+ [ ] Locally
+ [ ] On a staging site
+ [ ] On a production website

+ Browser name and version

alaa-alshamy:
## Description

This pull request is to complete the work of following pull request in the lifterlms-block repo: https://github.com/gocodebox/lifterlms-blocks/pull/75

and it's just to make the title of the instructors block suitable for all cases (course, membership)

## How has this been tested?

Manually

## Checklist:

### DIFF
--- a/templates/course/author.php
+++ b/templates/course/author.php
@@ -3,7 +3,8 @@
  * LifterLMS Course Author Info
  *
  * @since   3.0.0
- * @version 3.25.0
+ * @since [version] Use post type's name in the title to cover membership case.
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -20,7 +21,7 @@ $count = count( $instructors );
 ?>
 
 <section class="llms-instructor-info">
-	<h3 class="llms-meta-title"><?php echo _n( 'Instructor', 'Instructors', count( $instructors ), 'lifterlms' ); ?></h3>
+	<h3 class="llms-meta-title"><?php printf( _n( '%s Instructor', '%s Instructors', count( $instructors ), 'lifterlms' ), $course->get_post_type_label() ); ?></h3>
 	<div class="llms-instructors llms-cols">
 		<?php foreach ( $instructors as $instructor ) : ?>
 			<div class="llms-col-<?php echo $count <= 4 ? $count : 4; ?>">
--- a/templates/course/author.php
+++ b/templates/course/author.php
@@ -20,7 +20,7 @@ $count = count( $instructors );
 ?>
 
 <section class="llms-instructor-info">
-	<h3 class="llms-meta-title"><?php echo _n( 'Course Instructor', 'Course Instructors', count( $instructors ), 'lifterlms' ); ?></h3>
+	<h3 class="llms-meta-title"><?php echo _n( 'Instructor', 'Instructors', count( $instructors ), 'lifterlms' ); ?></h3>
 	<div class="llms-instructors llms-cols">
 		<?php foreach ( $instructors as $instructor ) : ?>
 			<div class="llms-col-<?php echo $count <= 4 ? $count : 4; ?>">


thomasplevy (changes requested):
@alaa-alshamy 

Thanks for your work here, however, I'm hesitant to make this language change to be more generic. Any change we make like this (where language on the frontend of the website changes) has the potential to create unhappiness among existing users who don't wish to have their language changed.

When there's an easy way to *undo* the change it's less of a problem but from experience a change like this will frustrate a (relatively) small number of users.

In order to preserve the experience for these users it would be better to pull in the post type's name information from the `LLMS_Post_Model` class and use that in favor of removing the post name all together.

Here's the method that can be used to easily pull in the post type's label: https://developer.lifterlms.com/reference/classes/llms_post_model/get_post_type_label/

Thanks!

alaa-alshamy:
@thomasplevy I did the requested change and updated the docblock

but i have a question, should we move the template to outside the course folder after it used from membership also ?

thomasplevy:
@alaa-alshamy good question... this is why nothing is ever as simple as it seems.

We *should* have a semantic name for the shared template file but as soon as we change the name we also need to deprecate a file, probably deprecate at least one function, and then modify several more.

I think the *best* approach is to have a shared template which is loaded through the existing course/author.php and a new membership/instructors.php files. Similar to how the course/membership archive files both include the main loop.php file.

This will allow individual overrides of each template to allow developers to customize the two templates independently (easily).

However, there's quite a few changes that should be made in order to make this happen and the dynamic block loader will have to change to accommodate.

I don't have the internal capacity to work on this right now I welcome the help from your side. Otherwise we'll have to hang out for a bit until we can make all the necessary changes.

I am also kind of okay with letting this slide and making these changes later and having memberships call the course file despite this being semantically incorrect as far as where the file is stored and what it's called. 

@eri-trabiccolo do you want to weigh in on this?

alaa-alshamy:
@thomasplevy I got your point, and if it's ok to postponing this changes that's will be great

eri-trabiccolo:
@thomasplevy 
I can work on this yeah.

thomasplevy:
Closing in favor of #1475

PR #1474, opened 2020-12-31
Composer libs

### LINKED ISSUE #1473: Move vendor packages to a directory not ignored by `wp i18n make-pot` command
Strings from vendor-included package plugins (lifterlms-blocks and lifterlms-rest) are not translatable via GlotPress (despite their inclusion in the plugin-included pot file.

My previous understanding is that all strings included in the plugin's language directory are imported into the GlotPress project for the plugin but it turns out there is a "build step" that generates these strings using `wp i18n make-pot` and imports the generated pot file into the GlotPress install.

As a result, our strings need not be included in the pot file included with the plugin (and by extension including a pot file I suppose isn't even necessary, but I'm not touching that because I cannot find a single translatable plugin that doesn't include one).

Details of what I've learned in Slack at https://wordpress.slack.com/archives/C02RP50LK/p1606857219220800

To solve this problem our composer packages should be moved to be installed in directory that will not be ignored by composer.

There's a few other things to think about here:

1) We have packages that we wish should remain ignored: WC Action Scheduled and WP Background Process.

We should move the background process lib into the vendor directory and leave action scheduler where it is.

The background processor files should be deprecated in their current location and moved out to the vendor lib.

2) lifterlms-blocks and lifterlms-lms rest should be moved into a new directory, I propose `libs` which is not ignored by the default wp cli command

3) We can use the [composer/installers](https://github.com/composer/installers) to customize the directory where packages are installed

thomasplevy:
## Description

Fixes #1473

+ Moves core-maintained "external" library plugins to be installed into `/libraries/` in favor of default composer `/vendor/` dir.
+ Installs WP Background Processing via composer (into the vendor dir) in favor of manual installation and version control

## How has this been tested?

+ Manually
+ Existing unit tests

## Types of changes

+ Bug fix
+ Deprecation of 2 files (WP Background Process), they're moved to the default composer vendor directory.
+ Technically moving the location of the library plugins out of the composer directory could be noted as a backwards incompatible change? In order to properly move all of these (with deprecation warnings) we'd need to have, essentially, two copies of the libraries included. Perhaps this should be "held" for LifterLMS 5.0 and we can note the movement of the library as a breaking change during that upgrade. We'll need to release a 5.0.0 along side the forthcoming fields updates so this would be held for maybe a month (or so) in this scenario.

## Checklist:


includes/processors/class.llms.processors.php
```diff
@@ -89,30 +92,26 @@ public function get( $name ) {
...
+		/**
+		 * Filter the list of available processors to be loaded
+		 *
+		 * Third parties can use this filter to load custom processors.
+		 *
+		 * @since [version]
+		 *
+		 * @see llms_load_processor_path To add a custom load path for the loaded processor.
+		 *
+		 * @param string[] $this->classes A list of processor class ids/slugs.
```
eri-trabiccolo:
```suggestion
		 * @param string[] $classes A list of processor class ids/slugs.
```

tests/phpunit/unit-tests/processors/class-llms-test-processors.php
```diff
@@ -0,0 +1,75 @@
...
+		$this->main = LLMS_Processors::instance();
+	}
+
+	/**
+	 * Test instance()
+	 *
+	 * @since [version]
+	 *
+	 * @runInSeparateProcess
+	 * @preserveGlobalState disabled
```
eri-trabiccolo, line 31: uh interesting. The reason behind this choice, just "scoping"?

eri-trabiccolo (changes requested):
Looks fine to me.

tests/phpunit/unit-tests/processors/class-llms-test-processors.php
thomasplevy: The instance is instantiated in main `llms()` instance so `@runInSeparateProcess` lets us instantiate a "clean" instance.

The `@preserveGlobalState disabled` is due to this: https://core.trac.wordpress.org/ticket/50482

thomasplevy:
> Technically moving the location of the library plugins out of the composer directory could be noted as a backwards incompatible change? In order to properly move all of these (with deprecation warnings) we'd need to have, essentially, two copies of the libraries included. Perhaps this should be "held" for LifterLMS 5.0 and we can note the movement of the library as a breaking change during that upgrade. We'll need to release a 5.0.0 along side the forthcoming fields updates so this would be held for maybe a month (or so) in this scenario.

Thoughts on this? I'd like to just ship it but you never know what kinds of problems might arise so maybe it'd be better to do it in 5.0 just to be safe?

thomasplevy:
@eri-trabiccolo :arrow_up:

tests/phpunit/unit-tests/processors/class-llms-test-processors.php
eri-trabiccolo: thanks!

eri-trabiccolo:
@thomasplevy 
I see what you mean. Yeah I mean I can see issues with some kind of aggressive caching.
Yes let's do it in 5.0.

PR #1478, opened 2021-01-05
LifterLMS 5.0.0 (Working Branch)

thomasplevy:


languages/states.php
```diff
@@ -0,0 +1,4429 @@
+<?php
+/**
+ * States
+ *
+ * Returns a multi-demensional array of countries and country states (or provinces / regions) and their respective codes.
+Countries with an empty array have no states.
```
eri-trabiccolo:
```suggestion
 * Returns a multi-dimensional array of countries and country states (or provinces / regions) and their respective codes.
 * Countries with an empty array have no states.
```

languages/states.php
```diff
@@ -0,0 +1,4429 @@
...
+ * modify or submit pull requests on this file directly. If you've located *
+ * an issue with any of the data contained within this file please open a  *
+ * new issue at https://github.com/gocodebox/lifterlms/issues/new/choose.  *
+ *                                                                         *
+ * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
+ *
+ * @package LifterLMS/i18n
+ *
+ * @since 5.0.0
+ * @version [vesion]
```
eri-trabiccolo:
```suggestion
 * @since [version]
 * @version [version]
```

languages/states.php
```diff
@@ -0,0 +1,4429 @@
...
+		'57' => __( 'The Marches', 'lifterlms' ),
+		'62' => __( 'Latium', 'lifterlms' ),
+		'65' => __( 'Abruzzo', 'lifterlms' ),
+		'67' => __( 'Molise', 'lifterlms' ),
+		'72' => __( 'Campania', 'lifterlms' ),
+		'75' => __( 'Apulia', 'lifterlms' ),
+		'77' => __( 'Basilicate', 'lifterlms' ),
+		'78' => __( 'Calabria', 'lifterlms' ),
+		'82' => __( 'Sicily', 'lifterlms' ),
+		'88' => __( 'Sardinia', 'lifterlms' ),
```
eri-trabiccolo: This is a "problem".
The list here is for italian regions.
In Italy we don't use them when specifying a billing (or even shipping) address. We use to specify what we call Provinces. Example.
I'm in Pinerolo (city/town) which is in province of Turin (province) in the Piedmont region.
Naples is both a city/town and a province in the Campania region.

When I specify my address I say:
Pinerolo (TO): TO is the italian province code for Turin.
This is the actual list we should use for italian provinces:
https://github.com/woocommerce/woocommerce/blob/e3b69b901143de3c42d90b9d4bdc97a8821133e2/i18n/states.php#L645

Maybe we should double check (compare with wc) other countries' states/provinces too...

languages/countries.php
```diff
@@ -0,0 +1,286 @@
...
+ * modify or submit pull requests on this file directly. If you've located *
+ * an issue with any of the data contained within this file please open a  *
+ * new issue at https://github.com/gocodebox/lifterlms/issues/new/choose.  *
+ *                                                                         *
+ * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
+ *
+ * @package LifterLMS/i18n
+ *
+ * @since 5.0.0
+ * @version [vesion]
```
eri-trabiccolo:
```suggestion
 * @since [version]
 * @version [version]
```

eri-trabiccolo (changes requested):
Just something I noticed while testing the forms. :D

languages/states.php
thomasplevy: Thank you, fixed in https://github.com/gocodebox/dev-tools/commit/390f7625404f4367267957a71b32e111f3c07ad0#diff-2e062f03cfac811baea27d27f0fc3b2b13174df3f095b88fce4c90766ca72d47R184

languages/states.php
thomasplevy: Fixed in https://github.com/gocodebox/dev-tools/commit/629cd8cc54952373299170728323b77228456c50

languages/countries.php
thomasplevy: https://github.com/gocodebox/dev-tools/commit/629cd8cc54952373299170728323b77228456c50

languages/states.php
eri-trabiccolo: Ohhh sorry, I forgot that.

eri-trabiccolo (commented):
Jeez the suspance created by this commit message is going to kill me... 😜

Issue #1484, opened 2021-01-08
Add a Database Indices to Improve Performance of Frequently-Running Queries


reedhewitt:
The short version: adding a few new indices on a couple of database tables made one of the often-running background processes in LifterLMS run seven times faster.  I'm posting it here in case it helps anyone else and in case it can be included in a future release of LifterLMS.

**Is your feature request related to a problem? Please describe.**

We noticed that LifterLMS was always running some variation of the following query (and by “always”, I mean these queries were literally running 24/7), and it appeared to be triggered by the cron event “llms_calculate_course_data”:

```sql
SELECT
    SQL_CALC_FOUND_ROWS u.ID AS id,
    (
      SELECT
          meta_value
        FROM
          wp_lifterlms_user_postmeta
        WHERE
          meta_key = '_status'
          AND user_id = id
          AND post_id IN (274)
        ORDER BY
          updated_date DESC LIMIT 1
    ) AS status,
    (
      SELECT
          updated_date
        FROM
          wp_lifterlms_user_postmeta
        WHERE
          meta_key = '_status'
          AND user_id = id
          AND post_id IN (274)
        ORDER BY
          updated_date DESC LIMIT 1
    ) AS `date`,
    m_last.meta_value AS last_name,
    m_last.meta_value AS first_name
  FROM
    wp_users AS u JOIN wp_usermeta AS m_first
      ON u.ID = m_first.user_id
    AND m_first.meta_key = 'first_name' JOIN wp_usermeta AS m_last
      ON u.ID = m_last.user_id
    AND m_last.meta_key = 'last_name'
  HAVING
    status IS NOT NULL
    AND status IN ('enrolled')
  ORDER BY
    date DESC,
    status ASC,
    last_name ASC,
    first_name ASC,
    id ASC
  LIMIT 0, 100
```

The limit number and post_id would change periodically as the batch job progressed.  Our WordPress instance has over 34,000 users, so it would take a long time to cycle through them all in sets of 100.  Also, the query shown above would take over 40 seconds to run, even on a big server (64GB RAM and 16 CPU cores) without much other load at the time.  Basically, one of our CPU cores was dedicated to running that query all day every day.

Looking in the LifterLMS Status > Tools & Utilities page, I saw an option to clear all the pending batch processes.  There were over 100,000 of them pending.  It appears that the batch processes were being initiated faster than they could be completed, and the queue was just getting deeper over time.

**Describe the solution you'd like**

[...]

PR #1485, opened 2021-01-11
Improvements (mostly performance) to the course data background processor

thomasplevy:
## Description

Per #1484 

## How has this been tested?

New unit tests

## Types of changes

+ Performance improvements
+ Bug fixes

## Checklist:

### DIFF
--- a/includes/processors/class.llms.processor.course.data.php
+++ b/includes/processors/class.llms.processor.course.data.php
@@ -43,18 +43,16 @@ class LLMS_Processor_Course_Data extends LLMS_Abstract_Processor {
 
 	/**
 	 * Maximum number of students allowed in a course
-	 * when enrollment is higher than this number
-	 * throttling the calculations will be delayed
+	 *
+	 * When enrollment is higher than this number
+	 * throttling the calculations will be delayed.
 	 *
 	 * @var int
 	 */
 	private $throttle_max_students;
 
 	/**
-	 * When a calculation is throttled based on the number of students
-	 * this will determine the frequency with which the query can run
-	 * Should be time in seconds
-	 * default is HOUR_IN_SECONDS * 4
+	 * Frequency of calculation process when the process is throttled.
 	 *
 	 * @var int
 	 */
@@ -63,9 +61,9 @@ class LLMS_Processor_Course_Data extends LLMS_Abstract_Processor {
 	/**
 	 * Called when queue is emptied and process is complete
 	 *
-	 * @return   void
-	 * @since    3.15.0
-	 * @version  3.15.0
+	 * @since 3.15.0
+	 *
+	 * @return void
 	 */
 	protected function complete() {
 
@@ -77,16 +75,16 @@ class LLMS_Processor_Course_Data extends LLMS_Abstract_Processor {
 	/**
 	 * Action triggered to queue queries needed to make the calculation
 	 *
-	 * @param    int $course_id  WP Post ID of the course
-	 * @return   void
-	 * @since    3.15.0
-	 * @version  3.15.0
+	 * @since 3.15.0
+	 *
+	 * @param int $course_id WP Post ID of the course.
+	 * @return void
 	 */
 	public function dispatch_calc( $course_id ) {
 
 		$this->log( sprintf( 'course data calculation dispatched for course %d', $course_id ) );
 
-		// cancel process in case it's currently running
+		// Cancel process in case it's currently running.
 		$this->cancel_process();
 
 		$args = array(
@@ -96,16 +94,16 @@ class LLMS_Processor_Course_Data extends LLMS_Abstract_Processor {
 			'per_page' => 100,
 		);
 
-		// get total number of pages
+		// Get total number of pages.
 		$query = new LLMS_Student_Query( $args );
 
-		// only queue if we have students in the course
+		// Only queue if we have students in the course.
 		if ( $query->found_results ) {
 
-			// throttle dispatch?
+			// Throttle dispatch?.
 			if ( $this->maybe_throttle( $query->found_results ) ) {
 
-				// schedule to run again in the future
+				// Schedule to run again in the future.
 				$last_run = $this->get_data( 'last_run', 0 );
 				$this->schedule_calculation( $course_id, $last_run + $this->throttle_frequency );
 
@@ -114,7 +112,7 @@ class LLMS_Processor_Course_Data extends LLMS_Abstract_Processor {
 
 			}
 
-			// add each page to the queue
+			// Add each page to the queue.
 			while ( $args['page'] <= $query->max_pages ) {
 
 				$this->push_to_queue( $args );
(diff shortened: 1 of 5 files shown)


includes/processors/class.llms.processor.course.data.php
```diff
@@ -172,45 +190,109 @@ protected function init() {
...
+		 * @param LLMS_Processor_Course_Data $processor       Instance of the data processor class.
+		 */
+		$this->throttle_max_students = apply_filters( 'llms_data_processor_course_data_throttle_count', 500, $this );
+
+		/**
+		 * Throttles course data processing based on the number of a students in a course.
+		 *
+		 * If the number of students in a course is greater than or equal to this number, the background
+		 * process will be throttled to run only once every N hours where N is equal to the number of hours
+		 * defined by the `llms_data_processor_course_data_throttle_frequency` filter.
```
eri-trabiccolo: copy&paste

includes/processors/class.llms.processor.course.data.php
```diff
@@ -172,45 +190,109 @@ protected function init() {
...
+		 */
+		$this->throttle_frequency = apply_filters( 'llms_data_processor_course_data_throttle_frequency', HOUR_IN_SECONDS * 4, $this );
+
+	}
+
+	/**
+	 * Determines if the supplied course is already being processed.
+	 *
+	 * If it's already being processed we'll throttle the processing so we'll wait until the course
+	 * completes it's current data processing and start again later.
```
eri-trabiccolo:
```suggestion
	 * completes its current data processing and start again later.
```

includes/processors/class.llms.processor.course.data.php
```diff
@@ -172,45 +190,109 @@ protected function init() {
...
+	 *
+	 * @since [version]
+	 *
+	 * @param [type] $course_id [description]
+	 * @return boolean [description]
```
eri-trabiccolo: :)

includes/processors/class.llms.processor.course.data.php
```diff
@@ -172,45 +190,109 @@ protected function init() {
...
-			return ( ( time() - $last_run ) <= $this->throttle_frequency );
+		} elseif ( $num_students >= $this->throttle_max_students ) {
+
+			$throttled = ( time() - $this->get_last_run( $course_id ) <= $this->throttle_frequency );
 
 		}
 
-		return false;
+		/**
+		 * Filter whether or not data processing in throttled for a request
```
eri-trabiccolo:
```suggestion
		 * Filters whether or not data processing is throttled for a request
```

includes/processors/class.llms.processor.course.data.php
```diff
@@ -265,37 +348,34 @@ public function schedule_calculation( $course_id, $time = null ) {
...
-	 * @since    3.15.0
-	 * @version  3.15.0
+	 * Execute calculation for each item in the queue until all students in the course have been polled
+	 *
+	 * Stores the data in the postmeta table to be accessible via LLMS_Course.
+	 *
+	 * @since 3.15.0
+	 * @since [version] Moved task completion logic to `task_complete()`.
+	 *
+	 * @param array $args Query arguments passed to LLMS_Student_Query..
```
eri-trabiccolo:
```suggestion
	 * @param array $args Query arguments passed to LLMS_Student_Query.
```

includes/processors/class.llms.processor.course.data.php
```diff
@@ -172,45 +190,109 @@ protected function init() {
...
+	 *
+	 * @since [version]
+	 *
+	 * @param [type] $course_id [description]
+	 * @return boolean [description]
+	 */
+	protected function is_already_processing_course( $course_id ) {
+
+		if ( ! $this->is_queue_empty() ) {
+			return in_array( $course_id, wp_list_pluck( $this->get_batch()->data, 'post_id' ), true );
```
eri-trabiccolo: I'm not sure this does **exactly** what we want.
When I look at `WP_Background_Process::get_batch()` I see it returns the first batch of the type "course_data" in the queue, so I guess that ideally there can be N "processing courses" but we only check the first of the queue.
Are we fine with this? I think we "can accept it"...

includes/processors/class.llms.processor.course.data.php
```diff
@@ -43,107 +54,114 @@ class LLMS_Processor_Course_Data extends LLMS_Abstract_Processor {
...
 		$query = new LLMS_Student_Query( $args );
 
-		// only queue if we have students in the course
+		// Only queue if we have students in the course.
 		if ( $query->found_results ) {
 
-			// throttle dispatch?
-			if ( $this->maybe_throttle( $query->found_results ) ) {
+			// Throttle dispatch based on number of students or because the course is already processing.
+			if ( $this->maybe_throttle( $query->found_results, $course_id ) ) {
```
eri-trabiccolo: If the course is already processing we actually can throttle _before_ any query.
I think is reasonable considering that if the action that triggers the `dispatch_calc` has been fired, and the course is already processing then there are enrolled students... and even if they're not it's something that it'll be addressed later. Meantime we've avoided an expensive query. What do you think?
If you agree with this, I would make a check on if `is_already_processing_course( $course_id )` before performing the query. Pseudo-code:
```php
$throttle = $this->is_already_processing_course( $course_id );
if ( ! $throttle ) {
	$query    = new LLMS_Student_Query( $args );
	$throttle = $this->maybe_throttle( $query->found_results, $course_id, false );
}
if ( $throttle ) { 
	do_throttle();
	return;
}
// no throttling.
if ( $query->found_results ) {
	// Add each page to the queue.
	// Save queue and dispatch the process.
}
```
The `false` as third parameter for `maybe_throttle()` method would be to avoid checking on if the course is already processing twice:
```php
	protected function maybe_throttle( $num_students, $course_id, $check_already_processing_course = true ) {

		$throttled = false;

		if ( $check_already_processing_course && $this->is_already_processing_course( $course_id ) ) {

			$throttled = true;

		} 
```

well this is the _idea_. What do you think?

eri-trabiccolo (changes requested):
Aside from the changes requested there's also a possible improvement suggested.

includes/processors/class.llms.processor.course.data.php
thomasplevy: mmm, yea good point, I missed that...

I was thinking about "locking" a course with temporary postmeta data and decided not to because I could pull it this way but I didn't realize it only returned a single batch.

I'll fix it.

thomasplevy: Most recent push uses course meta data to create (and check) a lock

includes/processors/class.llms.processor.course.data.php
thomasplevy: I like this idea, makes sense. Especially more so when checking the meta data to determine if the course's processing is locked. Much better.

I wrote the update a bit differently than your pseudo code here but I think it achieves the same end. Thoughts?

eri-trabiccolo (approved):

PR #1492, opened 2021-01-15
Staging updates

thomasplevy:
## Description

Updates staging and site feature code

The main goal is to ensure that visiting the admin panel (and logging in as an admin) is not required to automatically disable site features (like recurring payments) when a cloned site is detected.

There's two changes of note:

1) The method that is used to check if the site has been cloned runs on `init` in favor of `admin_init`

This change makes it so that after migrating a site it is *not necessary* to login to the site as an admin as a trigger for recurring payments to be disabled.

The current system assumes that no one would ever clone a site and then not immediately login to it.

2) I added the ability to define site features via constants. The reasoning for this would be to allow developers to define things that are always true in a particular environment.

For example, `define( 'LLMS_SITE_FEATURE_RECURRING_PAYMENTS`, false );` could be added to a staging/test site's wp-config.php file and that would not require *any login at all* to ensure that recurring payments are disabled.

## How has this been tested?

+ Manually
+ New & existing unit tests

## Types of changes

+ Improvements, bug fixes

## Checklist:

### DIFF
--- a/includes/class-llms-staging.php
+++ b/includes/class-llms-staging.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Classes
  *
  * @since 3.32.0
- * @version 3.35.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -14,7 +14,6 @@ defined( 'ABSPATH' ) || exit;
  * LLMS_Staging class.
  *
  * @since 3.32.0
- * @since 3.35.0 Sanitize input data.
  */
 class LLMS_Staging {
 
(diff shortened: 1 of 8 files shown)


thomasplevy:
The only add-on using the hook which might be affected is the LifterLMS Helper

https://github.com/gocodebox/lifterlms-helper/blob/trunk/includes/class-llms-helper-cloned.php

After a review of the code I don't believe this impacts the helper at all

includes/class-llms-staging.php
```diff
@@ -87,12 +119,39 @@ public static function menu_warning() {
...
 			if ( 'edit.php?post_type=llms_order' === $item[2] ) {
-				$menu[ $index ][0] .= ' <span class="update-plugins">' . esc_html__( 'Staging', 'lifterlms' ) . '</span>';
+				$menu[ $index ][0] .= self::get_menu_warning_bubble();
 			}
 		}
 
 	}
 
+	/**
+	 * Output a notice informing the use the site was put into staging mode.
```
eri-trabiccolo:
```suggestion
	 * Output a notice informing the user the site was put into staging mode.
```

eri-trabiccolo (changes requested):
Looks good to me, except for a minor typo in the doc.

PR #1493, opened 2021-01-16
Add a hook for 3rd party tools

thomasplevy:
## Description

While working on https://github.com/gocodebox/lifterlms-gateway-authorize-net/issues/10 I noticed that it's quite difficult to load a custom admin tool from outside the core plugin and ensure it also triggers properly

This is due to the loading order and triggered actions.

I was able to get a tool display by loading the tool class at `admin_init` with priority 10 or later but was unable to get the `handle()` method of the tool class to run.

This adds an action which can be used to load the tool immediately after core tools and, as such, it will both display and be handled as expected.

## How has this been tested?

+ Manually

## Types of changes

+ 3rd party compat enchancement

## Checklist:

### DIFF
--- a/includes/admin/class.llms.admin.menus.php
+++ b/includes/admin/class.llms.admin.menus.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Admin/Classes
  *
  * @since 1.0.0
- * @version 4.7.0
+ * @version [version]
  */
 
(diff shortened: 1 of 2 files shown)


eri-trabiccolo (approved):
Although I would have loved to see some proper tests... :P

PR #1497, opened 2021-01-21
Staging Feature "Reversion" (kind of)

thomasplevy:
## Description

Reverts (one) change made in #1492 resorting previous functionality that runs site duplication checks only on the admin panel for logged in users.

One user (so far) reported a false-positive in production as a result of these changes. We haven't been able to identify the source but after internal discussion we've determined it is best to rollback this change.

Additionally adds another constant `LLMS_SITE_IS_CLONE` which can be used to force a site as a clone (or not a clone)

Fixes an issue @eri-trabiccolo identified (but we never reported) when using the "Reset Automatic Payment Status" tool on the admin panel. When clicking the tool the page reloads but the notice wouldn't display until another subsequent page load (on the admin panel). This forces the check and then redirects to prevent that issue.

## How has this been tested?

+ Manually
+ New Tests
+ Existing Tests

## Types of changes

+ Revert 

## Checklist:

### DIFF
--- a/tests/phpunit/unit-tests/class-llms-test-staging.php
+++ b/tests/phpunit/unit-tests/class-llms-test-staging.php
@@ -28,6 +28,7 @@ class LLMS_Test_Staging extends LLMS_Unit_Test_Case {
 	 * Test clone_detected()
 	 *
 	 * @since 4.12.0
+	 * @since [version] Add tests for all potential conditions.
 	 *
 	 * @return void
 	 */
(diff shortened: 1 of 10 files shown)


eri-trabiccolo (approved):

eri-trabiccolo:
there's a test here that is breaking some view manager tests with PHP 7.3 and WP 5.2.
Annoying...

thomasplevy:
Yea, I'm not sure what's happening there... It's on my task list for the day

eri-trabiccolo:
@thomasplevy 
I was troubleshooting this, this is the culprit test -> https://github.com/gocodebox/lifterlms/pull/1497/files#diff-3644d4066da32948d6570e5923d025094cbf2d91d881f178150f193d56acbef5R121
I think it's the `set_current_screen()`.

tests/phpunit/unit-tests/class-llms-test-staging.php
```diff
@@ -137,6 +137,9 @@ public function test_clone_detected() {
 		$this->assertTrue( LLMS_Admin_Notices::has_notice( 'maybe-staging' ) );
 		LLMS_Admin_Notices::delete_notice( 'maybe-staging' );
 
+		// Return to front.
+		set_current_screen( 'front' );
```
thomasplevy, line 141: @eri-trabiccolo here's the line... :rage:

thomasplevy:
> @thomasplevy
> I was troubleshooting this, this is the culprit test -> https://github.com/gocodebox/lifterlms/pull/1497/files#diff-3644d4066da32948d6570e5923d025094cbf2d91d881f178150f193d56acbef5R121
> I think it's the `set_current_screen()`.

OH MAN. I just spent the last hour or so reveiwing new tests to arrive (independently) at the same conclusion... :rage:

tests/phpunit/unit-tests/class-llms-test-staging.php
eri-trabiccolo: @thomasplevy yeah... I actually was running some tests, and found the same :D <3

PR #1348, opened 2020-09-22
Upcoming payment reminder

niluzok:
## Description
I added a notification about the upcoming recurrent payment. Notification is scheduled on 1 day before payment due date.

## How has this been tested?
I ve done only manual testing. I created several orders with recurrent payments, than looked for scheduled actions and their dates: actual payment and notification the day before. Than i ran this scheduled actions by hand and watched as they reappeared. Than i tested order due date change and order cancelation.

## Types of changes
- New notification controller + view
- Added code for scheduling and unscheduling notification invocation in order model next to code for scheduling and unscheduling or recurrent payments

## Checklist:
- [ ] My code has been tested.
- [ ] My code passes all existing automated tests. 
- [x] My code follows the LifterLMS Coding & Documentation Standards.

PR #1545, opened 2021-02-26
Upcoming payment reminder

eri-trabiccolo:
## Description
This aimed just to add some tests to https://github.com/gocodebox/lifterlms/pull/1348
but as often happens revealed to be a little bit more complicated than it was supposed to be.
I've also took the occasion to update docs and cs of "involved" files.
Improve (hopefully) the logic, abstract something, un-abstract (sic!) something else.

## How has this been tested?
manually, existing and new unit tests

## Types of changes

## Checklist:


includes/class.llms.emails.php
```diff
@@ -196,17 +221,150 @@ public function get_emails() {
...
+	 * Return an associative array with the table's tags inline style
+	 *
+	 * It makes sure that all the required tags (table, tr, td) are set.
+	 *
+	 * @since [version]
+	 *
+	 * @param string Optional. Any of ARRAY_A | ARRAY_N constants. Default is ARRAY_N.
+	 *               With ARRAY_A returns an associative array whose keys are the tags `table`, `tr`, `td`, and their styles as values.
+	 *               With ARRAY_N returns an array with the following order 0 = table style, 1 = tr style, 2 = td style.
+	 *               Will defaults to ARRAY_N if the provided value is different than ARRAY_A | ARRAY_N constants.
```
thomasplevy: Missing param name

```suggestion
	 * @param string $output Optional. Any of ARRAY_A | ARRAY_N constants. Default is ARRAY_N.
	 *                       With ARRAY_A returns an associative array whose keys are the tags `table`, `tr`, `td`, and their styles as values.
	 *                       With ARRAY_N returns an array with the following order 0 = table style, 1 = tr style, 2 = td style.
	 *                       Will defaults to ARRAY_N if the provided value is different than ARRAY_A | ARRAY_N constants.
```

includes/class.llms.emails.php
```diff
@@ -196,17 +221,150 @@ public function get_emails() {
...
+	 * @return array
+	 */
+	private function get_parsed_table_style( $output = ARRAY_N ) {
```
thomasplevy: After reviewing the usage of this I think it might make sense to just drop the logic for returning an `ARRAY_A`. If we're only going to use the numeric version internally I don't see why bothering to have an alternative version, especially on a private function.

While I do think that an associative array is easier to understand, parse, and work with, if we don't need it why bother?

Alternatively, perhaps we can drop the `ARRAY_N` and always return the associative array. When using it with `list()` later we could pass it through `array_values()`.

Or perhaps I'm missing a good reason for having both.

includes/class.llms.emails.php
```diff
@@ -196,17 +221,150 @@ public function get_emails() {
...
+	}
+
+	/**
+	 * Return the table's `table` tag inline style
+	 *
+	 * @since [version]
+	 *
+	 * @return string
+	 */
+	protected function get_table_table_style() {
```
thomasplevy, line 299: what a method name :laughing:

includes/controllers/class.llms.controller.orders.php
```diff
@@ -169,14 +170,44 @@ public function complete_order( $order, $old_status ) {
 
 		unset( LLMS()->session->llms_coupon );
 
-		// Trigger order complete action.
+		/**
+		 * Action fired on order complete
+		 *
+		 * Prior to the students being enrolled.
+		 *
+		 * @since Unknown
```
thomasplevy: This is almost certainly from version 1

```suggestion
		 * @since 1.0.0
```

includes/controllers/class.llms.controller.orders.php
```diff
@@ -477,39 +521,66 @@ public function pending_cancel_order( $order ) {
 	 * @since 3.0.0
 	 * @since 3.32.0 Record order notes and trigger actions during errors.
 	 * @since 3.36.1 Made sure to process only proper LLMS_Orders of existing users.
+	 * @since [version] Fixed buggy logging on gateway error because it doesn't support recurring payments.
```
thomasplevy, line 523: where's the bugfix?

includes/models/model.llms.order.php
```diff
@@ -632,13 +641,37 @@ public function get_customer_name() {
...
 	/**
-	 * An array of default arguments to pass to $this->create()
-	 * when creating a new post
+	 * Retrieve the customer's full billing address
+	 *
+	 * @since [version]
+	 *
+	 * @return string
+	 */
+	public function get_customer_full_address() {
```
thomasplevy, line 650: We should get a test for this function and should probably also consider (later) where else addresses are output.

I know there's other places that addresses are output.

I'm also wondering if we should consider outputting these with newlines (for proper actual address formatting) and then there's the i18n consideration that this is the US address format and other countries have their own formats..

includes/models/model.llms.order.php
```diff
@@ -829,19 +868,54 @@ public function get_next_payment_due_date( $format = 'Y-m-d H:i:s' ) {
...
+		 * A timestamp should always be returned as the conversion to the requested format
+		 * will be performed on the returned value.
+		 *
+		 * @since [version]
+		 *
+		 * @param int        $upcoming_payment_reminder_time Unix timestamp for the next payment due date.
+		 * @param LLMS_Order $order                          Order object.
+		 * @param string     $format                         Requested date format.
+		 */
+		$upcoming_payment_reminder_time = apply_filters( 'llms_order_get_next_upcoming_payment_reminder_date', strtotime( '-1 day', strtotime( $next_payment_date ) ), $this, $format );
```
thomasplevy: I think it would make sense to add an additional filter allowing customization of the number of days directly. With this filter users will have to calculate the date themselves if they want to send the email 3 days before payment instead of 1.

We should leave this filter but I'm thinking just add another filter like `llms_order_payment_reminder_days`, default it to 1 and then users can quickly change 1 to 3 or 5 or 50.

This line can then pass in a variable (always converted to a negative number) instead of hardcoded 1.

Do you agree?

I also think that this is very likely something that many users will want to customize. 1 day maybe makes general sense but I actually think for me I'd want it to be 3 days.

I think we should add an option on the notification settings allowing end user customziation of this number without code

thomasplevy (commented):
Incomplete review, but some feedback and questions / thoughts

includes/class.llms.emails.php
eri-trabiccolo: Makes sense, but on the other hand we're going to always return an associative array when we already know that we'll use only the values...
So I will just drop the ARRAY_A case.

includes/class.llms.emails.php
eri-trabiccolo: Not my choice, I made a contest and that won by far.

includes/controllers/class.llms.controller.orders.php
eri-trabiccolo: @thomasplevy, I fixed this buggy line:
https://github.com/gocodebox/lifterlms/blob/4.17.0/includes/controllers/class.llms.controller.orders.php#L540
where `'recurring-payments'` and  `$order_id` are swapped.

includes/models/model.llms.order.php
eri-trabiccolo: @thomasplevy 
I've added a test for this function:
https://github.com/gocodebox/lifterlms/blob/e626c2bdaf3250d49b24129fb24f119357082ad9/tests/phpunit/unit-tests/models/class-llms-test-model-llms-order.php#L1266

Ok the only purpose for this was to reduce redundancy.
Add a todo here and open an issue or do you want to address it in this PR?
In this case I should document myself on i18n addresses :D

includes/models/model.llms.order.php
eri-trabiccolo: I thought that too, and I have noted as something to ask you when the PR was ready.
Long story short: I fully agree.

includes/class.llms.emails.php
thomasplevy: :+1:

includes/controllers/class.llms.controller.orders.php
thomasplevy: oh whoops :+1:

includes/models/model.llms.order.php
eri-trabiccolo: > I think we should add an option on the notification settings allowing end user customization of this number without code

On a second thought... this means that from the order model we're going to schedule an action at a time which depends on a notification's setting?

eri-trabiccolo: @thomasplevy ^

thomasplevy: Yes, that's what I'm thinking. The notification setting is something of a "global" setting that applies to all notifications of that type (regardless of order). I don't see a problem here. Do you see an issue that I'm not seeing or are you just clarifying?

eri-trabiccolo: Well no it's fine, it's just that I already wasn't very sure that the method to retrieve the upcoming payment reminder of an order had to be a method of the order, but yeah it's fine so.

thomasplevy:
@eri-trabiccolo I got a notification of a mention but I think the force push buried it. Or something. Is this ready for me to have a review of now?

eri-trabiccolo:
@thomasplevy 
Not ready yet. I've just moved the notification scheduling logic from the orders model to the notification controller as agreed. Still need to add the option to change the default number of days and then add tests.
I'm going to add you as reviewer when ready.

eri-trabiccolo:
@thomasplevy 
this is ready for an "initial" review.
Missing:
- ~more unit tests for the new methods~
- ~method for sending a test email too~
- I think the notification should be not enabled by default for old customers. We discussed a way to do this with the db updater... shall we?

includes/abstracts/llms.abstract.notification.controller.php
```diff
@@ -367,6 +367,18 @@ public function get_supported_types() {
...
+	/**
+	 * Get an array of additional options to be added to the notification view in the admin panel
+	 *
+	 * @since [version]
+	 *
+	 * @param string $type Type of the notification.
+	 * @return array
+	 */
+	public function get_additional_options( $type ) {
+		return apply_filters( 'llms_notification_' . $this->id . '_additional_options', $this->set_additional_options( $type ), $this, $type );
```
thomasplevy: + Filter needs docs
+ file header needs version update

tests/phpunit/unit-tests/notifications/controllers/class-llms-test-notification-controller-upcoming-payment-reminder.php
```diff
@@ -0,0 +1,198 @@
...
+		LLMS_Site::update_feature( 'recurring_payments', true );
+
+		// Check notification sent for existing order and student.
+		$this->assertTrue( $this->controller->action_callback( $order->get( 'id' ) ) );
+
+		// Check notification not sent for error gateway.
+		$order->set( 'payment_gateway', 'garbage' );
+		$this->assertFalse( $this->controller->action_callback( $order->get( 'id' ) ) );
+
+		// Check notification not sent for gateway that do not suppor recurring payments
```
thomasplevy:
```suggestion
		// Check notification not sent for gateway that do not support recurring payments
```

includes/notifications/controllers/class.llms.notification.controller.upcoming.payment.reminder.php
```diff
@@ -0,0 +1,426 @@
...
+	 * @since [version]
+	 *
+	 * @param string $type Type of the notification.
+	 * @return array
+	 */
+	protected function set_additional_options( $type ) {
+
+		return array(
+			array(
+				'id'                => $this->get_option_name( 'reminder_days' ),
```
thomasplevy: We should utilize the type passed into this so that as to allow the different notification types to have different options (like other existing options)

This also means we should have a helper method `get_reminder_days( $type )` and replace calls to `get_option( 'reminder_days' )` with that helper method that'll always pull the option for the given type

```suggestion
				'id'                => $this->get_option_name( $type . '_reminder_days' ),
```

includes/notifications/views/class.llms.notification.view.upcoming.payment.reminder.php
```diff
@@ -0,0 +1,263 @@
...
+	/**
+	 * Setup notification subject for output
+	 *
+	 * @since [version]
+	 *
+	 * @return string
+	 */
+	protected function set_subject() {
+		// Translators: %s = The product title.
+		return sprintf( __( 'You will be charged for your subscription to %1$s tomorrow', 'lifterlms' ), '{{PRODUCT_TITLE}}' );
```
thomasplevy:
```suggestion
		return sprintf( __( 'Upcoming payment reminder for your subscription to %1$s', 'lifterlms' ), '{{PRODUCT_TITLE}}' );
```

includes/notifications/views/class.llms.notification.view.upcoming.payment.reminder.php
```diff
@@ -0,0 +1,263 @@
...
+	 *
+	 * @return string
+	 */
+	protected function set_title() {
+		if ( 'email' === $this->notification->get( 'type' ) ) {
+			// Translators: %s = The order ID.
+			return sprintf( __( 'You will be charged for your subscription tomorrow. Order #%s', 'lifterlms' ), '{{ORDER_ID}}' );
+		}
+		// Translators: %s = The product title.
+		return sprintf( __( 'You will be charged for your subscription to %s', 'lifterlms' ), '{{PRODUCT_TITLE}}' );
```
thomasplevy:
```suggestion
		return __( 'Upcoming Subscription Payment', 'lifterlms' );
```

includes/notifications/views/class.llms.notification.view.upcoming.payment.reminder.php
```diff
@@ -0,0 +1,263 @@
...
+		<p>
+		<?php
+			// Translators: %s= The customer name.
+			printf( __( 'Hello %s,', 'lifterlms' ), '{{CUSTOMER_NAME}}' );
+		?>
+		</p>
+		<p>
+		<?php
+			// Translators: %1$s = The product title, %2$s The upcoming payment due date.
+			printf( __( 'You will be charged for your subscription to %1$s tomorrow on %2$s.', 'lifterlms' ), '{{PRODUCT_TITLE}}', '{{NEXT_PAYMENT_DATE}}' );
```
thomasplevy:
```suggestion
			printf( __( 'You will be charged for your subscription to %1$s on %2$s.', 'lifterlms' ), '{{PRODUCT_TITLE}}', '{{NEXT_PAYMENT_DATE}}' );
```

includes/notifications/views/class.llms.notification.view.upcoming.payment.reminder.php
```diff
@@ -0,0 +1,263 @@
...
+
+	/**
+	 * Setup default notification body for basic notifications
+	 *
+	 * @since [version]
+	 *
+	 * @return string
+	 */
+	private function set_body_basic() {
+		return esc_html__( 'You will be charged for your subscription tomorrow.', 'lifterlms' );
```
thomasplevy:
```suggestion
		return sprintf( esc_html__( 'You will be charged for your subscription to %1$s tomorrow.', 'lifterlms' ),  '{{PRODUCT_TITLE}}' );
```

thomasplevy (changes requested):
Some initial review. This looks very good but a few minor suggestions (mostly language which we can fight about if you like)

I'm still running manual tests but it looks like it should work right (save the change to the options I mention)

thomasplevy:
Manual testing looks stellar :+1:

includes/notifications/controllers/class.llms.notification.controller.upcoming.payment.reminder.php
eri-trabiccolo: Mmm alright, I actually thought to have one "shared" option on purpose... but makes sense and why not.

eri-trabiccolo: @thomasplevy but this means that we have to schedule two different actions...
This complicates the things a bit... do we really want this?

eri-trabiccolo: Alright I think the answer is yes we really want this.

eri-trabiccolo: @thomasplevy 
done in https://github.com/gocodebox/lifterlms/pull/1545/commits/0ee4aecc858af901641e24360c0db3e1d8783d52
needs more tests though - I haven't ran any manual test (except for the manual sending which doesn't imply the scheduling).

thomasplevy:
@eri-trabiccolo I think this looks pretty good, let me know when you're satisfied and I'll do more testing

eri-trabiccolo:
@thomasplevy I ran some additional tests. I think I'm satisfied.

eri-trabiccolo:
> @thomasplevy
> this is ready for an "initial" review.
> Missing:
> 
> * ~more unit tests for the new methods~
> * ~method for sending a test email too~
> * I think the notification should be not enabled by default for old customers. We discussed a way to do this with the db updater... shall we?

^
@thomasplevy about the defaults strategy then?

thomasplevy:
@eri-trabiccolo yes... we did discuss this.

I think that this should *be enabled by default* for new users but existing users should have to opt-in to it. This kind of a notification, I think, is probably one that will be polarizing. We've had users in the past complain that sending the receipt was problematic because users would request a refund and ask to cancel a subscription when they were reminded they had one...

My thinking was that we'd add a DB update/migration that would flip the options off during an upgrade. Based on how upgrades work a new user wouldn't have the upgrade run whereas an existing user would.

Sound good?

thomasplevy:
And by "disabled" i mean there shouldn't be any subscribers set

thomasplevy (approved):
Everything seems pretty good here to me, I approve

eri-trabiccolo:
> @eri-trabiccolo yes... we did discuss this.
> 
> I think that this should _be enabled by default_ for new users but existing users should have to opt-in to it. This kind of a notification, I think, is probably one that will be polarizing. We've had users in the past complain that sending the receipt was problematic because users would request a refund and ask to cancel a subscription when they were reminded they had one...
> 
> My thinking was that we'd add a DB update/migration that would flip the options off during an upgrade. Based on how upgrades work a new user wouldn't have the upgrade run whereas an existing user would.
> 
> Sound good?

Yup, we discussed about this and I agree.
I'm going to wait for the kind of db update we discussed to be ready then...

eri-trabiccolo:
@thomasplevy done. <3

includes/class.llms.install.php
```diff
@@ -108,6 +108,10 @@ class LLMS_Install {
 			'llms_update_500_update_db_version',
 			'llms_update_500_add_admin_notice',
 		),
+		'5.1.4'  => array(
```
thomasplevy:
```suggestion
		'5.2.0'  => array(
```

includes/class.llms.install.php
```diff
@@ -108,6 +108,10 @@ class LLMS_Install {
...
+			'llms_update_514_upcoming_reminder_notification_backward_compat',
```
thomasplevy:
```suggestion
			'llms_update_520_upcoming_reminder_notification_backward_compat',
```

includes/class.llms.install.php
```diff
@@ -108,6 +108,10 @@ class LLMS_Install {
...
+			'llms_update_514_update_db_version',
```
thomasplevy:
```suggestion
			'llms_update_520_update_db_version',
```

includes/functions/llms.functions.updates.php
```diff
@@ -25,6 +25,7 @@
 require_once 'updates/llms-functions-updates-450.php';
 require_once 'updates/llms-functions-updates-4150.php';
 require_once 'updates/llms-functions-updates-500.php';
+require_once 'updates/llms-functions-updates-514.php';
```
thomasplevy:
```suggestion
require_once 'updates/llms-functions-updates-520.php';
```

includes/functions/updates/llms-functions-updates-514.php
```diff
@@ -0,0 +1,50 @@
+<?php
+/**
+ * Update functions for version 5.1.4
```
thomasplevy:
```suggestion
 * Update functions for version 5.2.0
```

includes/functions/updates/llms-functions-updates-514.php
```diff
@@ -0,0 +1,50 @@
...
+ * @since [version]
+ * @version [version]
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Turn off autoload for accounting legacy options
+ *
+ * @since 5.1.4
```
thomasplevy:
```suggestion
 * @since [version]
```

includes/functions/updates/llms-functions-updates-514.php
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @return bool True if it needs to run again, false otherwise.
+ */
+function llms_update_514_upcoming_reminder_notification_backward_compat() {
```
thomasplevy:
```suggestion
function llms_update_520_upcoming_reminder_notification_backward_compat() {
```

includes/functions/updates/llms-functions-updates-514.php
```diff
@@ -0,0 +1,50 @@
...
+ * @package LifterLMS/Functions/Updates
+ *
+ * @since [version]
+ * @version [version]
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Turn off autoload for accounting legacy options
```
thomasplevy:
```suggestion
 * Explicitly set no subscribers for the new upcoming payment reminder notification
```

includes/functions/updates/llms-functions-updates-514.php
```diff
@@ -0,0 +1,50 @@
...
+	foreach ( $subscribers_for_type as $type => $subscribers ) {
+		add_option( "llms_notification_upcoming_payment_reminder_{$type}_subscribers", array_fill_keys( $subscribers, 'no' ) );
+	}
+
+	return false;
+
+}
+
+/**
+ * Update db version to 5.1.4
```
thomasplevy:
```suggestion
 * Update db version to 5.2.0
```

includes/functions/updates/llms-functions-updates-514.php
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @since [version]
+ *
+ * @return void|true True if it needs to run again, nothing if otherwise.
+ */
+function llms_update_514_update_db_version() {
```
thomasplevy:
```suggestion
function llms_update_520_update_db_version() {
```

includes/functions/updates/llms-functions-updates-514.php
```diff
@@ -0,0 +1,50 @@
...
+	LLMS_Install::update_db_version( '5.1.4' );
```
thomasplevy:
```suggestion
	LLMS_Install::update_db_version( '5.2.0' );
```

includes/functions/updates/llms-functions-updates-514.php
```diff
@@ -0,0 +1,50 @@
+<?php
```
thomasplevy: update filename to llms-functions-updates-520.php

tests/phpunit/unit-tests/functions/updates/class-llms-test-functions-updates-514.php
```diff
@@ -0,0 +1,101 @@
+<?php
+/**
+* Test updates functions when updating to 5.1.4
```
thomasplevy:
```suggestion
* Test updates functions when updating to 5.2.0
```

tests/phpunit/unit-tests/functions/updates/class-llms-test-functions-updates-514.php
```diff
@@ -0,0 +1,101 @@
+<?php
```
thomasplevy: update filename

tests/phpunit/unit-tests/functions/updates/class-llms-test-functions-updates-514.php
```diff
@@ -0,0 +1,101 @@
...
+/**
+* Test updates functions when updating to 5.1.4
+ *
+ * @package LifterLMS/Tests/Functions/Updates
+ *
+ * @group functions
+ * @group updates
+ * @group updates_514
```
thomasplevy:
```suggestion
 * @group updates_520
```

tests/phpunit/unit-tests/functions/updates/class-llms-test-functions-updates-514.php
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * @since [version]
+ * @version [version]
+ */
+class LLMS_Test_Functions_Updates_514 extends LLMS_UnitTestCase {
```
thomasplevy:
```suggestion
class LLMS_Test_Functions_Updates_520 extends LLMS_UnitTestCase {
```

tests/phpunit/unit-tests/functions/updates/class-llms-test-functions-updates-514.php
```diff
@@ -0,0 +1,101 @@
...
+	 *
+	 * Include update functions file.
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public static function setupBeforeClass() {
+		parent::setupBeforeClass();
+		require_once LLMS_PLUGIN_DIR . 'includes/functions/updates/llms-functions-updates-514.php';
```
thomasplevy:
```suggestion
		require_once LLMS_PLUGIN_DIR . 'includes/functions/updates/llms-functions-updates-520.php';
```

tests/phpunit/unit-tests/functions/updates/class-llms-test-functions-updates-514.php
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	/**
+	 * Test llms_update_514_upcoming_reminder_notification_backward_compat() method
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function test_llms_update_514_upcoming_reminder_notification_backward_compat() {
```
thomasplevy:
```suggestion
	public function test_llms_update_520_upcoming_reminder_notification_backward_compat() {
```

tests/phpunit/unit-tests/functions/updates/class-llms-test-functions-updates-514.php
```diff
@@ -0,0 +1,101 @@
...
+	 *
+	 * @return void
+	 */
+	public static function setupBeforeClass() {
+		parent::setupBeforeClass();
+		require_once LLMS_PLUGIN_DIR . 'includes/functions/updates/llms-functions-updates-514.php';
+	}
+
+	/**
+	 * Test llms_update_514_upcoming_reminder_notification_backward_compat() method
```
thomasplevy:
```suggestion
	 * Test llms_update_520_upcoming_reminder_notification_backward_compat() method
```

tests/phpunit/unit-tests/functions/updates/class-llms-test-functions-updates-514.php
```diff
@@ -0,0 +1,101 @@
...
+
+		foreach ( $subscribers_for_type as $type => $subscribers ) {
+			$this->assertEquals(
+				array(),
+				get_option( "llms_notification_upcoming_payment_reminder_{$type}_subscribers", array() )
+			);
+		}
+
+		// Run the update.
+		llms_update_514_upcoming_reminder_notification_backward_compat();
```
thomasplevy:
```suggestion
		llms_update_520_upcoming_reminder_notification_backward_compat();
```

tests/phpunit/unit-tests/functions/updates/class-llms-test-functions-updates-514.php
```diff
@@ -0,0 +1,101 @@
...
+			$this->assertNotEquals(
+				array_fill_keys( $subscribers, 'no' ),
+				get_option( "llms_notification_upcoming_payment_reminder_{$type}_subscribers", array() )
+			);
+		}
+
+	}
+
+	/**
+	 * Test llms_update_500_update_db_version()
```
thomasplevy:
```suggestion
	 * Test llms_update_520_update_db_version()
```

tests/phpunit/unit-tests/functions/updates/class-llms-test-functions-updates-514.php
```diff
@@ -0,0 +1,101 @@
...
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function test_update_db_version() {
```
thomasplevy:
```suggestion
	public function test_update_520_update_db_version() {
```

tests/phpunit/unit-tests/functions/updates/class-llms-test-functions-updates-514.php
```diff
@@ -0,0 +1,101 @@
...
+
+		$orig = get_option( 'lifterlms_db_version' );
+
+		// Remove existing db version.
+		delete_option( 'lifterlms_db_version' );
+
+		llms_update_514_update_db_version();
```
thomasplevy:
```suggestion
		llms_update_520_update_db_version();
```

tests/phpunit/unit-tests/functions/updates/class-llms-test-functions-updates-514.php
```diff
@@ -0,0 +1,101 @@
...
+
+		$this->assertEquals( '5.1.4', get_option( 'lifterlms_db_version' ) );
```
thomasplevy:
```suggestion
		$this->assertEquals( '5.2.0', get_option( 'lifterlms_db_version' ) );
```

thomasplevy (changes requested):
@eri-trabiccolo 

This looks great but I'm sorry this will be version 5.2.0 not 5.1.4 (this new feature constitutes a minor release)

I've noted, I think, all the places in the code that things need to change but filenames need to be updated too.

For future updates lets talk about a version number before you write the code... I recall us doing this at least once before today also :-(

eri-trabiccolo:
@thomasplevy ready to ship. <3

PR #1569, opened 2021-03-17
WIP: Fix default options when empty string is passed

### LINKED ISSUE #1568: LLMS_Abstract_Options_Data::get_option() issues with empty values
### Reproduction Steps

I can't explain this in a user story but I can explain it with the code found in this commit: https://github.com/gocodebox/lifterlms/commit/1e97aa7ce05b2ca0f9378350c0176f08f2d0c8cc

An option that is explicitly stored as an empty string `''` will cause a the passed default value of the option to be passed.

This isn't exactly a bug, but it is kind of a bug. We're leveraging this as "expected behavior" in certain places (see: https://github.com/gocodebox/lifterlms/commit/1e97aa7ce05b2ca0f9378350c0176f08f2d0c8cc#commitcomment-48354224) and in others it's creating bugs: https://github.com/gocodebox/lifterlms-integration-woocommerce/pull/102

### Expected Behavior

When an empty string is explicitly saved as the option value, it should be returned.

### Actual Behavior

+ The passed default value is returned instead

thomasplevy:
## Description

Fixes #1568

This adds an opt-in version of the abstract that concrete classes can implement in order to choose to use the new version of the `get_option()` method.

The new version will additionally automatically add a filter on `default_option_{$option}` (https://developer.wordpress.org/reference/hooks/default_option_option/) which will allow autoloading of options from integration settings (or other places), depending on the implementation provided by the extending class. 

## How has this been tested?

Unit tests only. I've written tests that expose the bug (and fix the bug) using stubs only.

This means that nothing will actually change until we start defining `$version=2` in some classes.

I've marked this as a WIP so we can discuss the problem better.

## Types of changes
Bugfix / forward compat

## Checklist:

### DIFF
--- a/includes/abstracts/llms.abstract.integration.php
+++ b/includes/abstracts/llms.abstract.integration.php
@@ -168,24 +168,27 @@ g	 */
 	 */
 	protected function get_settings() {
 
-		$settings[] = array(
-			'type'  => 'sectionstart',
-			'id'    => 'llms_integration_' . $this->id . '_start',
-			'class' => 'top',
+		$settings = array(
+			array(
+				'type'  => 'sectionstart',
+				'id'    => 'llms_integration_' . $this->id . '_start',
+				'class' => 'top',
+			);
+			array(
+				'desc'  => $this->description,
+				'id'    => 'llms_integration_' . $this->id . '_title',
+				'title' => $this->title,
+				'type'  => 'title',
+			);
+			array(
+				'desc'    => __( 'Check to enable this integration.', 'lifterlms' ),
+				'default' => 'no',
+				'id'      => $this->get_option_name( 'enabled' ),
+				'type'    => 'checkbox',
+				'title'   => __( 'Enable / Disable', 'lifterlms' ),
+			),
 		);
-		$settings[] = array(
-			'desc'  => $this->description,
-			'id'    => 'llms_integration_' . $this->id . '_title',
-			'title' => $this->title,
-			'type'  => 'title',
-		);
-		$settings[] = array(
-			'desc'    => __( 'Check to enable this integration.', 'lifterlms' ),
-			'default' => 'no',
-			'id'      => $this->get_option_name( 'enabled' ),
-			'type'    => 'checkbox',
-			'title'   => __( 'Enable / Disable', 'lifterlms' ),
-		);
+
 		if ( ! $this->is_installed() && ! empty( $this->description_missing ) ) {
 			$settings[] = array(
 				'id'    => 'llms_integration_' . $this->id . '_missing_requirements_desc',
--- a/includes/abstracts/llms.abstract.integration.php
+++ b/includes/abstracts/llms.abstract.integration.php
@@ -20,7 +20,7 @@ abstract class LLMS_Abstract_Integration extends LLMS_Abstract_Options_Data {
 	/**
 	 * Integration ID
 	 *
-	 * Defined by extending class as a variable
+	 * Defined by extending class as a variable.
 	 *
 	 * @var string
 	 */
@@ -61,8 +61,8 @@ abstract class LLMS_Abstract_Integration extends LLMS_Abstract_Options_Data {
 	 *
 	 * In the `configure()` method call `plugin_basename()` on the main plugin file.
 	 *
-	 * @var strin
-g	 */
+	 * @var string
+	 */
 	protected $plugin_basename = '';
 
 	/**
--- a/includes/abstracts/llms.abstract.integration.php
+++ b/includes/abstracts/llms.abstract.integration.php
@@ -61,8 +61,8 @@ abstract class LLMS_Abstract_Integration extends LLMS_Abstract_Options_Data {
 	 *
 	 * In the `configure()` method call `plugin_basename()` on the main plugin file.
 	 *
-	 * @var string
-	 */
+	 * @var strin
+g	 */
 	protected $plugin_basename = '';
 
 	/**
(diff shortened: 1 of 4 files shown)


includes/abstracts/llms.abstract.integration.php
```diff
@@ -14,62 +14,63 @@
...
- * @since 3.21.1 Updated.
- * @since 3.33.1 Added `get_priority` method to allow reading of the protected priority property.
- * @since 3.37.9 Added automatically generated "Settings" link to plugins screen.
  */
 abstract class LLMS_Abstract_Integration extends LLMS_Abstract_Options_Data {
 
 	/**
 	 * Integration ID
+	 *
 	 * Defined by extending class as a variable
```
eri-trabiccolo:
```suggestion
	 * Defined by extending class as a variable.
```

includes/abstracts/llms.abstract.integration.php
```diff
@@ -14,62 +14,63 @@
...
 	/**
 	 * Reference to the integration plugin's main plugin file basename
 	 *
-	 * In the configure() method call `plugin_basename()` on the main plugin file.
+	 * In the `configure()` method call `plugin_basename()` on the main plugin file.
 	 *
-	 * @var string
-	 */
+	 * @var strin
+g	 */
```
eri-trabiccolo:
```suggestion
	 * @var string
	 */
```

eri-trabiccolo (changes requested):
docs.
Quite fine for the rest, but it doesn't account for [this](https://github.com/gocodebox/lifterlms/pull/1569#commitcomment-48347493):

> In LifterLMS core you can see this bug occurring in the notifications:
> Try to set an empty body/subject/title, or even remove the notification icon (with no replacement), and save: the defaults will back.

thomasplevy:
I think it does take care of this because I don't believe any of these fields should actually be allowed to be blank, do you disagree?

If we upgrade the controller abstract to v2 it would still work the same way since we'd still pass the default value in, which for this set of options I think is okay.

eri-trabiccolo:
> I think it does take care of this because I don't believe any of these fields should actually be allowed to be blank, do you disagree?

Am I not even allowed to remove the notification icon?
What if I want a notification with only the title and no body because I'm a minimalist?

I can see what you mean when it's about emails though...

And maybe this has been used as a way to reset to defaults by some customer?!

> If we upgrade the controller abstract to v2 it would still work the same way since we'd still pass the default value in, which for this set of options I think is okay.

Not sure I get this. 
If we upgrade the notification view (and/or controller, my example actually involved the view) abstract then it will work with the new method (not the deprecated one). Because v2 and because by default the `get_option_default_value` callback returns the passed default value.
Are we saying the same thing?

thomasplevy:
> Am I not even allowed to remove the notification icon?

Mmm, okay. I see what you're saying. This makes sense. You should be able to remove the icon.

I think we should handle this as a separate issue since no one is currently complaining about this. I also don't know what the visual ramifications of not having an icon might be. By that I mean we may have to add some CSS and possibly even modify the HTML / PHP structure to accommodate an empty icon.

(I didn't read the code right now to check whether or not this is true though)

> And maybe this has been used as a way to reset to defaults by some customer?!

Yea, possibly. If you remove the icon and then want it back it would actually pretty hard to get the default one back since it's served from the plugin's asset directory and not the media library.

> > If we upgrade the controller abstract to v2 it would still work the same way since we'd still pass the default value in, which for this set of options I think is okay.
> 
> Not sure I get this.
> If we upgrade the notification view (and/or controller, my example actually involved the view) abstract then it will work with the new method (not the deprecated one). Because v2 and because by default the `get_option_default_value` callback returns the passed default value.
> Are we saying the same thing?

https://github.com/gocodebox/lifterlms/pull/1569/files#diff-b3fc2539fa0790dc6471ca24e53362e777aa155cf829420ff982f86203f21b34R245-R248

And

https://github.com/gocodebox/lifterlms/blob/f6a059fd0791a8fdcf8732593fd01cdeb8e81460/includes/abstracts/llms.abstract.notification.view.php#L464-L476

Basically what I'm saying is that on v2, if a default value is passed to the `get_option()` method it *disables* the autoloading automatically. 

So that means if the value is cleared (to an empty string) it will be "repopulated" with the default value automatically.



In both cases I think we should not worry about this right this second. We can enter this rabbit hole with notifications later. Right now I'm more worried about fixing a bug that's actually affecting a user in favor of getting lost discussing a theoretical bug (which I think you're correct is actually a bug) but no one is currently complaining about.

If someone really wants no notification icon we can give them some CSS to hide it while we figure out what to do about this.

eri-trabiccolo:
> 
> https://github.com/gocodebox/lifterlms/pull/1569/files#diff-b3fc2539fa0790dc6471ca24e53362e777aa155cf829420ff982f86203f21b34R245-R248
> 
> And
> 
> https://github.com/gocodebox/lifterlms/blob/f6a059fd0791a8fdcf8732593fd01cdeb8e81460/includes/abstracts/llms.abstract.notification.view.php#L464-L476
> 
> Basically what I'm saying is that on v2, if a default value is passed to the `get_option()` method it _disables_ the autoloading automatically.
> 
> So that means if the value is cleared (to an empty string) it will be "repopulated" with the default value automatically.
> 
> In both cases I think we should not worry about this right this second. We can enter this rabbit hole with notifications later. Right now I'm more worried about fixing a bug that's actually affecting a user in favor of getting lost discussing a theoretical bug (which I think you're correct is actually a bug) but no one is currently complaining about.
> 
> If someone really wants no notification icon we can give them some CSS to hide it while we figure out what to do about this.

I see, but you're missing that that callback is only called when the given option is not set, not just when it's empty.
So using v2 and passing a default, will ensure that the default is used only if the option is not set, but that it's not used when the option is set but it's empty - which is basically the native wp `get_option()` behavior

thomasplevy:
@eri-trabiccolo Can you come up with a better solution for us then?

eri-trabiccolo:
@thomasplevy
There's nothing to fix if we decide, as it seems, that this is what we want for the notifications.
I was just saying that v2 defines a different behavior even when passing defaults to the `get_option()` method. And, as we documented in the code, v2 will be the future standard behavior, so sooner or later we have to address this change.

PR #1576, opened 2021-03-24
WIP: Build a Fields API

thomasplevy:


includes/class-llms-form-field-data.php
```diff
@@ -0,0 +1,332 @@
...
+		$err = new WP_Error();
+		foreach ( $data as $key => &$value ) {
+
+			if ( ! self::validate( $key, $value ) ) {
+				$err->add(
+					sprintf( 'llms-form-field-invalid-data--%s', $key ),
+					sprintf( __( 'Invalid value submitted for %s', 'lifterlms' ), $key ),
+					compact( 'value' )
+				);
+			}
```
eri-trabiccolo, line 242:
```suggestion
				);
				continue;
			}
```

includes/class-llms-form-field-data.php
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	public function del_meta( $meta_key ) {
+		return $this->get_metas_obj( $meta_key )->delete();
+	}
+
+
+	private function get_metas_obj( $meta_key = null ) {
+		return new LLMS_Form_Field_Meta_Data( $this->get_id(), $meta_key );
+	}
```
eri-trabiccolo, line 324: ~is there a reason why we instantiate a new `LLMS_Form_Field_Meta_Data` each time we need to refer to it?
Can't we just do it in this constructor and then use a class member?~

eri-trabiccolo: forget about the constructor stuff. I just misread.
So mmm we can just actually _cache_ this by `$meta_key`.

PR #1579, opened 2021-03-29
Allow submitting numeric fields with no max/min attribute set

### LINKED ISSUE #1577: 5.0.0 alpha issues/feedback
Feedback / Issues from @eri-trabiccolo that we haven't investigate too thoroughly yet but we're recording here to avoid potential duplicates if any other tests stumble on these issues:

+ [x] 1) radio buttons in the editor look like checkboxes - so squared and with a small border radius, it’s because of our CSS that applies a small border radius to all the fields
+ [x] 2) checkboxes: it’s not possible to remove a default option once set (in the right panel)
+ [x] 3) checkboxes in the right panel look like radio buttons, and they shouldn’t, also you can check only one option by default, and this doesn’t make a lot of sense for checkboxes
+ [x] 4) a numeric field with no limits set doesn’t validate for positive numbers in front after submission, I got an error saying that the value should be less than 0 (sorry I think that was the error I haven’t copied :()
+ [x] 5) I’ve noticed in the DB that the post author is 0 for the default llms_form post types. I think it should be 1.
+ [x] 6) I’ve rapidly submitted a form with checkboxes checked, in the DB, though the checkbox value was empty.
+ [x] 7) When assigning a location to a form, in the “location modal” there are, supposedly, links to the selected product, with the “external link”(open in a new tab) icon, although clicking nothing happens.
+ [x] 8) When you Show keys of the default two options for both radio and checkboxes the "saved values" are empty.
+ [x] 9) Multiple defaults for checkboxes (grouped) are not rendered, only the latest is.

I have to say that I noticed this and noted rapidly because I was actually looking for something else (how the data was saved, retrieved etc) so I might have missed something e.g for the points 4 and 6

eri-trabiccolo:
## Description

Fixes #1577|4

> a numeric field with no limits set doesn’t validate for positive numbers in front after submission, I got an error saying that the value should be less than 0 (sorry I think that was the error I haven’t copied :()

## How has this been tested?
manually

## Types of changes

## Checklist:

### DIFF
--- a/includes/class-llms-form-validator.php
+++ b/includes/class-llms-form-validator.php
@@ -189,10 +189,10 @@ class LLMS_Form_Validator {
 			// Translators: %1$s field label or name; %2$s = user submitted value.
 			return new WP_Error( 'llms-form-field-invalid', sprintf( __( 'The %1$s "%2$s" is not valid number.', 'lifterlms' ), isset( $field['label'] ) ? $field['label'] : $field['name'], $posted_value ) );
 		} elseif ( isset( $field['attributes'] ) ) {
-			if ( ! empty( $field['attributes']['min'] ) || ( isset( $field['attributes']['min'] ) && '0' === $field['attributes']['min'] ) && $temp_value < $field['attributes']['min'] ) {
+			if ( ( ! empty( $field['attributes']['min'] ) || ( isset( $field['attributes']['min'] ) && '0' === $field['attributes']['min'] ) ) && $temp_value < $field['attributes']['min'] ) {
 				// Translators: %1$s field label or name; %2$s = user submitted value; %3$d = minimum allowed number.
 				return new WP_Error( 'llms-form-field-invalid', sprintf( __( 'The %1$s "%2$s" must be greater than or equal to %3$d.', 'lifterlms' ), isset( $field['label'] ) ? $field['label'] : $field['name'], $posted_value, $field['attributes']['min'] ) );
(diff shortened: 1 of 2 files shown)


thomasplevy (changes requested):
I don't think this works right.

It does seem to solve the issue you identified and even though it does pass [these tests](https://github.com/gocodebox/lifterlms/blob/5b37a2883d29c82d8f5d53abf24c9d1ca3b141c5/tests/phpunit/unit-tests/class-llms-test-form-validator.php#L564-L573) it's not working quite right.

If you dump out the errors the wrong error is passed for these. The first bank of tests for numbers that are smaller than the minimum work fine but the second bank (too large) results in messages like this: "The number_field "1,000" must be greater than or equal to 25."

The tests can be improved to check the error message to see if it's too big or too small (as expected) and then this bug needs to be addressed differently.

I'd like to see additional assertions or a new test case added to cover the case identified by the bug too

eri-trabiccolo:
@thomasplevy 
going to look into this and add new tests, although since I left the original checks on the posted data and I only changed the checks on the field settings, I would say that the issue was already there?! It might be just that I'm not lucid enough now :D

thomasplevy:
Just updated the tests to check for the expected message: https://github.com/gocodebox/lifterlms/commit/31ec86eeca3e6f6169a80a90e09bc44895d38e59

On dev-500 the proper error message is displaying

The problem with this set of tests is that the first set of tests (testing for the default values of the form) don't have min/max values set whereas when a number field is actually created it *will have empty* default values. If you update the $field to have min/max of an empty string you'll see the error you've reported occur within the tests.

The validation should treat an empty string value as "no value set" and the field not actually existing (as if it was created programmatically and left blank) the same way.

eri-trabiccolo:
matter of priority then... should be fine now. is it?

PR #1581, opened 2021-03-29
Handle fields that allow multiple values (e.g. checkboxes)

eri-trabiccolo:
## Description
WIP, aims to fix https://github.com/gocodebox/lifterlms/issues/1577

> 6 - I’ve rapidly submitted a form with checkboxes checked, in the DB, though the checkbox value was empty.
> 9 - Multiple defaults for checkboxes (grouped) are not rendered, only the latest is

At the moment it already fixes 6)

## How has this been tested?
manually and existing (modified) unit tests

## Types of changes

## Checklist:

### DIFF
--- a/includes/class-llms-form-validator.php
+++ b/includes/class-llms-form-validator.php
@@ -57,7 +57,15 @@ class LLMS_Form_Validator {
 
 		$func = isset( $map[ $field['type'] ] ) ? $map[ $field['type'] ] : 'sanitize_text_field';
 
-		return trim( call_user_func( $func, $posted_value ) );
+		// Turn the submitted value into array, so to unify validation of scalar and array posted values.
+		$to_sanitize = is_array( $posted_value ) ? $posted_value : array( $posted_value );
+		$sanitized   = array();
+
+		foreach ( $to_sanitize as $value ) {
+			$sanitized[] = trim( call_user_func( $func, $value ) );
+		}
+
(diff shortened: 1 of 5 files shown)


thomasplevy (commented):
I think this is generally okay although I don't think we need to delete the meta additional logic.

If the value isn't set, add an empty value, in this case an empty array, to the prepared data and let the update "clear" the existing data

eri-trabiccolo:
@thomasplevy working version pushed.
I think you would love new unit tests, isn't it?

includes/class-llms-form-validator.php
```diff
@@ -57,7 +57,15 @@ public function sanitize_field( $posted_value, $field ) {
 
 		$func = isset( $map[ $field['type'] ] ) ? $map[ $field['type'] ] : 'sanitize_text_field';
 
-		return trim( call_user_func( $func, $posted_value ) );
+		// Turn the submitted value into array, so to unify validation of scalar and array posted values.
```
thomasplevy: beautiful golf, I saw this the other day and thought "WE NEED TO GOLF THIS UPDATE" and you did it!

thomasplevy:
> I think you would love new unit tests, isn't it?

Yes

This looks pretty good otherwise. I didn't manually test this but I feel we need a few new tests to test checkbox data (as we currently don't have any or at least enough that would have caught this bug)

PR #1584, opened 2021-04-08
Introduce an util to retrieve the ID of the first available user who can install llms stuff

### LINKED ISSUE #1577: 5.0.0 alpha issues/feedback

eri-trabiccolo:
and use it as post_author for the default form

## Description

Fixes #1577|5
5) I’ve noticed in the DB that the post author is 0 for the default llms_form post types. I think it should be 1.
as we discussed in the comments there below

## How has this been tested?
only manually as of now

## Types of changes

## Checklist:

@thomasplevy do you agree with this?

I mean, if you don't have an administrator user or your administrator doesn't have the manage_options cap, the forms author will be 0, we can't search forever for a valid user... right?

### DIFF
--- a/tests/phpunit/unit-tests/class-llms-test-forms.php
+++ b/tests/phpunit/unit-tests/class-llms-test-forms.php
@@ -385,6 +385,95 @@ class LLMS_Test_Forms extends LLMS_UnitTestCase {
 	}
 
 	/**
+	 * Test forms author on install
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function test_forms_author_on_install() {
+
+		// Clean user* tables.
+		global $wpdb;
+		$wpdb->query( "TRUNCATE TABLE $wpdb->users" );
+		$wpdb->query( "TRUNCATE TABLE $wpdb->usermeta" );
+
+		// Create a subscriber.
+		$subscriber = $this->factory->user->create( array( 'role' => 'subscriber' ) );
+
+		$locs = $this->forms->get_locations();
+
+		// Install forms
+		$installed = $this->forms->install();
+
+		foreach ( $installed as $loc => $id ) {
+			// No admin users, expect 0.
+			$this->assertEquals( 0, get_post( $id )->post_author, $id );
+		}
+
+		// Delete forms.
+		$wpdb->delete( $wpdb->posts, array( 'post_type' => 'llms_form' ) );
+
+		// Create two admins.
+		$admins = $this->factory->user->create_many( 2, array( 'role' => 'administrator' ) );
+
+		// Install forms.
+		$installed = $this->forms->install();
+
+		foreach ( $installed as $loc => $id ) {
+			// Expect the first admin to be the forms author.
+			$this->assertEquals( $admins[0], get_post( $id )->post_author, $id );
+		}
+
+		// Delete forms.
+		$wpdb->delete( $wpdb->posts, array( 'post_type' => 'llms_form' ) );
+
+		// Log in as subscriber.
+		wp_set_current_user( $subscriber );
+
+		// Install forms.
+		$installed = $this->forms->install();
+
+		foreach ( $installed as $loc => $id ) {
+			// Expect the first admin to be the forms author.
+			$this->assertEquals( $admins[0], get_post( $id )->post_author, $id );
+		}
+
+		// Delete forms.
+		$wpdb->delete( $wpdb->posts, array( 'post_type' => 'llms_form' ) );
+
+		// Log in as first admin.
+		wp_set_current_user( $admins[0] );
+
+		// Install forms.
+		$installed = $this->forms->install();
+
+		foreach ( $installed as $loc => $id ) {
+			// Expect the first admin to be the forms author.
+			$this->assertEquals( $admins[0], get_post( $id )->post_author, $id );
+		}
+
+		// Delete forms.
+		$wpdb->delete( $wpdb->posts, array( 'post_type' => 'llms_form' ) );
+
+		// Log in as second admin.
+		wp_set_current_user( $admins[1] );
+
+		// Install forms.
+		$installed = $this->forms->install();
+
+		foreach ( $installed as $loc => $id ) {
+			// Expect the first admin to be the forms author.
+			$this->assertEquals( $admins[1], get_post( $id )->post_author, $id );
+		}
+
+		// Clean user* tables.
+		$wpdb->query( "TRUNCATE TABLE $wpdb->users" );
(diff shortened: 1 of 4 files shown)


thomasplevy:
@eri-trabiccolo I see no issues with this implementation. Tests and, of course, usage of the function to fix the issue, and I think we're good to go :+1:

eri-trabiccolo:
tests added

on a _different_ subject
I'm not sure why the `create` method allows the update... and why in that case we don't use `wp_update_post`

thomasplevy:
> on a _different_ subject
> I'm not sure why the `create` method allows the update... and why in that case we don't use `wp_update_post`

`wp_update_post()` uses `wp_insert_post()` under the hook, with some additional logic. By using the same function, instead of conditionally using one or the other, we can reduce logic.

The method allows an update so that we can update. I believe I added "update" logic after initially writing the creation method and chose to use the same method in favor of creating a different one because the logic is *nearly* identical. I created the update option so that I could programatically re-install the forms (over top of the existing ones) during development as I rewrote and adjusted the templates.

If the problem is semantic usage of language, I think it's best to think of the `$update` flag as a "recreate" flag, in favor of an "update the form" flag. Notice that you cannot *pass in any updates*, you simply pass in that you *want* to update it. It will then update the existing form post with the template for the requested location.

The most common public use case I can see for this would be developing a "reset" function that could maybe be accessed via an action button on the posts table list or maybe from the admin "tools" ui.

I don't see any issue with using `wp_insert_post()` (when passing an ID) for both creation and update. If you see a specific issue (other than not understanding my reasoning) let me know.

tests/phpunit/unit-tests/class-llms-test-forms.php
```diff
@@ -384,6 +384,95 @@ public function test_create() {
...
+
+		foreach ( $installed as $loc => $id ) {
+			// Expect the first admin to be the forms author.
+			$this->assertEquals( $admins[1], get_post( $id )->post_author, $id );
+		}
+
+		// Clean user* tables.
+		$wpdb->query( "TRUNCATE TABLE $wpdb->users" );
+		$wpdb->query( "TRUNCATE TABLE $wpdb->usermeta" );
+
```
thomasplevy:
```suggestion
```

We don't need to clean up the test users we created for this test. This won't hurt anything but I'm almost positive it's unnecessary

tests/phpunit/unit-tests/class-llms-test-install.php
```diff
@@ -316,4 +327,57 @@ public function test_install() {
...
+		// Log in as second admin.
+		wp_set_current_user( $admins[1] );
+
+		// Expect the second admin to be returned.
+		$this->assertEquals( $admins[1], LLMS_Install::get_can_install_user_id() );
+
+		// Clean user* tables.
+		$wpdb->query( "TRUNCATE TABLE $wpdb->users" );
+		$wpdb->query( "TRUNCATE TABLE $wpdb->usermeta" );
+
```
thomasplevy:
```suggestion
```

Like above, pretty sure we don't need this cleanup

thomasplevy (approved):

eri-trabiccolo:
> > on a _different_ subject
> > I'm not sure why the `create` method allows the update... and why in that case we don't use `wp_update_post`
> 
> `wp_update_post()` uses `wp_insert_post()` under the hook, with some additional logic. By using the same function, instead of conditionally using one or the other, we can reduce logic.
> 
> The method allows an update so that we can update. I believe I added "update" logic after initially writing the creation method and chose to use the same method in favor of creating a different one because the logic is _nearly_ identical. I created the update option so that I could programatically re-install the forms (over top of the existing ones) during development as I rewrote and adjusted the templates.
> 
> If the problem is semantic usage of language, I think it's best to think of the `$update` flag as a "recreate" flag, in favor of an "update the form" flag. Notice that you cannot _pass in any updates_, you simply pass in that you _want_ to update it. It will then update the existing form post with the template for the requested location.
> 
> The most common public use case I can see for this would be developing a "reset" function that could maybe be accessed via an action button on the posts table list or maybe from the admin "tools" ui.
> 
> I don't see any issue with using `wp_insert_post()` (when passing an ID) for both creation and update. If you see a specific issue (other than not understanding my reasoning) let me know.

I'm fine with this. Maybe 'recreate' or 'reset' or 'force' would be semantically better? Maybe.
Anyways yeah I know wp_update_post calls wp_insert_post under the hood 😜 + some _minor_ additional logic, I wanted to understand why we decided we didn't need that additional logic. Ok now I get it, and I'm fine with this.

tests/phpunit/unit-tests/class-llms-test-forms.php
eri-trabiccolo: I see what you mean, as of now we wouldn't need that clean up... But I would like to have our tests to not leave anything in the db that could arm another test. If any test would take care of the clean up before quitting (and we generally do that in the `tearDown()`) we wouldn't bore about cleaning at the beginning of a test (when necessary) nor we would spend time on debugging a test that doesn't work as expected (as a result of "environment" changes produced by a previous test).
Am I wrong/missing something obvious/being too much anxious? 😅

PR #1585, opened 2021-04-09
(5.0.*) When populating field data from the INPUT_POST make sure checkboxes are treated as array

eri-trabiccolo:
## Description
Fixes an issue I haven't opened. Basically if a form gets a validation error server side, the checkbox values in the form are not correctly set, as they're not treated as arrays.
(A way to reproduce the validation error would be to submit a password change on the account edit form, using a wrong current password.)

## How has this been tested?
manually

## Types of changes

## Checklist:


eri-trabiccolo:
@thomasplevy I'm in a rush and wanted to submit this before I leave.
If you don't need to merge it today I'll write tests on monday...

PR #1587, opened 2021-04-13
Allow for improved user login validation with new filter

paulgoodchild:
Current login form validation involves modifying the actual form submission data before sending it to be validated.  There is no filter applied to the result of the validation.

Instead of modifying existing filters which is likely to break other integrations, I opted to create a new filter

 Re-using either of the filter before/after this new filter could cause unexpected behaviour with other integrations if they're relying on the data to be in a certain format.

In this new way we can now integrate with the login flow more thoroughly by supplying a `WP_Error` during the login process and triggering an exit of the login quite early, if necessary.

## Description

Neither of the 2 pre-existing filters allows for full login integration.

The filter immediately preceeding the new filter is applied to the data being validated, not the result of the validation. This is probably a bug and (I think) should probably behave more like the filter `lifterlms_user_registration_data` used in user registration.

The filter immediately following our new filter is only called when there is *already* a validation error, preventing the creation of a new error.

## How has this been tested?
This is simply a brand new filter that doesn't affect any existing flows or modifies any data. Tested on test LifterLMS installation.

## Types of changes

New Feature / Non-Breaking Change
A new filter to allow improved 3rd party integration with user login to (in)validate login as required. The existing code structure doesn't fully allow for integration (this is probably a bug)

## Checklist:

### DIFF
--- a/includes/class.llms.person.handler.php
+++ b/includes/class.llms.person.handler.php
@@ -676,6 +676,7 @@ class LLMS_Person_Handler {
 		// Validate the fields & allow custom validation to occur.
 		$valid = self::validate_fields( apply_filters( 'lifterlms_user_login_data', $data ), 'login' );
 
+		$valid = apply_filters( 'lifterlms_pre_user_login_errors', $valid, $data, false );
 		// If errors found, return them.
 		if ( is_wp_error( $valid ) ) {
 			return apply_filters( 'lifterlms_user_login_errors', $valid, $data, false );
--- a/includes/class.llms.person.handler.php
+++ b/includes/class.llms.person.handler.php
@@ -676,8 +676,15 @@ class LLMS_Person_Handler {
 		// Validate the fields & allow custom validation to occur.
 		$valid = self::validate_fields( apply_filters( 'lifterlms_user_login_data', $data ), 'login' );
 
-		// Allow for custom validation of user login data (previous filter doesn't fully allow for this)
-		$valid = apply_filters( 'llms_after_user_login_data_validation', $valid, $data, 'login' );
+		/**
+		 * Filters the validation result of user-submitted login data
+		 * 
+		 * @since [version]
+		 *
+		 * @param WP_Error|boolean $valid An error object containing validation errors or `true` if no validation errors found.
+		 * @param array            $data  User submitted login data.
+		 */
+		$valid = apply_filters( 'llms_after_user_login_data_validation', $valid, $data );
 
 		// If errors found, return them.
 		if ( is_wp_error( $valid ) ) {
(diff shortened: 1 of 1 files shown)


includes/class.llms.person.handler.php
```diff
@@ -676,6 +676,9 @@ public static function login( $data ) {
 		// Validate the fields & allow custom validation to occur.
 		$valid = self::validate_fields( apply_filters( 'lifterlms_user_login_data', $data ), 'login' );
 
+		// Allow for custom validation of user login data (previous filter doesn't fully allow for this)
+		$valid = apply_filters( 'llms_after_user_login_data_validation', $valid, $data, 'login' );
```
thomasplevy: Docblocks & dropped the 3rd parameter that isn't really necessary (though it does conform with other similar filters)

```suggestion
		/**
		 * Filters the validation result of user-submitted login data
		 * 
		 * @since [version]
		 *
		 * @param WP_Error|boolean $valid An error object containing validation errors or `true` if no validation errors found.
		 * @param array            $data  User submitted login data.
		 */
		$valid = apply_filters( 'llms_after_user_login_data_validation', $valid, $data );
```

thomasplevy (commented):
@paulgoodchild thanks for your interest in LifterLMS

We're on the tail end of alpha testing on LifterLMS 5.0. The primary goal of 5.0 is rebuilding (in a lot of areas completely rebuilding) user login/registration forms and handlers. This change is welcome but I'm going to move it around and change it a bit because the login handling in 5.0 is different and I've already locked the deprecations for 5.0 so I don't want to add any new features today that will be changed (in this case very slightly (see my suggestion)) when 5.0 comes out (probably within the next 30-45 days, assuming beta testing goes well).

On 5.x I think this filter belongs on the return of `LLMS_Person_Handler::validate_login_fields()`:

https://github.com/gocodebox/lifterlms/blob/762778c1d6e4c5a04f462be9c0e16fde16d361bd/includes/class.llms.person.handler.php#L502

I think this should fulfill your requirements and if it all seems good to you we'll get this merged in for our next release and I'll move it to the more appropriate location on the dev-500 branch before moving that to beta. There won't be a change in behavior with the move.

paulgoodchild:
@thomasplevy Sounds great! I'm delighted you'll have this pushed through for the next release - we can complete the integration quite quickly then.

And yep, if the "location" changes for v5.x but the functionaliy doesn't then that's perfect, too.

Thanks for taking a look so quickly!

thomasplevy:
@paulgoodchild would you like to update your branch with the suggestions I've made?

paulgoodchild:
> @paulgoodchild would you like to update your branch with the suggestions I've made?

All done... hadn't realised you needed me to merge it to my branch.  I've built our integration on this, so we're good to once this is merged into your next release.

Cheers!

PR #1595, opened 2021-04-15
Certificates export: embed/inline allowed external assets

### LINKED ISSUE #1509: Missing stylesheets in saved certificates, when serving them (stylesheets) trough a CDN.
### Reproduction Steps
Edit:

What I said below, while it's technically true, doesn't take into account the fact that, because of this line:
https://github.com/gocodebox/lifterlms/blob/4.13.0/includes/class.llms.certificates.php#L263-L265
we totally skip external stylesheets, and remove them later:
https://github.com/gocodebox/lifterlms/blob/4.13.0/includes/class.llms.certificates.php#L285-L289

~~This is more a guess based on the code and on something I could actually reproduce.
Basically I theorized this when handling the ticket HS-151532 
I suggested to disable the CDN and that very issue seemed to be resolved, so I'm quite sure my guess was good.
Also because I verified that retrieving one of those stylesheets hosted on the CDN with `file_get_content()` on my machine - with `fopen_wrappers` disabled - was working fine, so it _shouldn't_ be a problem related to the CDN settings/firewall. All these are guesses at this stage.~~

~~So let's go with reproduction steps.
(implies [fopen_wrappers](https://www.php.net/manual/en/filesystem.configuration.php#ini.allow-url-fopen) are disabled).~~

+ Use CDN for your stylesheets :D
or
+ for testing purposes, you can just enqueue some external stylesheet
+ save a certificate 
+ Inspect the html

### Expected Behavior

+ I'd expect the style to be correctly inlined

### Actual Behavior

+ You'll not see the "external" stylesheet enqueued no

The problem is that I think we "assume" that all the stylesheets are local here:
https://github.com/gocodebox/lifterlms/blob/4.13.0/includes/class.llms.certificates.php#L263-L265

Also we need to keep in mind that if we just avoid that skip we can still have problems retrieving external stylesheets via `file_get_contents()`.
https://github.com/gocodebox/lifterlms/blob/4.13.0/includes/class.llms.certificates.php#L268-L269

We know that we just can't assume `file_get_contents()` will work when using URLs, so I think that when we get an external URL we should go with curl (or `wp_remote_get()`).

The idea is to keep skipping some external sylesheets: e.g. fonts cdn (with a filterable blocklist); but to retrieve remote ones.

### Error Messages / Logs

n/a

LiferLMS 4.13.0

</details>

### This issue has beeb recreated:
+ [x] Locally
+ [ ] On a staging site
+ [x] On a production website

n/a

### Related User Information

HS-151532

eri-trabiccolo:
## Description
Fixes #1509 

## How has this been tested?
manually and new unit tests

## Types of changes

## Checklist:


thomasplevy (commented):
This looks good

However, I'm wondering if we should the stylesheet and image functionality into submethods

For example:
`private function modify_dom_links( $dom ) { ... }`
`private function modify_dom_images( $dom ) { ... }`

Pass the `$dom` variable into these sub methods and perform the logic in there, it wouldn't have to return anything.

Making these smaller methods would make it a little easier to test the sub methods independently and would also improve the CC code quality metrics

eri-trabiccolo:
@thomasplevy 
ready.

Needed to added further images from unsplash:
https://unsplash.com/photos/rDMacl1FDjw
https://unsplash.com/photos/R7ftweJR8ks

because one of the two already available in the test assets was too big and broke the DOM so that the tests failed.

PR #1612, opened 2021-04-28
Added public getter to retrieve llms form fields from a list of wp blocks

eri-trabiccolo:
## Description
Need this to retrieve the list of fields in a reusable block's content, so that when the reusable block is saved we can manipulate the fields (e.g. update the "tracked fields" option).

## How has this been tested?
manually, it just consists in moving part of the logic of an already existing method into a new public one.

## Types of changes
n/a

## Checklist:


thomasplevy:
I wrote your tests for you <3

PR #1631, opened 2021-05-10
Admin profile fields

eri-trabiccolo:
## Description
wip

This PR aims to refactor the way we add llms core and custom fields to the user profile (admin).

## How has this been tested?
manually and existing unit tests

Misses new tests and deprecations.

## Types of changes

## Checklist:
- [x ] My code follows the LifterLMS Coding & Documentation Standards.

### DIFF
--- a/includes/admin/class-llms-admin-profile.php
+++ b/includes/admin/class-llms-admin-profile.php
@@ -13,7 +13,7 @@ defined( 'ABSPATH' ) || exit;
 /**
  * Handle extra profile fields for users in admin
  *
- * Applies to edit-user.php, & profile.php.
+ * Applies to edit-user.php & profile.php.
  *
  * @since [version]
  */
@@ -85,47 +85,6 @@ class LLMS_Admin_Profile {
 	}
 
 	/**
-	 * Undocumented function
-	 *
-	 * Only who can 'manage_lifterlms' will be able to edit own and other profiles.
-	 *
-	 * @param WP_User $user Instance of WP_User for the user being updated.
-	 * @return boolean
-	 */
-	private function current_user_can_manage_user_custom_fields( $user ) {
-
-		return current_user_can( 'manage_lifterlms' ) && current_user_can( 'edit_users', $user );
-
-	}
-
-	/**
-	 * Maybe print validation errors
-	 *
-	 * @since [version]
-	 *
-	 * @param WP_Error $errors Instance of WP_Error, passed by reference.
-	 * @return void
-	 */
-	public function add_errors( &$errors ) {
-
-		if ( is_wp_error( $this->errors ) && $this->errors->has_errors() ) {
-			foreach ( $this->errors->get_error_messages() as $message ) {
-				$errors->add(
-					'',
-					sprintf(
-						// Translators: %1$s = Opening strong tag; %2$s = Closing strong tag; %3$s = The error message.
-						esc_html__( '%1$sError%2$s: %3$s', 'lifterlms' ),
-						'<strong>',
-						'</strong>',
-						$message
-					)
-				);
-			}
-		}
-
-	}
-
-	/**
 	 * Maybe save customer meta fields
 	 *
 	 * @since [version]
@@ -168,6 +127,75 @@ class LLMS_Admin_Profile {
 	}
 
 	/**
+	 * Maybe print validation errors
+	 *
+	 * @since [version]
+	 *
+	 * @param WP_Error $errors Instance of WP_Error, passed by reference.
+	 * @return void
+	 */
+	public function add_errors( &$errors ) {
+
+		if ( is_wp_error( $this->errors ) && $this->errors->has_errors() ) {
+			$this->merge_llms_fields_errors( $errors );
+		}
+
+	}
+
+	/**
+	 * Check whether the current user can manage a given user's custom fields
+	 *
+	 * Only who can 'manage_lifterlms' will be able to edit own and other profiles.
+	 *
+	 * @since [version]
+	 *
+	 * @param WP_User $user Instance of WP_User for the user being updated.
+	 * @return boolean
+	 */
+	private function current_user_can_manage_user_custom_fields( $user ) {
+
+		return current_user_can( 'manage_lifterlms' ) && current_user_can( 'edit_users', $user );
+
+	}
+
+	/**
+	 * Merge llms fields errors into the passed WP_Error
+	 *
+	 * @since [version]
+	 * @todo Remove the fallback when minimum required WP version will be 5.6+.
+	 *
+	 * @param WP_Error $errors Instance of WP_Error, passed by reference.
+	 * @return void
+	 */
+	private function merge_llms_fields_errors( &$errors ) {
+
+		foreach ( $this->errors->get_error_codes() as $code ) {
+			foreach ( $this->errors->get_error_messages( $code ) as $error_message ) {
+				$errors->add(
+					$code,
+					sprintf(
+						// Translators: %1$s = Opening strong tag; %2$s = Closing strong tag; %3$s = The error message.
+						esc_html__( '%1$sError%2$s: %3$s', 'lifterlms' ),
+						'<strong>',
+						'</strong>',
+						$error_message
+					)
+				);
+			}
+			/**
+			 * `WP_Error::get_all_error_data()` has been introduced in WP 5.6.0.
+			 */
+			$error_data = method_exists( $this->errors, 'get_all_error_data' ) ?
+					$this->errors->get_all_error_data( $code ) : $this->errors->get_error_data( $code );
+
+			foreach ( $error_data as $data ) {
+				$errors->add_data( $data, $code );
+			}
+		}
+
+	}
+
+	/**
 	 * Get fields to be added in the profile screen
 	 *
 	 * @since [version]
@@ -198,7 +226,7 @@ class LLMS_Admin_Profile {
 					),
 					array(
 						'type'           => 'text',
-						'label'          => __( 'Additional address information', 'lifterlms' ), // It's used in the error messages.
+						'label'          => __( 'Address line 2', 'lifterlms' ), // It's used in the error messages.
 						'placeholder'    => __( 'Apartment, suite, etc...', 'lifterlms' ),
 						'name'           => 'llms_billing_address_2',
 						'id'             => 'llms_billing_address_2',
--- a/includes/admin/class-llms-admin-profile.php
+++ b/includes/admin/class-llms-admin-profile.php
@@ -73,6 +73,10 @@ class LLMS_Admin_Profile {
 		wp_enqueue_script( 'llms-metaboxes' );
 		wp_enqueue_script( 'llms-select2' );
 		llms()->assets->enqueue_style( 'llms-select2-styles' );
+		wp_add_inline_script(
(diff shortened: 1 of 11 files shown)


includes/admin/class-llms-admin-profile.php
```diff
@@ -0,0 +1,269 @@
...
+	 * Maybe print validation errors
+	 *
+	 * @since [version]
+	 *
+	 * @param obj     $errors Instance of WP_Error, passed by reference.
+	 * @param bool    $update `true` if updating a profile, `false` if a new user.
+	 * @param WP_User $user   Instance of WP_User for the user being updated.
+	 * @return void
+	 */
+	public function add_errors( &$errors, $update, $user ) {
```
thomasplevy: $update and $user args aren't used, they can be removed, I imagine?

includes/admin/class-llms-admin-profile.php
```diff
@@ -0,0 +1,269 @@
...
+
+		if ( is_wp_error( $this->errors ) && $this->errors->has_errors() ) {
+			foreach ( $this->errors->get_error_messages() as $message ) {
+				$errors->add(
+					'',
```
thomasplevy: We should add an error code here 

Ideally we'd probably want to just use this: https://developer.wordpress.org/reference/classes/wp_error/copy_errors/ but it wasn't added to 5.6

But the logic could be duplicated from that function so we have an error code. Mainly this will be useful for debugging and testing but I think it's still important to have an error code

includes/admin/class-llms-admin-profile.php
```diff
@@ -0,0 +1,269 @@
...
+		if ( ! isset( $this->fields ) ) {
+
+			$this->fields = apply_filters(
+				/**
+				 * Fields to be added in the profile screen
+				 *
+				 * @since [version]
+				 *
+				 * @param array $fields Array of fields.
+				 */
```
thomasplevy: Move this up above the `apply_filters()` call

includes/admin/class-llms-admin-profile.php
```diff
@@ -0,0 +1,269 @@
...
+
+		if ( ! isset( $this->fields ) ) {
+
+			$this->fields = apply_filters(
+				/**
+				 * Fields to be added in the profile screen
+				 *
+				 * @since [version]
+				 *
+				 * @param array $fields Array of fields.
```
thomasplevy:
```suggestion
				 * @param array[] $fields Array of fields.
```

includes/admin/class.llms.admin.user.custom.fields.php
```diff
@@ -104,72 +106,15 @@ public function add_errors( &$errors, $update, $user ) {
 	 *
 	 * @since 2.7.0
 	 * @since 3.13.0 Unknown.
+	 * @since [version] Removed LLMS core fields.
 	 *
 	 * @return array
 	 */
 	public function get_fields() {
 
 		$fields = apply_filters(
```
thomasplevy: Use `apply_filters_deprecated()` (with a filter doc)

includes/forms/class-llms-form-field.php
```diff
@@ -58,29 +58,57 @@ class LLMS_Form_Field {
...
+		if ( ! is_null( $data_source ) ) {
+			$data_source = $data_source instanceof WP_User ? $data_source : get_user_by( 'ID', $data_source );
+		} elseif ( is_user_logged_in() ) {
+			$data_source = wp_get_current_user();
+		}
+
+		$this->user = $data_source;
+
+	}
+
```
thomasplevy: I think I'd like to see `$this->user` converted to `$this->data_source`, it's a semantic change but I think it'll be better for us in the future when we add other data sources

includes/forms/class-llms-form-field.php
```diff
@@ -846,9 +874,8 @@ protected function prepare_value() {
 		}
 
 		// Auto-populate field from the datastore if we have a user and datastore information.
-		if ( is_null( $user_val ) && is_user_logged_in() && $this->settings['data_store_key'] && in_array( $this->settings['data_store'], array( 'users', 'usermeta' ), true ) ) {
-			$user     = wp_get_current_user();
-			$user_val = $user->get( $this->settings['data_store_key'] );
+		if ( is_null( $user_val ) && ( ! is_null( $this->user ) ) && $this->settings['data_store_key'] && in_array( $this->settings['data_store'], array( 'users', 'usermeta' ), true ) ) {
```
thomasplevy: I think the logic can be simplified here. In `define_data_source()` we validate the presence of a data store so here I think all we need to do is verify that a `data_store_key` exists for the field, right?

```suggestion
		if ( is_null( $user_val ) && ( ! is_null( $this->user ) ) && $this->settings['data_store_key'] ) {
```

includes/forms/class-llms-form-handler.php
```diff
@@ -338,12 +356,11 @@ public function submit( $posted_data, $location, $args = array() ) {
...
-		 * @param array $posted_data Array of user-submitted data (passed by reference).
-		 * @param string $location Form location.
-		 * @param array[] $fields Array of LifterLMS Form Fields (passed by reference).
-		 * @param array $args Additional arguments from the form retrieval function.
+		 * @param array   $posted_data Array of user-submitted data (passed by reference).
+		 * @param string  $location    Form location.
+		 * @param array[] $fields      Array of LifterLMS Form Fields (passed by reference).
 		 */
-		do_action_ref_array( "lifterlms_before_user_${action}", array( &$posted_data, $location, &$fields, $args ) );
+		do_action_ref_array( "lifterlms_before_user_${action}", array( &$posted_data, $location, &$fields ) );
```
thomasplevy, line 362: Why did the $args arg get removed from these actions/filters?

thomasplevy (changes requested):
I didn't run this I'm just reading code.

I think it looks quite good so far

includes/forms/class-llms-form-field.php
eri-trabiccolo: I don't agree because then we would need to add an additional check on the nature of `$data_source`, instead I want to be sure that $this->user is either unset or it's a WP_User.
Then we'll have `$this->post`.
What do you think? Am I stupid?

includes/forms/class-llms-form-field.php
eri-trabiccolo: What you say here makes totally sense, if `$this->user` stays as I said above :P

includes/forms/class-llms-form-handler.php
eri-trabiccolo: Because I don't want to pass `$args` to the `submit_fields()` method, because submit_fields knows nothing about the args, which per documentation are only used to short-circuit the filter of `LLMS_Forms::instance()->get_form_fields( $location, $args );` - but I forgot to remove it.
If we need to pass the $args to a filter then we need to add new filters here:
https://github.com/gocodebox/lifterlms/pull/1631/files/1b276509680466ae69fd4507f022100c71f8af5e#diff-bdd71f993bfe78b9a844459acbbfdf891da73dd0223f2ce1113b3769169c1003R330
and here:
https://github.com/gocodebox/lifterlms/pull/1631/files/1b276509680466ae69fd4507f022100c71f8af5e#diff-bdd71f993bfe78b9a844459acbbfdf891da73dd0223f2ce1113b3769169c1003R332
(and of course return after the filter).
But I don't have a way to justify that `$args` in `submit_fields()`.

includes/admin/class.llms.admin.user.custom.fields.php
eri-trabiccolo: we can't because we'll continue to use it here -> 
https://github.com/gocodebox/lifterlms/blob/fbdf449c3f65d7a0d772f255bb6c98e4a1eaa8fc/includes/admin/class.llms.admin.user.custom.fields.php#L199

https://github.com/gocodebox/lifterlms/blob/5.0.0-alpha.6/includes/admin/class.llms.admin.user.custom.fields.php#L76
My idea was actually to raise a deprecation warning if `$fields` is not empty.

includes/admin/class-llms-admin-profile.php
eri-trabiccolo: Can we just do if 5.6+ use the new methods otherwise deal with the fact that we won't reinvent the wheel and there's no error code.
What do you think?

thomasplevy: @eri-trabiccolo that sounds good :+1:

includes/admin/class.llms.admin.user.custom.fields.php
thomasplevy: Okay yea that sounds better

includes/forms/class-llms-form-field.php
thomasplevy: Yes, you're right... but we'll have to run an additional check either way, won't we, in order to determine the data source being used? EG:

`if ( ! is_null( $this->post ) && is_a( WP_User, $this->user ) )`, right?

If we use `$this->data_source` we can infer the type of data source from the already-existing `data_store` settings.

Although, maybe we should do as you're suggesting or...

What if we add a `$this->data_source_type` (also set in the `define_data_source()` method which would be `user` or `post` depending on the context.

Then we can check against the type later?

Maybe you're right though...

includes/forms/class-llms-form-handler.php
thomasplevy: sounds sensible, the argument should be removed from the method in that case too then, right?

includes/forms/class-llms-form-field.php
eri-trabiccolo: > Yes, you're right... but we'll have to run an additional check either way, won't we, in order to determine the data source being used? EG:
> 
> if ( ! is_null( $this->post ) && is_a( WP_User, $this->user ) ), right?
> 
> If we use $this->data_source we can infer the type of data source from the already-existing data_store settings.

well my idea was just to check `if ( ! isset( $this->post ) )` throughout the code, it can be set or not and that should suffice.
But I think I see your "abstraction" point and will go with `$this->data_source_type` and `$this->data_source`.

includes/forms/class-llms-form-handler.php
eri-trabiccolo: yep, I did it in https://github.com/gocodebox/lifterlms/pull/1631/commits/97242940ed39d7e10c693dbed13397a48a05498f

includes/admin/class.llms.admin.user.custom.fields.php
eri-trabiccolo: I was wrong, I didn't realize that apply_filters_deprecated actually checks whether the hook as at leas one filter callback added. So let's go with `apply_filters_deprecated` as you suggested.

includes/admin/class-llms-admin-profile.php
eri-trabiccolo: And you were right on this too.
I didn't feel to duplicate `copy_errors()` as  backward compat code but use the actual `copy_errors()` with wp 5.6+ since we still have to "parse" our error messages to make them look more like the wp ones:
`'<strong>Error</strong>: ...'` e.g. https://github.com/WordPress/WordPress/blob/de519f9dfefa1a10c93510c86b6cb941ae168ce8/wp-includes/user.php#L136-L141

PR #1632, opened 2021-05-11
Students can only review their own quiz attempt results

thomasplevy:
## Description

Fixes https://github.com/gocodebox/private-issues/issues/19

## How has this been tested?

+ Manually

## Types of changes
+ Bugfix

## Checklist:

### DIFF
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -166,7 +166,7 @@ class LLMS_User_Permissions {
 	 *   + Users can view their own grades.
 	 *   + Admins and LMS Managers can view anyone's grade.
 	 *   + Any user who has been explicitly granted the `view_grades` cap can view anyone's grade (via custom code).
-	 *   + Any user who can `edit_post` for the course the quiz belongs to can view grades of the students within that course.
+	 *   + Any instructor/assistant who can `edit_post` for the course the quiz belongs to can view grades of the students within that course.
 	 *
 	 * @since [version]
 	 *
--- a/includes/class.llms.user.permissions.php
+++ b/includes/class.llms.user.permissions.php
@@ -160,6 +160,57 @@ class LLMS_User_Permissions {
 	}
 
 	/**
+	 * Modify a users ability to `view_grades`
+	 *
+	 * Users can view the grades (quiz results) if one of the following conditions is met:
+	 *   + Users can view their own grades.
+	 *   + Admins and LMS Managers can view anyone's grade.
+	 *   + Any user who has been explicitly granted the `view_grades` cap can view anyone's grade (via custom code).
+	 *   + Any user who can `edit_post` for the course the quiz belongs to can view grades of the students within that course.
+	 *
+	 * @since [version]
+	 *
+	 * @param bool[] $allcaps Array of key/value pairs where keys represent a capability name and boolean values
+	 *                        represent whether the user has that capability.
+	 * @param array  $args {
+	 *   Arguments that accompany the requested capability check.
+	 *
+	 *     @type string $0 Requested capability: 'view_grades'.
+	 *     @type int    $1 Current User ID.
+	 *     @type int    $2 Requested User ID.
+	 *     @type int    $3 WP_Post ID of the quiz.
+	 * }
+	 * @return array
+	 */
+	private function handle_cap_view_grades( $allcaps, $args ) {
+
+		// Logged out user or missing required args.
+		if ( empty( $args[1] ) || empty( $args[2] ) || empty( $args[3] ) ) {
+			return $allcaps;
+		}
+
+		list( $requested_cap, $current_user_id, $requested_user_id, $post_id ) = $args;
+
+		// Administrators and LMS managers explicitly have the cap so we don't need to perform any further checks.
+		if ( ! empty( $allcaps[ $requested_cap ] ) ) {
+			return $allcaps;
+		}
+
+		// Users can view their own grades.
+		if ( $current_user_id === $requested_user_id ) {
+			$allcaps[ $requested_cap ] = true;
+		} elseif ( current_user_can( 'edit_post', $post_id ) ) {
+			$instructor = llms_get_instructor( $current_user_id );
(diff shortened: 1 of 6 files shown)


tests/phpunit/unit-tests/user/class-llms-test-student-quizzes.php
```diff
@@ -1,38 +1,64 @@
...
+	private function get_student_with_quizzes( $attempts = 3 ) {
 
 		$uid = $this->factory->user->create();
 		$student = llms_get_student( $uid );
-		$courses = $this->generate_mock_courses( 3, 1, 1, 1 );
+		$courses = $this->generate_mock_courses( $attempts, 1, 1, 1 );
 		$this->complete_courses_for_student( $uid, $courses );
 		return $student;
 
 	}
```
eri-trabiccolo, line 59: _oh gimme gimme some docs_

includes/models/model.llms.student.quizzes.php
```diff
@@ -110,22 +120,39 @@ public function get_attempts_by_quiz( $quiz_id, $args = array() ) {
 	/**
 	 * Retrieve an attempt by attempt id
 	 *
-	 * @param    int $attempt_id  Attempt ID
-	 * @return   LLMS_Quiz_Attempt
-	 * @since    3.16.0
-	 * @version  3.16.0
+	 * @since 3.16.0
+	 * @since [version] Only return
```
eri-trabiccolo: typical case of _documentation interruptus_

includes/models/model.llms.student.quizzes.php
```diff
@@ -45,47 +44,58 @@ public function count_attempts_by_quiz( $quiz_id ) {
...
 	 * Retrieve quiz attempts
 	 *
-	 * @param    int   $quiz_id  WP Post ID of the quiz
-	 * @param    array $args     additional args to pass to LLMS_Query_Quiz_Attempt
-	 * @return   LLMS_Quiz_Attempt[]
 	 * @since    3.16.0
-	 * @version  3.16.0
+	 *
+	 * @param int   $quiz_id  WP Post ID of the quiz.
+	 * @param array $args     Additional args to pass to LLMS_Query_Quiz_Attempt.
```
eri-trabiccolo:
```suggestion
	 * @param int   $quiz_id WP Post ID of the quiz.
	 * @param array $args    Additional args to pass to LLMS_Query_Quiz_Attempt.
```

includes/models/model.llms.student.quizzes.php
```diff
@@ -110,22 +120,39 @@ public function get_attempts_by_quiz( $quiz_id, $args = array() ) {
...
 	 * Decodes an attempt string and returns the associated attempt
 	 *
-	 * @param    string $attempt_key  encoded attempt key
-	 * @return   LLMS_Quiz_Attempt|false
-	 * @since    3.9.0
-	 * @version  3.16.0
+	 * @since 3.9.0
+	 * @since 3.16.0 Unknown.
+	 *
+	 * @param string $attempt_key  Encoded attempt key.
```
eri-trabiccolo:
```suggestion
	 * @param string $attempt_key Encoded attempt key.
```

tests/phpunit/unit-tests/user/class-llms-test-student-quizzes.php
```diff
@@ -1,38 +1,64 @@
...
- * @group    quizzes
- * @group    LLMS_Student
- * @since    3.9.0
- * @version  3.16.11
+ *
+ * @group quizzes
+ * @group student_quizzes
+ * @group LLMS_Student
+ *
+ * @since 3.9.0
```
eri-trabiccolo, line 9: missing version tag?

eri-trabiccolo (changes requested):
changes to doc stuff requested.

This PR fixes the issue it aims to fix (similar thing should be done for assignments) but introduces a regression for the groups add-on.
Because it allowed to view other members attempts, although it was not ideal because any user could, while we should allow that only to users who can.

The thing is that we should be able to initialize the `LLMS_Student_Quizzes` with the user id we need, and then check on whether the current user can view others (students|reports)?

What do you think?
(I'm sorry this is not the easy win you were looking for :P)

thomasplevy:
@eri-trabiccolo thanks for catching this. I've reworked this to introduce a new capability, `view_grades` which is used by the core to allow admins & managers to view anyone's quiz results (which they can already do on the frontend). Instructors will be able to view results for students on their content.

The core will not implement any true frontend for these users to arrive on these screens but it's implemented in a "forward compatible" way as at some point I can see us introducing reporting for instructors/others like the groups plugin currently does.

Users can always view their own results.

The groups plugin is adding it's own handling via https://github.com/gocodebox/lifterlms-groups/pull/141 to allow group members who can view group reports to continue viewing quiz results as expected

includes/class.llms.user.permissions.php
```diff
@@ -159,6 +159,57 @@ public static function get_editable_roles() {
...
 	}
 
+	/**
+	 * Modify a users ability to `view_grades`
+	 *
+	 * Users can view the grades (quiz results) if one of the following conditions is met:
+	 *   + Users can view their own grades.
+	 *   + Admins and LMS Managers can view anyone's grade.
+	 *   + Any user who has been explicitly granted the `view_grades` cap can view anyone's grade (via custom code).
+	 *   + Any user who can `edit_post` for the course the quiz belongs to can view grades of the students within that course.
```
eri-trabiccolo: according to the code... this is true assuming the user is an instructor or an instructor assistant, right?

eri-trabiccolo (approved):
Looks fine to me.
I left a comment (question) though.

includes/class.llms.user.permissions.php
thomasplevy: Yes, updating the comment thanks

thomasplevy:
```suggestion
	 *   + Any instructor/assistant who can `edit_post` for the course the quiz belongs to can view grades of the students within that course.
```

PR #1638, opened 2021-05-20
[user] information shortcode

thomasplevy:
@eri-trabiccolo 

I'm working on figuring out a few different things and I need to create yet another list of our fields (but with slightly different information) so I've conceived something like this and wanted to see what you think.

My idea is that the existing llms-blocks-schema.php can use field data from this to populate the blocks created on install

This can be used to setup the fields on the admin panel

Custom fields could potentially hook into this using data from the tracker option, too, maybe? (I haven't thought through that thourougly)

The field name validation can use this (instead of the existing flat list of field names)

I'm finishing the [user] info shortcode I started way back and nixed but it's actually almost done except for outputting the available fields -- which is where this came from... I didn't want another filter/method for outputting fields that we already have listed elsewhere.

I need a way to compile a list of all LifterLMS (and later custom) and also WP core information fields that we want to be able to use through this shortcode.

I'm also thinking 3rd parties could hook into this with their own custom field information (WooCommerce comes readily to mind and also possibly social learning) 

This likely won't run because it's partial but I wanted to push the relevant files so you can review the strategy.


eri-trabiccolo:
Well yeah of course, as you can imagine, I thought that too when I implemented the "admin" list. So the strategy looks good to me.

About:
> Custom fields could potentially hook into this using data from the tracker option, too, maybe? (I haven't thought through that thourougly).

What you mean? We already use the `llms_admin_profile_fields` filter hook to "inject" the tracked fields so they're disolayed in the admin user profile screen. Do you mean another filter hook to achieve a different goal?

thomasplevy:
@eri-trabiccolo this is pretty functional. Some small changes need to be made to the custom fields plugin too. You can work on that if you have time or I'll get to it tomorrow.

Any thoughts on this? Sorry it's a big changeset. There's also changes pushed to dev in blocks that uses this.

I think that these two can be combined: 

https://github.com/gocodebox/lifterlms-custom-fields/blob/17aa5fbf74542603e4b1f1d753defb585d9dd397/includes/class-llms-cf-fields-tracker.php#L54-L57

The logic from add_custom_fields_user_profile can be used (I think) and the second filter (and function) can probably be removed completely.

This can be used:  https://github.com/gocodebox/lifterlms/pull/1638/files#diff-db9f7286e61fefa59cf6b7fde13e2c8621282c44ffb301f1b874fd4e40d1bf46R58

Probably...

includes/admin/tools/class-llms-admin-tool-install-forms.php
```diff
@@ -0,0 +1,110 @@
...
+ * @package LifterLMS/Admin/Tools/Classes
+ *
+ * @since [version]
+ * @version [version]
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Admin tool to delete pending batches created by a background processor
```
eri-trabiccolo: copy&paste

includes/admin/tools/class-llms-admin-tool-install-forms.php
```diff
@@ -0,0 +1,110 @@
...
+			array(
+				'posts_per_page' => -1,
+				'no_found_rows'  => true,
+				'post_type'      => 'wp_block',
+				'meta_key'       => '_llms_field_id',
+				'meta_compare'   => 'EXISTS',
+			)
+		);
+
+		foreach ( $blocks->posts as $post ) {
```
eri-trabiccolo: so here we delete even custom fields which are reusable blocks, mmm...
Can't we just remove the default ones?
What's the purpose of removing the custom reusable blocks too?

includes/forms/class-llms-form-templates.php
```diff
@@ -229,6 +229,43 @@ private static function get_reusable_block_schema( $field_id ) {
 
 	}
 
+	/**
+	 * Prepares block attributes for a given reusable block
+	 *
+	 * This method loads a reusable block from the blocks schema and attempts to located a user information field
```
eri-trabiccolo:
```suggestion
	 * This method loads a reusable block from the blocks schema and attempts to locate a user information field
```

includes/functions/llms-functions-user-information-fields.php
```diff
@@ -0,0 +1,125 @@
...
+ * @package LifterLMS/Functions
+ *
+ * @since [version]
+ * @version [version]
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Retrieve a single user information field by it's ID attribute.
```
eri-trabiccolo:
```suggestion
 * Retrieve a single user information field by its ID attribute.
```

includes/functions/llms-functions-user-information-fields.php
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * @since [version]
+ *
+ * @param string $id The field's id
+ * @return array|boolean Returns the field settings array or `false` when the field cannot be found.
+ */
+function llms_get_user_information_field( $id ) {
```
eri-trabiccolo: I don't know if the `id` is the correct field's field to look for, I mean, at a certain point we actually decided that the "unique" field was the `name` and not the `id`, and you implemented a js code to prevent duplicates.

eri-trabiccolo (changes requested):
I need to test it and integrate custom fields to understand this all better :/

includes/admin/tools/class-llms-admin-tool-install-forms.php
thomasplevy: Good catch. We'll have to add additional meta data to properly track them but that shouldn't be a problem.

The reason I'm removing them is because if someone *modifies* one of them and they use the tool to restore to the default the *modifications* will persist after the restoration. In order to fully restore to the default versions they *must* be deleted and recreated.

eri-trabiccolo: Ok, but you're talking about removing the llms core ones, right?
Here:
> What's the purpose of removing the custom reusable blocks too?

I meant the purpose of removing the reusable blocks which are custom fields.

includes/functions/llms-functions-user-information-fields.php
thomasplevy: good call

includes/admin/tools/class-llms-admin-tool-install-forms.php
thomasplevy: actually @eri-trabiccolo this *does not* delete reusable blocks created by the custom fields add-ons. The `_llms_field_id` is used by the core to keep track of the reusable blocks created by the core and these are the only ones being removed here.

Reusable blocks containing custom fields would have an `_is_llms_field` meta value (handled in the blocks library), but these are not being removed by this tool.

Is this what you were thinking?

eri-trabiccolo: Oh alright, I confused them. Sorry.

thomasplevy:
@eri-trabiccolo I made some small changes to your CF PR and merged, I think this is better, what do you think?

This does away with the field groups completely in favor of using an exclude list on the admin panel user profile area.

includes/functions/llms-functions-user-information-fields.php
```diff
@@ -0,0 +1,104 @@
...
+	$fields = llms_get_user_information_fields();
+
+	foreach ( $fields as $field ) {
+
+		if ( isset( $field['name'] ) && $name === $field['name'] ) {
+			return $field;
+		}
+	}
+
+	return false;
```
eri-trabiccolo:
```suggestion
	$field_index = array_search( $name, array_column( $fields, 'name' ), true );
	return false === $field_index ? false : $fields[ $field_name ];
```
golf

includes/functions/llms-functions-user-information-fields.php
```diff
@@ -0,0 +1,104 @@
...
+	 * @since [version]
+	 *
+	 * @param array[] $fields List of field definitions.
+	 */
+	return apply_filters( 'llms_user_information_fields', $fields );
+
+}
+
+/**
+ * Retrieve user infor fields used by the block editor
```
eri-trabiccolo:
```suggestion
 * Retrieve user information fields used by the block editor
```

includes/functions/llms-functions-user-information-fields.php
```diff
@@ -0,0 +1,104 @@
...
+	// Add a value so we can use array_interect_key() later.
+	$keys = array_fill_keys( $keys, 1 );
+
+	// Return a reduced list.
+	return array_map(
+		function( $field ) use ( $keys ) {
+			return array_intersect_key( $field, $keys );
+		},
+		$fields
+	);
```
eri-trabiccolo, line 97: <3

includes/schemas/llms-reusable-blocks.php
```diff
@@ -0,0 +1,150 @@
+<?php
+/**
+ * Default form field blocks schema
+ *
+ * This file returns a list of the default LifterLMS form fields
+ * used to build an initial set of reusable blocks used across the
+ * core user information forms (checkout, registration, and account).
+ *
+ * Each field block is an incomplete form field definition. Each field
+ * is linked to a user information form field by it's name attribute which
```
eri-trabiccolo:
```suggestion
 * is linked to a user information form field by its name attribute which
```

includes/schemas/llms-reusable-blocks.php
```diff
@@ -0,0 +1,150 @@
...
+ * will match an info field by it's id attribute.
```
eri-trabiccolo:
```suggestion
 * will match an info field by its id attribute.
```

includes/schemas/llms-user-information-fields.php
```diff
@@ -0,0 +1,156 @@
+<?php
+/**
+ * User information fields schema
+ *
+ * A list of user information fields used by LifterLMS in various places, namely to build
+ * the editable user information forms (Checkout, Registration, and Edit Account).
+ *
+ * Each item in this list should be an array compatible with the `LLMS_Form_Field` classes
```
eri-trabiccolo:
```suggestion
 * Each item in this list should be an array compatible with the `LLMS_Form_Field` class'
```

includes/schemas/llms-user-information-fields.php
```diff
@@ -0,0 +1,156 @@
...
+ * settings array.
+ *
+ * Fields can be added and modified using the `llms_user_information_fields` filter.
+ *
+ * @package LifterLMS/Schemas
+ *
+ * @since [version]
+ * @version [version]
+ *
+ * @see llms_get_user_information_fields() Retrieves the (filtered) schema
```
eri-trabiccolo:
```suggestion
 * @see llms_get_user_information_fields() Retrieves the (filtered) schema.
```

eri-trabiccolo (changes requested):
Plus I would add:
- Shortcodes do not display default value. To reproduce: try to use a shortcode of a radio button form field (which as a default by default :P).
- I think that 'user' is not a good shortcode tag, we risk collisions, we are not obliged to prefix them but: https://developer.wordpress.org/plugins/shortcodes/#shortcode-best-practices
- I've found one possible problem when using the tool to reinstall the forms: If you created another form (with the custom fields add-on), which even by default uses the llms core fields, since they are reusable blocks and they get deleted and re-created (with different ids of course), that custom form will contain reusable blocks which actually do not exist anymore:
Schermata 2021-05-27 alle 16 41 56

thomasplevy:
> * Shortcodes do not display default value. To reproduce: try to use a shortcode of a radio button form field (which as a default by default :P).

1. Should custom fields add-on implement default value functionality? Probably through a filter in the core? No core fields (excepting country) has a default value.
2. Should the default value be used as the default value of the `or` (fallback) attribute? Probably, right? Then the default can be overridden if needed?

> * I think that 'user' is not a good shortcode tag, we risk collisions, we are not obliged to prefix them but: https://developer.wordpress.org/plugins/shortcodes/#shortcode-best-practices

I was aiming for something short and I did consider there could be a collision but maybe arrogantly dismissed this concern.

I wanted to keep it very short. Do you have an alternative suggestion?

> * I've found one possible problem when using the tool to reinstall the forms: If you created another form (with the custom fields add-on), which even by default uses the llms core fields, since they are reusable blocks and they get deleted and re-created (with different ids of course), that custom form will contain reusable blocks which actually do not exist anymore:

Mmmm... I'm thinking that we should do a find & replace on non core forms when the tool runs in this case. Does that make sense to you?

includes/functions/llms-functions-user-information-fields.php
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * @param string $name The field's name.
+ * @return array|boolean Returns the field settings array or `false` when the field cannot be found.
+ */
+function llms_get_user_information_field( $name ) {
+
+	$fields = llms_get_user_information_fields();
+
+	$field_index = array_search( $name, array_column( $fields, 'name' ), true );
+	return false === $field_index ? false : $fields[ $field_name ];
```
eri-trabiccolo:
```suggestion
	return false === $field_index ? false : $fields[ $field_index ];
```

eri-trabiccolo (changes requested):

eri-trabiccolo:
> > * Shortcodes do not display default value. To reproduce: try to use a shortcode of a radio button form field (which as a default by default :P).
> 
> 1. Should custom fields add-on implement default value functionality? Probably through a filter in the core? No core fields (excepting country) has a default value.

Mmm I think it should be a core thing though.

> 2. Should the default value be used as the default value of the `or` (fallback) attribute? Probably, right? Then the default can be overridden if needed?

Yeah, makes sense.

> > * I think that 'user' is not a good shortcode tag, we risk collisions, we are not obliged to prefix them but: https://developer.wordpress.org/plugins/shortcodes/#shortcode-best-practices
> 
> I was aiming for something short and I did consider there could be a collision but maybe arrogantly dismissed this concern.
> 
> I wanted to keep it very short. Do you have an alternative suggestion?

If you say so it means that you don't like "llms_user".... so no I don't have anything not prefixed...


> > * I've found one possible problem when using the tool to reinstall the forms: If you created another form (with the custom fields add-on), which even by default uses the llms core fields, since they are reusable blocks and they get deleted and re-created (with different ids of course), that custom form will contain reusable blocks which actually do not exist anymore:
> 
> Mmmm... I'm thinking that we should do a find & replace on non core forms when the tool runs in this case. Does that make sense to you?

It could, but since we're there, can't we just replace the content of the "original" reusable blocks and put them again in the core forms?

thomasplevy:
> > > > * Shortcodes do not display default value. To reproduce: try to use a shortcode of a radio button form field (which as a default by default :P).
> > > 
> > > 
> > > 
> > > 1. Should custom fields add-on implement default value functionality? Probably through a filter in the core? No core fields (excepting country) has a default value.
> > 
> > Mmm I think it should be a core thing though.
> 
> How are we going to get the default value of the field for a custom field then? I think we need to pull it from the tracked fields list, right?

Mmm defaults is something we pass through the filter, along with the "options" which is useful to get the "label" of a key.
https://github.com/gocodebox/lifterlms-custom-fields/blob/6de847914db2a07a1d016413c1ee8088401cf253/includes/class-llms-cf-fields-tracker.php#L112-L121

So this is an information you should already have, no?

thomasplevy:
Mmmmm, okay yes you're right. We can do it in teh core then

thomasplevy:
I suppose `llms_user` for the shortcode name isn't that long but it just feels clunky to me

eri-trabiccolo:
> I suppose `llms_user` for the shortcode name isn't that long but it just feels clunky to me

I know, but we're kinda forced I think. Best practice != Required, I know...

thomasplevy:
Shortcode updated to [llms-user]

I've decided that we're not going to implement an automatic fallback system... If someone wants the default value from the "registered" field they can enter it in the `or` attribute.

Implementing the automatic isn't terribly difficult but it required a lot of changes across other functions to make it possible and I don't want to rewrite tests and add a whole host of new tests to add the necessary functionality.

I also found it nearly impossible to implement an easy way to disable the automatic fallback if that functionality wasn't desired -- without adding additional attributes which require additional documentation.

I don't think it's that hard to say "Add the default value in the shortcode if you want one" -- maybe I'll be proven incorrect, we can add it later if it seems necessary through lots of support requests or something

thomasplevy:
@eri-trabiccolo when you have a moment can you look at my recent commits and let me know if you see any issues?

includes/abstracts/llms.abstract.notification.view.php
```diff
@@ -159,6 +159,7 @@ abstract protected function set_title();
 	 * @since 3.8.0
 	 * @since 3.31.0 Add filter on `$basic_options` class class property.
 	 * @since 3.37.19 Moved the retrieval of the associated llms post into a protected method.
+	 * @since [versoin] Force [llms-user] shortocde to the user ID of the user who triggered the notification.
```
eri-trabiccolo:
```suggestion
	 * @since [version] Force [llms-user] shortocde to the user ID of the user who triggered the notification.
```

includes/emails/class.llms.email.php
```diff
@@ -5,8 +5,7 @@
  * @package LifterLMS/Emails/Classes
  *
  * @since 1.0.0
- * @version 4.0.0
- */
+ * @version [version]
```
eri-trabiccolo, line 8:
```suggestion
 * @version [version]
 */
```

includes/emails/class.llms.email.php
```diff
@@ -255,13 +254,14 @@ public function add_recipients( $recipients = array() ) {
 	 *  Format string method
 	 *  Finds and replaces merge fields with appropriate data
```
eri-trabiccolo:
```suggestion
	 *  
	 *  Finds and replaces merge fields with appropriate data.
```

eri-trabiccolo (changes requested):
Docs

~Will review it better on Monday~
Looks good, aside from what I requested below.

PR #1643, opened 2021-05-28
Add missing form post type labels

### LINKED ISSUE #1641: llms_form post type, missing labels
llms_form labels:
```
			'labels'              => array(
				'name'          => __( 'LifterLMS Forms', 'lifterlms' ),
				'singular_name' => __( 'LifterLMS Form', 'lifterlms' ),
				'search_items'  => __( 'Search Forms', 'lifterlms' ),
				'menu_name'     => __( 'Forms', 'lifterlms' ),
			),
```
course labels:
```
				'labels'              => array(
					'name'               => __( 'Courses', 'lifterlms' ),
					'singular_name'      => __( 'Course', 'lifterlms' ),
					'menu_name'          => _x( 'Courses', 'Admin menu name', 'lifterlms' ),
					'add_new'            => __( 'Add Course', 'lifterlms' ),
					'add_new_item'       => __( 'Add New Course', 'lifterlms' ),
					'edit'               => __( 'Edit', 'lifterlms' ),
					'edit_item'          => __( 'Edit Course', 'lifterlms' ),
					'new_item'           => __( 'New Course', 'lifterlms' ),
					'view'               => __( 'View Course', 'lifterlms' ),
					'view_item'          => __( 'View Course', 'lifterlms' ),
					'search_items'       => __( 'Search Courses', 'lifterlms' ),
					'not_found'          => __( 'No Courses found', 'lifterlms' ),
					'not_found_in_trash' => __( 'No Courses found in trash', 'lifterlms' ),
					'parent'             => __( 'Parent Course', 'lifterlms' ),
				),
```
As you can see a bunch of them are missing and are nice to have in admin.

eri-trabiccolo:
## Description
Fixes #1641 

## How has this been tested?
manually

## Screenshots 
n/a

## Types of changes

## Checklist:


eri-trabiccolo:
I left out the `parent` label as forms are not hierarchical.
Also If we want a button to be shown among these:
Schermata 2021-05-28 alle 12 37 01

we should set `'show_in_admin_bar'   => true,`
but I think we should even revert teh singular name to 'LifterLMS Form' so to avoid collisions with other plugins.
@thomasplevy what do you think?

thomasplevy:
@eri-trabiccolo I don't think we need to include forms in the admin bar list. Do you? I don't think this menu is actually particularly useful anyway but I don't think forms will be created frequently enough to warrant it. We can add it later if we get feedback on the contrary.

> but I think we should even revert teh singular name to 'LifterLMS Form' so to avoid collisions with other plugins.

I think that it wont actually "collide" so much as list the same thing in two places (or possible in the same list)

Mostly I don't think in most cases we need to say "LifterLMS Forms" at the top of the screen or in the sidebar because, well, when you get there you know where you came from and why you're there.

We don't list courses as "LifterLMS Courses" or orders as "LifterLMS orders" although these are less-commonly used than forms, I suppose.

Mostly my thinking is that every time I'm editing a form in the block editor and see "LifterLMS Form" as the title of the settings panel on the right-hand side I think "it should just say form" but as you're noting it could be confusing in other areas...

Maybe we should call them "User Information Forms" to avoid this kind of confusion?

@willmiddleton-lifterlms @chrisbadgett @nrherron92 @toyinogunseinde any thoughts to help us figure out what to call these things given the issue @eri-trabiccolo has raised?

thomasplevy:
@eri-trabiccolo Going with this for now we can adjust later if needed

PR #1644, opened 2021-05-28
Fix disabled fields not clearing usermeta when needed

### LINKED ISSUE #1633: [5.x] When changing to a Country with no States, previous State's value is no cleared.
### Reproduction Steps

+ Register with Country US and State AR
+ Edit your account
+ Change your country to something that has no states, e.g. United States Minor Outlying Islands
+ the States select field will disappear
+ Save
+ Look at your profile in admin (or through reporting)

### Expected Behavior

+ I'd expect the state to be empty

### Actual Behavior

+ The state is still AR
(we should not disable the field, but rather empy it and send it over anyways, right?)

### Error Messages / Logs

n/a

```
LifterLMS 5.0.0-alpha.6
```

+ [x] Locally
+ [ ] On a staging site
+ [ ] On a production website

n/a

eri-trabiccolo:
## Description
Fixes #1633 

@thomasplevy  what do you think? This is the simplest/shortest/cleanest(?) solution I could think about.

## How has this been tested?
manually

## Screenshots 
n/a

## Types of changes

## Checklist:

### DIFF
--- a/assets/js/app/llms-forms.js
+++ b/assets/js/app/llms-forms.js
@@ -338,14 +338,11 @@ LLMS.Forms = {
 	 */
 	update_locale_info_for_field: function( $field, label ) {
 
-		var $parent = this.get_field_parent( $field );
-		$field.removeAttr( 'disabled' );
 		if ( label ) {
 			this.update_label( $field, label );
-			$parent.show();
+			this.enable_field( $field );
 		} else {
-			$field.attr( 'disabled', 'disabled' );
-			$parent.hide();
+			this.disable_field( $field );
 		}
 
 	},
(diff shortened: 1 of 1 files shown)


thomasplevy:
@eri-trabiccolo I didn't test this yet but this *should* generally work... However, I was initially thinking that this could be handled server side, possibly here:

https://github.com/gocodebox/lifterlms/blob/93bd4ca55232c6788f5fc11ddb1acbe7cc3bd838/includes/forms/class-llms-form-handler.php#L181-L208

Did you look at a server-side approach?

eri-trabiccolo:
Nope, I thought that since the "disabling" happened client side I should better handle this there.
But ok I guess I can do it server side with the same logic: zip code and/or state can be empty if the country is this and this.
Do you prefer the server side approach?

thomasplevy:
@eri-trabiccolo I was actually thinking this would make sure it works on the admin profile too but the JS is loaded there as well now so this works, let's proceed.

Thanks <3

PR #1654, opened 2021-06-09
Issue/1652

### LINKED ISSUE #1652: [5.x] Migration script
The 5.0 update scripts need to be rewritten:

1) An LLMS_Admin_Notice should pop to "welcome" users to LifterLMS 5.0. This will link to the welcome guide blog post that @chrisbadgett is working on (and will publish in tandem with the release of 5.0).
2) ~~Remove unused "legacy" option values from the db: https://github.com/gocodebox/lifterlms/issues/1592 & https://github.com/gocodebox/lifterlms/issues/1593~~ Change legacy options to not be autoloaded

Some of the legacy options are used to setup default values in the forms when created on plugin install. I need to review the order of operations but to ensure these legacy options can be used to generate new defaults (but then never need to be referenced again) we should check if the forms are installed prior to running an update or after. If they're installed after the migration script should include a form installation step and then remove the legacy options after

Text of the welcome notice:

**Welcome to LifterLMS 5.0!** [on it's own line within the notice, bolded so it stands out a bit]

This new version of LifterLMS brings you the power to build and customize your student information forms using a simple point and click interface constructed on top of the WordPress block editor. Customization like the removal of default fields, changing the text of field labels, or reordering fields within a form is all possible without any code or professional help.

[Read more][link to welcome blog post], [read the docs][link to docs category], or [Get started now][link to the forms admin page]!

eri-trabiccolo:
## Description
**wip
Opened to simplify the review/feedback (e.g. on the admin notice) process**
 
Fixes #1652 

Also:
1) we need to add the proper link to the announcement post when ready
2) we need to ad the proper link to the documentation when ready

## How has this been tested?
Manually and new unit tests

## Types of changes
Bug fix (non-breaking change which fixes an issue)/New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/includes/functions/updates/llms-functions-updates-500.php
+++ b/includes/functions/updates/llms-functions-updates-500.php
@@ -68,13 +68,13 @@ function llms_update_500_add_admin_notice() {
 			sprintf(
 				// Translators: %1$s = Link to the documentation on lifterlms.com; %2$s = Closing anchor tag.
 				__( '%1$sRead the docs%2$s', 'lifterlms' ),
-				'<a href="" target="_blank" rel="noopener">',
+				'<a class="button" href="" target="_blank" rel="noopener">',
 				'</a>'
 			),
 			sprintf(
 				// Translators: %1$s = Opening anchor tag to Forms admin page; %2$s = Closing anchor tag.
 				__( '%1$sGet Started%2$s', 'lifterlms' ),
-				'<a href="' . add_query_arg(
+				'<a class="button-primary" href="' . add_query_arg(
 					array(
 						'post_type' => 'llms_form',
 					),
@@ -86,11 +86,11 @@ function llms_update_500_add_admin_notice() {
 	);
 
 	LLMS_Admin_Notices::add_notice(
-		basename( __FILE__, '.php' ),
+		'v500-welcome-msg',
 		$html,
 		array(
 			'type'             => 'success',
-			'dismiss_for_days' => 730, // @TODO: there should be a "forever" setting here.
+			'dismiss_for_days' => 0,
 			'remindable'       => false,
 		)
 	);
--- a/includes/functions/updates/llms-functions-updates-500.php
+++ b/includes/functions/updates/llms-functions-updates-500.php
@@ -51,41 +51,40 @@ function llms_update_500_legacy_options_autoload_off() {
  */
 function llms_update_500_add_admin_notice() {
 
+	require_once LLMS_PLUGIN_DIR . 'includes/admin/class.llms.admin.notices.php';
+
+	$notice_id        = 'v500-welcome-msg';
+	$get_started_link = add_query_arg(
+		array(
+			'post_type'        => 'llms_form',
+		),
+		admin_url( 'edit.php' )
+	);
+
 	$html = sprintf(
 		'<strong>%1$s</strong><br><br>%2$s<br><br>%3$s',
 		__( 'Welcome to LifterLMS 5.0!', 'lifterlms' ),
 		__( 'This new version of LifterLMS brings you the power to build and customize your student information forms using a simple point and click interface constructed on top of the WordPress block editor. Customization like the removal of default fields, changing the text of field labels, or reordering fields within a form is all possible without any code or professional help.', 'lifterlms' ),
 		sprintf(
-			// Translators: %1$s = Link to the welcome blog post on lifterlms.com, %2$s = Link to the documentation on lifterlms.com, %3$s Link to the Forms admin page.
-			__( '%1$s, %2$s or %3$s', 'lifterlms' ),
+			// Translators: %1$s = Link to the welcome blog post on lifterlms.com, %2$s Link to the Forms admin page.
+			__( '%1$s or %2$s', 'lifterlms' ),
 			sprintf(
 				// Translators: %1$s = Opening anchor tag to the welcome blog post on lifterlms.com; %2$s = Closing anchor tag.
 				__( '%1$sRead More%2$s', 'lifterlms' ),
-				'<a href="" target="_blank" rel="noopener">',
-				'</a>'
-			),
-			sprintf(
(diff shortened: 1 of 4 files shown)


includes/functions/updates/llms-functions-updates-500.php
```diff
@@ -0,0 +1,117 @@
...
+		WHERE option_name IN (" . implode( ', ', array_fill( 0, count( $legacy_options_to_stop_autoloading ), '%s' ) ) . ')';
+
+	$wpdb->query(
+		$wpdb->prepare(
+			$sql,
+			$legacy_options_to_stop_autoloading
+		)
+	); // db call ok; no-cache ok.
+
+	set_transient( 'llms_update_500_autoload_off_legacy_options', 'complete', DAY_IN_SECONDS );
```
thomasplevy: I don't think we need a transient here, updating 3 options in a single query shouldn't require pagination, right?
```suggestion
```

includes/functions/updates/llms-functions-updates-500.php
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * @return void|true True if it needs to run again, nothing if otherwise.
+ */
+function llms_update_500_update_db_version() {
+
+	if ( 'complete' !== get_transient( 'llms_update_500_autoload_off_legacy_options' ) ) {
+		// Needs to run again.
+		return true;
+	}
+
```
thomasplevy:
```suggestion
```

Like above, should be able to remvoe this, yes?

includes/functions/updates/llms-functions-updates-500.php
```diff
@@ -0,0 +1,117 @@
...
+			)
+		)
+	);
+
+	LLMS_Admin_Notices::add_notice(
+		basename( __FILE__, '.php' ),
+		$html,
+		array(
+			'type'             => 'success',
+			'dismiss_for_days' => 730, // @TODO: there should be a "forever" setting here.
```
thomasplevy: Based on the way this message is displayed, it'll be added and once dismissed never added again (unless this function is called)

So we can actually set `dismiss_for_days` to `0` so that data is set to allow it to be re-displayed later.

```suggestion
			'dismiss_for_days' => 0,
```

includes/functions/updates/llms-functions-updates-500.php
```diff
@@ -0,0 +1,117 @@
...
+			sprintf(
+				// Translators: %1$s = Link to the documentation on lifterlms.com; %2$s = Closing anchor tag.
+				__( '%1$sRead the docs%2$s', 'lifterlms' ),
+				'<a href="" target="_blank" rel="noopener">',
+				'</a>'
+			),
+			sprintf(
+				// Translators: %1$s = Opening anchor tag to Forms admin page; %2$s = Closing anchor tag.
+				__( '%1$sGet Started%2$s', 'lifterlms' ),
+				'<a href="' . add_query_arg(
```
thomasplevy:
```suggestion
				'<a class="button-primary" href="' . add_query_arg(
```

includes/functions/updates/llms-functions-updates-500.php
```diff
@@ -0,0 +1,117 @@
...
+			sprintf(
+				// Translators: %1$s = Opening anchor tag to the welcome blog post on lifterlms.com; %2$s = Closing anchor tag.
+				__( '%1$sRead More%2$s', 'lifterlms' ),
+				'<a href="" target="_blank" rel="noopener">',
+				'</a>'
+			),
+			sprintf(
+				// Translators: %1$s = Link to the documentation on lifterlms.com; %2$s = Closing anchor tag.
+				__( '%1$sRead the docs%2$s', 'lifterlms' ),
+				'<a href="" target="_blank" rel="noopener">',
```
thomasplevy:
```suggestion
				'<a class="button" href="" target="_blank" rel="noopener">',
```

thomasplevy (changes requested):
Overall good, some notes here

And after reviewing what it actually looks like I think we can drop the link to the technical docs (we'll have a link to that in the blog post)

With only two options I think having a main action and a secondary action is satisfactory

Final thought: if someone clicks "getting started" i think the notice should be dismissed, we should be able to do that pretty easily by adding a nonce link that hooks into the existing actions to hide dismiss/remove notices

includes/functions/updates/llms-functions-updates-500.php
```diff
@@ -0,0 +1,117 @@
...
+					),
+					admin_url( 'edit.php' )
+				) . '" >',
+				'</a>'
+			)
+		)
+	);
+
+	LLMS_Admin_Notices::add_notice(
+		basename( __FILE__, '.php' ),
```
thomasplevy:
```suggestion
		'v500-welcome-msg',
```
Or something to that effect.

includes/functions/updates/llms-functions-updates-500.php
eri-trabiccolo: Yup, you're right

eri-trabiccolo:
> Final thought: if someone clicks "getting started" i think the notice should be dismissed, we should be able to do that pretty easily by adding a nonce link that hooks into the existing actions to hide dismiss/remove notices

@thomasplevy 

It actually isn't that simple because you know the message is saved in the db, and the nonce URL will be part of the message... Additionally we'll have two different nonces one for the get started link and another one for the dismiss icon link, with the latter that will win. Meaning that the nonce will be invalid for the get started link.

What do you think if we filter the notice message, a filter param will be the nonce.
The idea is that we add the notice with the Get Started button having a "merge code" in its href, so that we can replace it with the same nonce as the dismiss icon.

Let me know, if you "like" it or you have alternatives (maybe js?).

thomasplevy:
@eri-trabiccolo mmm you're right. Let's leave it as is then and when users want the message gone they hit the "X" in the right corner and it's gone.

thomasplevy:
Modified this slightly to remove the "or" between the buttons and switch the text CTA to a secondary (default) button

Thanks @eri-trabiccolo

PR #1657, opened 2021-06-11
Inc helper

thomasplevy:
Includes the Helper as a library

See https://github.com/gocodebox/lifterlms-helper/pull/22

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,8 @@
     "deliciousbrains/wp-background-processing": "1.0.2",
     "lifterlms/lifterlms-blocks": "2.0.0-beta.6",
     "lifterlms/lifterlms-rest": "1.0.0-beta.19",
-    "woocommerce/action-scheduler": "3.2.0"
+    "woocommerce/action-scheduler": "3.2.0",
+    "lifterlms/lifterlms-helper": "9999999-dev"
   },
   "require-dev": {
     "lifterlms/lifterlms-tests": "dev-trunk",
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -32,7 +32,8 @@
     "deliciousbrains/wp-background-processing": "1.0.2",
     "lifterlms/lifterlms-blocks": "2.0.0-beta.6",
     "lifterlms/lifterlms-rest": "1.0.0-beta.19",
-    "woocommerce/action-scheduler": "3.2.0"
+    "woocommerce/action-scheduler": "3.2.0",
+    "lifterlms/lifterlms-helper": "9999999-dev"
```
eri-trabiccolo: is this on purpose?

includes/class-llms-loader.php
```diff
@@ -303,16 +303,33 @@ public function includes_admin() {
...
+			array(
+				'const' => 'LLMS_HELPER_LIB',
+				'test'  => ! class_exists( 'LifterLMS_Helper' ),
+				'file'  => LLMS_PLUGIN_DIR . 'libraries/lifterlms-helper/lifterlms-helper.php',
+			),
+		);
+
+		foreach ( $libs as $lib ) {
+
+			// var_dump( $lib );
```
eri-trabiccolo:
```suggestion
```

includes/class-llms-loader.php
```diff
@@ -303,16 +303,33 @@ public function includes_admin() {
 	 */
```
eri-trabiccolo, line 303: missing changelog for the method and version "update" in the file header.

eri-trabiccolo (changes requested):
I guess we also need to remove the helper itself from the list under Status -> Beta Testing, I know it doesn't belong here, just as reminder :D

composer.json
thomasplevy: Yes, just for testing, that's equivalent to "dev-trunk" since there wasn't a composer installable release available (yet). Updating with the proper version now that it's released

Issue #1589, opened 2021-04-13
[5.x] UX for forms missing required fields


thomasplevy:
Currently we apply "validation" within the block editor when the User Email field is deleted/removed from a form. An error notice is added to the editor with an action button to restore the required field. This *basically* works but it's a little hacky and I think a server-side solution should be applied as well.

In the event that a required field is missing from a form, I think we should add the field dynamically (with base settings) via PHP on page load.

This will ensure that if there's any amount of monkeying the forms will continue to function.

Additionally, I'm not sure whether or not we should define the user password field as required as well.

I had initially wished to build a passwordless registration flow which would allow users to remove the password creation and trigger a password reset email after account creation. This is a not uncommon registration flow:

1. Enter email
2. Receive an email
3. Click the link
4. Add your password
5. Login

If we keep password as an "optional" field we need to add a handler to fire off the reset email upon account creation. Currently it's broken if password is ommitted.

Finally, on the account edit screen I think we should *force* email AND password to always be editable but allow users to determine what other fields they want on their account edit form. This means that we'll likely need to add some conditional behavior to determine what fields are required based on the form location.

eri-trabiccolo:
@thomasplevy 
I did the basic fix in #1688 meaning that I add the email field via PHP on checkout/registration form.

What about this instead?
> In the event that a required field is missing from a form, I think we should add the field dynamically (with base settings) via PHP on page load.
> 
> This will ensure that if there's any amount of monkeying the forms will continue to function.

I don't even understand what this means :D a required field is missing from a form, if it's missing how can it be required?
Can you elaborate with a use case?
Thanks <3

I also think that both the two below should be addressed in another issue. what do you think?

> Additionally, I'm not sure whether or not we should define the user password field as required as well.
> 
> I had initially wished to build a passwordless registration flow which would allow users to remove the password creation and trigger a password reset email after account creation. This is a not uncommon registration flow:
> 
> Enter email
> Receive an email
> Click the link
> Add your password
> Login
> If we keep password as an "optional" field we need to add a handler to fire off the reset email upon account creation. Currently it's broken if password is ommitted.
> 
> Finally, on the account edit screen I think we should force email AND password to always be editable but allow users to determine what other fields they want on their account edit form. This means that we'll likely need to add some conditional behavior to determine what fields are required based on the form location.

thomasplevy:
@eri-trabiccolo Your solution in #1688 looks pretty solid to me (with minor changes I noted in my review)

[...]

PR #1688, opened 2021-06-30
Fix checkout form with no (visible) fields and always require email fields in checkout/registration

### LINKED ISSUE #1607: [5.x] A checkout form with no (visible) fields is weird
### Reproduction Steps

+ Create a checkout form with only an email and password field
+ Set the visibility value of both fields to "logged out"
+ View the checkout form as logged in user

### Expected Behavior

+ I'm not entirely sure

### Actual Behavior

+ It looks weird

Screenshot_2021-04-21_09-22-22

eri-trabiccolo:
## Description
Fixes #1607 
Partial fix for #1589 (Only requires email address for checkout/registration fields).

## How has this been tested?
manually and with old and new unit tests (missing tests on the new `is_block_visible_in_list()`)

## Types of changes

## Checklist:

### DIFF
--- a/includes/forms/class-llms-form-handler.php
+++ b/includes/forms/class-llms-form-handler.php
@@ -308,7 +308,7 @@ class LLMS_Form_Handler {
 	 * Form submission handler
 	 *
 	 * @since 5.0.0
-	 * @since [version] Remove invisible fields from when loading the form.
+	 * @since [version] Remove invisible fields from when loading the checkout form.
 	 *
 	 * @param array  $posted_data User-submitted form data.
 	 * @param string $location    Form location ID.
@@ -320,10 +320,14 @@ class LLMS_Form_Handler {
 		// Determine the user action to perform.
 		$action = get_current_user_id() ? 'update' : 'registration';
 
-		// Load the form, filtering out invisible fields.
-		add_filter( 'llms_forms_remove_invisible_field', '__return_true', 999 );
+		// Load the form, filtering out invisible fields, only for checkout form.
+		if ( 'checkout' === $location ) {
+			add_filter( 'llms_forms_remove_invisible_field', '__return_true', 999 );
+		}
 		$fields = $this->get_fields( $action, $location, $args );
-		remove_filter( 'llms_forms_remove_invisible_field', '__return_true', 999 );
+		if ( 'checkout' === $location ) {
+			remove_filter( 'llms_forms_remove_invisible_field', '__return_true', 999 );
+		}
 
 		if ( is_wp_error( $fields ) ) {
 			return $this->submit_error( $fields, $posted_data, $action );
--- a/includes/forms/class-llms-form-handler.php
+++ b/includes/forms/class-llms-form-handler.php
@@ -340,6 +340,7 @@ class LLMS_Form_Handler {
 	 * Form fields submission
 	 *
 	 * @since 5.0.0
+	 * @since [version] Added "lifterlms_user_${action}_required_data" filter, to filter the required fields validity of the form submission.
 	 *
 	 * @param array   $posted_data User-submitted form data.
 	 * @param string  $location    Form location ID.
--- a/includes/forms/class-llms-form-handler.php
+++ b/includes/forms/class-llms-form-handler.php
@@ -124,6 +124,7 @@ class LLMS_Form_Handler {
 	 * Modify LifterLMS Fields prior to performing submit handler validations.
 	 *
 	 * @since 5.0.0
+	 * @since [version] Do not allow submitting a password change without providing a `password_current`
 	 *
 	 * @param array   $posted_data User submitted form data (passed by reference).
 	 * @param string  $location    Form location ID.
@@ -140,12 +141,24 @@ class LLMS_Form_Handler {
 		 * If email address and passwords aren't submitted we can mark them as "optional" fields.
 		 *
 		 * These fields are dynamically toggled and disabled if they're not modified.
+		 * Process `password_current` as last as it depends on `password` field submission.
 		 */
 		foreach ( array( 'email_address', 'password', 'password_current' ) as $field_id ) {
 
 			// If the field exists and it's not included (or empty) in the posted data.
 			$index = LLMS_Forms::instance()->get_field_by( $fields, 'id', $field_id, 'index' );
 			if ( false !== $index && empty( $posted_data[ $fields[ $index ]['name'] ] ) ) {
+
+				// When updating a password, the `password_current` is mandatory.
+				if ( 'account' === $location && 'password_current' === $field_id ) {
+					// Get `password` field.
+					$password_index = LLMS_Forms::instance()->get_field_by( $fields, 'id', 'password', 'index' );
+					// If a `passowrd` feld has been submitted then the `password_current` cannot be skipped.
+					if ( false !== $password_index &&
+							! empty( $posted_data[ $fields[ $password_index ]['name'] ] ) ) {
+						continue;
+					}
+				}
 
 				// Remove the field so we don't accidentally save an empty value later.
 				unset( $posted_data[ $fields[ $index ]['name'] ] );
--- a/includes/forms/class-llms-form-handler.php
+++ b/includes/forms/class-llms-form-handler.php
@@ -5,7 +5,7 @@
  * @package  LifterLMS/Classes
  *
  * @since 5.0.0
- * @version 5.0.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -308,6 +308,7 @@ class LLMS_Form_Handler {
 	 * Form submission handler
 	 *
 	 * @since 5.0.0
+	 * @since [version] Remove invisible fields from when loading the form.
 	 *
 	 * @param array  $posted_data User-submitted form data.
 	 * @param string $location    Form location ID.
--- a/includes/forms/class-llms-form-handler.php
+++ b/includes/forms/class-llms-form-handler.php
@@ -320,9 +320,9 @@ class LLMS_Form_Handler {
 		$action = get_current_user_id() ? 'update' : 'registration';
 
 		// Load the form, filtering out invisible fields.
-		add_filter( 'llms_filter_out_invisible_field', '__return_true', 999 );
+		add_filter( 'llms_forms_remove_invisible_field', '__return_true', 999 );
 		$fields = $this->get_fields( $action, $location, $args );
-		remove_filter( 'llms_filter_out_invisible_field', '__return_true', 999 );
+		remove_filter( 'llms_forms_remove_invisible_field', '__return_true', 999 );
 
 		if ( is_wp_error( $fields ) ) {
 			return $this->submit_error( $fields, $posted_data, $action );
--- a/includes/forms/class-llms-form-handler.php
+++ b/includes/forms/class-llms-form-handler.php
@@ -319,8 +319,11 @@ class LLMS_Form_Handler {
 		// Determine the user action to perform.
 		$action = get_current_user_id() ? 'update' : 'registration';
 
-		// Load the form.
+		// Load the form, filtering out invisible fields.
+		add_filter( 'llms_filter_out_invisible_field', '__return_true', 999 );
 		$fields = $this->get_fields( $action, $location, $args );
+		remove_filter( 'llms_filter_out_invisible_field', '__return_true', 999 );
+
 		if ( is_wp_error( $fields ) ) {
 			return $this->submit_error( $fields, $posted_data, $action );
 		}
@@ -333,6 +336,7 @@ class LLMS_Form_Handler {
 	}
 
 	/**
+	 * Form fields submission
 	 *
 	 * @since 5.0.0
 	 *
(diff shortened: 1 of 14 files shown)


includes/forms/class-llms-forms.php
```diff
@@ -130,10 +132,23 @@ public function are_usernames_enabled() {
 	 */
 	private function block_to_field_settings( $block ) {
 
+		$is_visible = $this->is_block_visible( $block );
+		/**
+		 * Filters whether or not invisible fields should be included
+		 *
+		 * If the block is not visible (according to LLMS block-level visibility settings)
+		 * it will return an empty array (signaling the field to be removed).
+		 *
```
thomasplevy, line 143:
```suggestion
		 *
		 * @since [version]
		 *
```

includes/forms/class-llms-forms.php
```diff
@@ -130,10 +132,23 @@ public function are_usernames_enabled() {
...
+		 * @param $filter Whether or not invisible fields should be included. Default is `false`.
+		 */
+		if ( ! $is_visible && apply_filters( 'llms_filter_out_invisible_field', false ) ) {
```
thomasplevy: "filter_out" feels programmy and "hide" seems to be more readibly understandable... I think?

Also, prefixing the filter with "forms" like many of the other hooks in this area of the codebase seems to make it less prone to future collisions.

```suggestion
		if ( ! $is_visible && apply_filters( 'llms_forms_remove_invisible_field', false ) ) {
```

includes/forms/class-llms-form-handler.php
```diff
@@ -319,8 +319,11 @@ public function submit( $posted_data, $location, $args = array() ) {
 		// Determine the user action to perform.
 		$action = get_current_user_id() ? 'update' : 'registration';
 
-		// Load the form.
+		// Load the form, filtering out invisible fields.
+		add_filter( 'llms_filter_out_invisible_field', '__return_true', 999 );
```
thomasplevy: Change explained below.

```suggestion
		add_filter( 'llms_forms_remove_invisible_field', '__return_true', 999 );
```

includes/forms/class-llms-form-handler.php
```diff
@@ -319,8 +319,11 @@ public function submit( $posted_data, $location, $args = array() ) {
...
 		$fields = $this->get_fields( $action, $location, $args );
+		remove_filter( 'llms_filter_out_invisible_field', '__return_true', 999 );
```
thomasplevy: Change explained below

```suggestion
		remove_filter( 'llms_forms_remove_invisible_field', '__return_true', 999 );
```

includes/forms/class-llms-form-handler.php
```diff
@@ -319,8 +319,11 @@ public function submit( $posted_data, $location, $args = array() ) {
 		// Determine the user action to perform.
 		$action = get_current_user_id() ? 'update' : 'registration';
 
-		// Load the form.
+		// Load the form, filtering out invisible fields.
```
thomasplevy: Method needs a version doc

includes/forms/class-llms-form-handler.php
```diff
@@ -333,6 +336,7 @@ public function submit( $posted_data, $location, $args = array() ) {
 	}
 
 	/**
+	 * Form fields submission
```
thomasplevy, line 357: File needs version doc

includes/forms/class-llms-forms.php
```diff
@@ -64,6 +65,7 @@ private function __construct() {
 		add_filter( 'render_block', array( $this, 'render_field_block' ), 10, 2 );
 		add_filter( 'llms_get_form_post', array( $this, 'maybe_load_preview' ) );
 
+		add_filter( 'llms_get_form_blocks', array( $this, 'maybe_add_required_email_field_block' ), 999, 3 );
```
thomasplevy: I think this hook & method should be moved to the `LLMS_Forms_Dynamic_Fields` class

includes/forms/class-llms-form-templates.php
```diff
@@ -121,13 +121,14 @@ private static function find_reusable_block( $field_id ) {
 	 * block modified by legacy options for the given location when `$reusable` is `false`.
 	 *
 	 * @since 5.0.0
+	 * @since [version] Method access set to public.
```
thomasplevy, line 125: file needs version update

includes/forms/class-llms-form-validator.php
```diff
@@ -376,14 +376,15 @@ protected function validate_field_voucher( $posted_value ) {
 	 * Validate submitted field values.
 	 *
 	 * @since 5.0.0
+	 * @since [version] Don't validate form with no user input only if the form is not empty itself (e.g. contains only invisible fields).
```
thomasplevy, line 379: file needs version update

includes/forms/class-llms-forms.php
```diff
@@ -1002,6 +1019,37 @@ public function render_field_block( $html, $block ) {
...
+	 * @param array   $args     Additional arguments passed to the short-circuit filter.
+	 * @return array[]
+	 */
+	public function maybe_add_required_email_field_block( $blocks, $location, $args ) {
+
+		if ( is_user_logged_in() || ! in_array( $location, array( 'checkout', 'registration' ), true ) ) {
+			return $blocks;
+		}
+
+		foreach ( $this->get_field_blocks( $blocks ) as $field_block ) {
```
thomasplevy: We can use `LLMS_Forms::get_field_by( ... )` to pull the email field out of the form by its id and then check if the block is visible. If it returns false, add the field.

If it returns the field and the block isn't visible we should modify the visibility settings to make it visible instead of adding a new field dynamically. My thinking on this is that if someone has made customizations to the email address field (label, description, etc...) we want to ensure those still hold

thomasplevy (changes requested):

includes/shortcodes/class.llms.shortcode.checkout.php
```diff
@@ -291,9 +292,27 @@ protected static function setup_plan_and_form_atts( $plan_id, $atts ) {
...
 
+	/**
+	 * Clean form fields html
+	 *
+	 * Properly detects empty form fields when the html is only composed of blanks and empty paragraphs.
+	 * In this case the form fields html is turned into an empty string.
+	 *
+	 * @since [version]
+	 *
+	 * @param array $fields Form Fields.
```
thomasplevy:
```suggestion
	 * @param array $fields_html Form Fields.
```

thomasplevy (changes requested):

includes/forms/class-llms-form-templates.php
eri-trabiccolo: jeez

includes/forms/class-llms-forms.php
eri-trabiccolo: not as it is right now, because I need $this->is_block_visible()

includes/forms/class-llms-forms.php
eri-trabiccolo: we decided to name it `'llms_forms_remove_invisible_field'`

includes/forms/class-llms-forms.php
eri-trabiccolo: I see what you mean, yeah

includes/forms/class-llms-forms.php
thomasplevy: I missed it

includes/forms/class-llms-forms.php
thomasplevy: mmmm okay... well i think to keep the "dynamic" stuff in a single class makes sense... Make the method public then!

eri-trabiccolo: ok

includes/forms/class-llms-forms.php
thomasplevy: Fixed review :arrow_up:

includes/forms/class-llms-forms.php
eri-trabiccolo: @thomasplevy 
what do you think about the invisible fields handling strategy in https://github.com/gocodebox/lifterlms/pull/1688/commits/7379f4ba582059c90ae560783edc6fbfc96752a7

includes/forms/class-llms-forms.php
```diff
@@ -816,6 +847,91 @@ private function is_block_visible( $block ) {
...
+
+	/**
+	 * Returns a list of block parents plus the block itself in reverse order
+	 *
+	 * @since [version]
+	 *
+	 * @param array   $block      Parsed block array.
+	 * @param array[] $block_list The list of WP Block array `$block` comes from.
+	 * @param int     $iterations Stores the number of iterations.
+	 * @return bool
```
thomasplevy, line 896:
```suggestion
	 * @return array[] List of WP_Block arrays describing or an empty array if $block cannot be found within $block_list
```

includes/forms/class-llms-forms.php
```diff
@@ -816,6 +847,91 @@ private function is_block_visible( $block ) {
...
+						}
+
+						// Merge.
+						return array_merge( $path, $to_merge );
+					}
+				}
+			}
+		}
+
+		return false;
```
thomasplevy:
```suggestion
		return array();
```

I've found a way to break this (cause a fatal error attempting to run `false` into `foreach()`) by passing in a `$block_list` that does not contain `$block` which technically shouldn't happen unless custom code uses this incorrectly *but* if we return `array()` in favor of false it works exactly as expected and `is_block_visible_in_list()` will return false, which I think would be the semantically correct response: the block isn't in the list so it cannot logically be visible within the list...

includes/forms/class-llms-forms.php
thomasplevy: Incorrect return

```suggestion
	 * @return array[] List of WP_Block arrays describing or an empty array if $block cannot be found within $block_list
```

thomasplevy (changes requested):
Overall I think that while the `get_block_path()` method is not the most appealing piece of code ever written it *seems to work*

I couldn't handle reading it alone and then manually testing all the conditions so I wrote a set of tests to test `get_block_path()` and `is_block_visible_in_list()` which I believe cover every possible condition we care about...

I've also made some changes to reduce code complexity as reported by codeclimate and make it a little easier to read (in my opinion, at least)

I haven't thoroughly reviewed the dynamic email/password functionality yet. I'll be doing that next

thomasplevy:
See https://github.com/eri-trabiccolo/lifterlms/pull/2

includes/forms/class-llms-forms-dynamic-fields.php
```diff
@@ -209,6 +210,151 @@ public function modify_account_form( $blocks, $location ) {
...
+		 * Filters the required block fields to add to the form
+		 *
+		 * @since [version]
+		 *
+		 * @param array[] $fields   Array of field_id => block_name required.
+		 * @param array[] $blocks   Array of parsed WP_Block arrays.
+		 * @param string  $location The request form location ID.
+		 * @param array   $args     Additional arguments passed to the short-circuit filter.
+		 */
+		$fields_to_require = apply_filters( 'llms_forms_required_block_fields', $fields_to_require, $blocks, $location, $args );
```
thomasplevy: Reduce method size by moving this to another method

includes/forms/class-llms-forms-dynamic-fields.php
```diff
@@ -209,6 +210,151 @@ public function modify_account_form( $blocks, $location ) {
...
+				unset( $fields_to_require[ $field_id ] );
+				if ( empty( $fields_to_require ) ) { // All the required blocks are present.
+					return $blocks;
+				}
+			}
+		}
+
+		$blocks_to_add = array();
+		foreach ( $fields_to_require as $field_id => $block_to_add ) {
+			$blocks_to_add[] = LLMS_Form_Templates::get_block( $block_to_add, $location, true );
```
thomasplevy: I think that *if* the reusable block exists it should be used here but if it *doesn't* exist it will be created on the page load (and then used again later)

And this just doens't feel like the proper way to handle this to me although it does make it more possible for users to customize the field.

The *best* scenario I think would be to lookup if the reusable exists and then use it if it does and use a non-reusable if it doesn't but 

Modified some things on my PR, thoughts?

thomasplevy (changes requested):

includes/forms/class-llms-forms-dynamic-fields.php
eri-trabiccolo: @thomasplevy 
All you say makes sense. And all I see in your PR makes sense too. I could just merge it now, what do you think? Then I'll give it a better review tomorrow.

thomasplevy:
@eri-trabiccolo with my changes on your fork at https://github.com/eri-trabiccolo/lifterlms/pull/2 I'm very happy with this. I've tested most of the scenarios manually and I think if you could take a review at my changes / suggestions on your fork and merge that in this PR is probably pretty good to ship.

thomasplevy:
Man this was a tough one...

includes/forms/class-llms-forms.php
```diff
@@ -816,6 +847,102 @@ private function is_block_visible( $block ) {
...
+
+	/**
+	 * Returns a list of block parents plus the block itself in reverse order
+	 *
+	 * @since [version]
+	 *
+	 * @param array   $block      Parsed block array.
+	 * @param array[] $block_list The list of WP Block array `$block` comes from.
+	 * @param int     $iterations Stores the number of iterations.
+	 * @return array[] List of WP_Block arrays describing or an empty array if $block cannot be found within $block_list
```
thomasplevy:
```suggestion
	 * @return array[] List of WP_Block arrays or an empty array if $block cannot be found within $block_list
```

tests/phpunit/unit-tests/forms/class-llms-test-form-handler.php
```diff
@@ -324,11 +325,85 @@ public function test_submit_success() {
...
 
+
+	/**
+	 * Test submit password change without providing, or with wrong current password
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function test_submit_password_update_wrong_current_password() {
```
thomasplevy, line 343: I think this test and the next test have transposed names/descriptions

eri-trabiccolo: you're right :D

Issue #1563, opened 2021-03-15
Course Builder: some icons are not visible due to plugin conflicts 


toyinogun:
### Reproduction Steps

There are three plugin conflicts with LifterLMS. The plugins are:

- 301 Redirects PRO
- Qubely
- Qubely PRO

I was able to install the free version of Qubely to confirm this: https://wordpress.org/plugins/qubely/

When any or all of the plugins are installed and activated, some icons in the course builder do not appear.

### Expected Behavior

Course builder Icons should appear

### Actual Behavior

Some Icons do not appear

### Error Messages / Logs

+ Include any relevant error messages or log files
```
N/A

```

### System and Environment Information

<details>
<summary>System Report</summary>

```
System Report:
Wordpress
-------------------------------------------

Home Url: [removed]
Site Url: [removed]
Login Url: [removed]/wp-login.php
Version: 5.6.2
Debug Mode: No
Debug Log: No
Debug Display: Yes
Locale: en_AU
Multisite: No
Page For Posts: Not Set
Page On Front: Home (#95) [[removed]/]
Permalink Structure: /%postname%/
Show On Front: page
Wp Cron: Yes


Settings
-------------------------------------------

Version: 4.19.0
Db Version: 4.19.0
Course Catalog: Courses (#22) [[removed]/courses/]
Membership Catalog: Membership Catalogue (#83966) [[removed]/memberships/]
Student Dashboard: My account (#99) [[removed]/my-account/]
Checkout Page: Purchase (#83967) [[removed]/purchase/]
Course Catalog Per Page: 9
Course Catalog Sorting: menu_order,ASC
Membership Catalog Per Page: 9
Membership Catalog Sorting: menu_order
Site Membership: Not Set
Courses Endpoint: my-courses
Edit Endpoint: edit-account
Lost Password Endpoint: lost-password
Vouchers Endpoint: redeem-voucher
Autogenerate Username: yes
Password Strength Meter: yes
Minimum Password Strength: strong
Terms Required: no
Terms Page: Not Set
Checkout Names: required
Checkout Address: hidden
Checkout Phone: hidden
Checkout Email Confirmation: yes
Open Registration: no
Registration Names: required
Registration Address: hidden
Registration Phone: hidden
Registration Voucher: optional
Registration Email Confirmation: no
Account Names: required
Account Address: optional
Account Phone: optional
Account Email Confirmation: yes
Confirmation Endpoint: confirm-payment
Force Ssl Checkout: no
[...]

PR #1708, opened 2021-07-22
Change admin menu icon from a PNG file to an icon font.

pondermatic:
## Description
If an administrator user edit's their profile and chooses an admin color scheme other than "Default", the luminance of the LifterLMS icon in the WordPress admin menu is different than the dash icons used in other menus items. It is especially noticeable when the scheme is "Light".

This pull request changes the use of a PNG file to an icon font, in the same way that the WordPress [Dashicons](https://developer.wordpress.org/resource/dashicons/) are used.

## How has this been tested?
PHPStorm
caniuse.com
PHPUnit tests

## Screenshots 
with PNG
image

with font
image

## Font
This is the procedure used to create the lifterlms-icons font. Should this go into an existing doc file or a new one?
1. Prepare the LIfterLMS rocket icon.
    1. Make a copy of `.wordpress-org/assets/icon.svg` and name it `lifterlms-icon.svg`.
    1. Edit the copy using Inkscape.
        1. Delete the "Icon" object.
        1. Delete the "background" object inside the "g13" group.
        1. In the "Edit" menu, select "Resize Page to Selection".
        1. Select the "path9" rocket path.
        1. In the Path menu, select "Break Apart".
        1. In the Path menu, select "Exclusion".
        1. Save.
1. Create the font.
    1. Use a web browser to go to https://icomoon.io/app/
    1. Create a new empty set.
    1. Choose "Import to Set" in the set's menu and select the edited icon file.
    1. Use the pencil tool to edit the icon.
        1. Set "Grid" to 14, which is the size of WordPress admin menu icons. (I do not know if this is strictly necessary.)
        1. Use the "Remove Colors" button in the "Color" tool.
        1. Set the Tags and Names to "lifterlms-icon-monochrome". (I do not know if this is strictly necessary.)
    1. Edit the set properties.
        1. Set the Name to "lifterlms-icons".
        1. Click the "Remove All Colors" button.
    1. Click on the "Select" tool and select the LifterLMS rocket icon.
    1. On the bottom of the page, click the "Generate Font" button.
    1. Set the code to "1f680", which is the Unicode number for the rocket emoji.
    1. Click the gear icon next to the Download button.
        1. Set the Font Name to "lifterlms-icons".
        1. Check "Include metadata in fonts".
        1. Set Metadata.  
            URL: https://lifterlms.com/
            Description: LifterLMS icons
            Copyright: 2021
            License: GPLv3
            License URL: https://www.gnu.org/licenses/gpl-3.0.html
        1. Set Version.
    1. Click the "Download" button.
    1. Extract the font files to the `assets/fonts` directory.
    1. Base64 encode the WOFF file and paste the output into `assets/scss/admin/_wp-menu.scss`.
        1. file_put_contents('lifterlms-icons.woff.b64', base64_encode(file_get_contents('lifterlms-icons.woff')));
    1. Compile SCSS to CSS.
        1. npm run build:styles

## Checklist:


thomasplevy:
Hey @pondermatic!

Thanks for the PR. I wasn't aware of this bug and I'm glad I am now!

I think this is a working solution to the proble. However, including an entire font set for a single icon feels a little heavy to me.

Especially considering that SVG is generally available across (most) browsers these days (with some minor and in my opinion acceptable inconsistencies on ie11) and as WP moves further down the React/JS path we'll see more and more SVG icons used in favor of icon fonts. Quite a few (possibly all?) of the icons used in the block editor are pure SVG now (via React components), example: https://github.com/WordPress/gutenberg/blob/trunk/packages/icons/src/library/wordpress.js

All of this to say I see a "better" solution here to follow suit with the WP core and switch this icon out to a true SVG and skip the "weight" of the icon font.

Furthermore, I know we'll be adding more custom icons in various places (see https://github.com/gocodebox/lifterlms/issues/1563). My plan here would be to take the icons found in lifterlms-blocks and move them into a package in the core monorepo so these icons can be imported. Basically following the pattern put forth by Gutenberg.

All of this is a long way of saying that this *fixes the problem* but not in the way I'd like to see it fixed now that I know it exists.

Also: e2e tests are failing because of this WP core issue: https://github.com/gocodebox/lifterlms/issues/1713

So thank you very much but I'm going to reject this implementation for the above reasons.

I've opened https://github.com/gocodebox/lifterlms/issues/1714 in order to ensure we track this issue as we should fix this.

Also: The solution I'd like to see here would be to pull the existing black/white icon we use in lifterlms-blocks (https://github.com/gocodebox/lifterlms-blocks/blob/dev/src/js/icons/lifterlms-icon.js) and pass that as an SVG to `add_menu_icon()` -- see note in the docs:

> Pass a base64-encoded SVG using a data URI, which will be colored to match the color scheme. This should begin with 'data:image/svg+xml;base64,'.

Source: https://developer.wordpress.org/reference/functions/add_menu_page/

pondermatic:
My first attempt at solving this issue was to pass a base 64 encoded version of the icon to `add_menu_icon()`. It was super easy and I _almost_ submitted a PR with it. Then I realized that the icon's original fill color is displayed until the WordPress SVG painter JavaScript changes the fill color to the value in the selected admin color scheme. The greater the difference in luminance from the original fill color and the scheme fill color, the more noticeable the change is.

The admin color scheme fill colors are admittedly similar to each other. The time it takes to paint the SVG icon is a minor issue. I was curious how WordPress Dashicons work and used this as a learning opportunity.

Adding a bunch of files to the project does seem heavy for such a small improvement. However, most browsers will use the encoded WOFF font in the CSS file and won't request the other font files. It was a heavy challenge figuring all of this out, but the web browsers wouldn't have any extra work to do.

thomasplevy:
While I agree that the browsers don't have to do anymore work, the plugin size increases and we also need to include a new (pretty complicated) piece of documentation on how to arrive at this should we decide to add more icons in this way.

The "flash" of incorrectly colored SVG is, I think, a better solution than including a bunch of new files, plus additional CSS, to solve what we can solve with a bit of SVG code added to a single file.

Also, just to see what the precedent here is, both WooCommerce and Yoast accept this flash:

https://github.com/Yoast/wordpress-seo/blob/96f6b382d6b519d0e35002ee43221db01c8eef0d/admin/menu/class-admin-menu.php#L53-L61

https://github.com/woocommerce/woocommerce/blob/e382e03f585c0eb4b84e25a2c14fdf6af364a23a/includes/admin/class-wc-admin-menus.php#L58-L64

Just because these "giants" do it this way doesn't mean we *must* follow suit but I feel like this is a simpler solution with a very minor downside that many people will likely never even notice.

And also, let's be real, how many users actually switch their theme anyway?

I am certain that this took quite a lot of effort to put together and that isn't at all lost on me but I just don't like the solution when there's a far simpler solution available to us.

:-(

pondermatic:
I agree with your points. The SVG is the easiest solution and is good enough. This was mostly a learning exercise.

I run several LifterLMS sites, so I choose a different admin color scheme on each to help me remember which one I'm working on.

PR #1709, opened 2021-07-22
Addon uninstall method

thomasplevy:
## Description

Adds an "uninstall" method to the `LLMS_Add_On` class to enable programmatic uninstallation of add-on plugins and themes.

## How has this been tested?

+ Manually
+ Unit tests are not easy to write for this until we have the CLI finished (and included) in the core. Then we can bootstrap an add-on plugin into the test environment to do more thorough testing against. It's technically possible without the CLI but it'll be a lot easier to wait until the CLI is done.

## Types of changes

## Checklist:

### DIFF
--- a/includes/models/model.llms.add-on.php
+++ b/includes/models/model.llms.add-on.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Models/Classes
  *
  * @since 3.22.0
- * @version 4.21.3
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -448,6 +448,40 @@ class LLMS_Add_On {
 		}
 
 		return false;
+
+	}
+
+	/**
+	 * Uninstall an add-on
+	 *
+	 * @since [version]
+	 *
+	 * @return string|WP_Error
+	 */
+	public function uninstall() {
+
+		$ret = sprintf( __( '%s was successfully uninstalled.', 'lifterlms' ), $this->get( 'title' ) );
+
+		if ( ! $this->is_installed() ) {
+			$ret = new WP_Error( 'not-installed', sprintf( __( '%s is not installed.', 'lifterlms' ), $this->get( 'title' ) ) );
+		} elseif ( $this->is_active() ) {
+			$ret = new WP_Error( 'uninstall-active', sprintf( __( '%s is active and cannot be uninstalled.', 'lifterlms' ), $this->get( 'title' ) ) );
+		} else {
+
+			$type = $this->get_type();
+			$file = $this->get( 'update_file' );
+			if ( 'plugin' === $type ) {
+				uninstall_plugin( $file );
+				$del = delete_plugins( array( $file ) );
+				$ret = is_wp_error( $del ) ? $del : $ret;
+			} elseif ( 'theme' === $type ) {
+				$del = delete_theme( $file );
+				$ret = is_wp_error( $del ) ? $del : $ret;
+			}
+
+		}
+
+		return $ret;
 
 	}
 
--- a/includes/models/model.llms.add-on.php
+++ b/includes/models/model.llms.add-on.php
@@ -452,35 +452,58 @@ class LLMS_Add_On {
 	}
 
 	/**
-	 * Uninstall an add-on and permanently deletes its files
+	 * Verifies the add-on can be uninstalled, and performs the uninstall (permanently deleting its files)
 	 *
 	 * @since [version]
 	 *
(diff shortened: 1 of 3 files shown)


includes/models/model.llms.add-on.php
```diff
@@ -451,4 +451,37 @@ private function lookup_add_on( $lookup_key, $lookup_val ) {
...
+	 */
+	public function uninstall() {
+
+		$ret = sprintf( __( '%s was successfully uninstalled.', 'lifterlms' ), $this->get( 'title' ) );
+
+		if ( ! $this->is_installed() ) {
+			$ret = new WP_Error( 'not-installed', sprintf( __( '%s is not installed.', 'lifterlms' ), $this->get( 'title' ) ) );
+		} elseif ( $this->is_active() ) {
+			$ret = new WP_Error( 'uninstall-active', sprintf( __( '%s is active and cannot be uninstalled.', 'lifterlms' ), $this->get( 'title' ) ) );
+		} else {
```
eri-trabiccolo: missing `Translators:...`

includes/models/model.llms.add-on.php
```diff
@@ -451,4 +451,37 @@ private function lookup_add_on( $lookup_key, $lookup_val ) {
 
 	}
 
+	/**
+	 * Uninstall an add-on
```
eri-trabiccolo:
```suggestion
	 * Uninstall an add-on and permanently deletes its files
```

includes/models/model.llms.add-on.php
```diff
@@ -451,4 +451,37 @@ private function lookup_add_on( $lookup_key, $lookup_val ) {
...
+			$type = $this->get_type();
+			$file = $this->get( 'update_file' );
+			if ( 'plugin' === $type ) {
+				uninstall_plugin( $file );
+				$del = delete_plugins( array( $file ) );
+				$ret = is_wp_error( $del ) ? $del : $ret;
+			} elseif ( 'theme' === $type ) {
+				$del = delete_theme( $file );
+				$ret = is_wp_error( $del ) ? $del : $ret;
+			}
```
eri-trabiccolo:
```suggestion
			$file = $this->get( 'update_file' );
			$del  = null;
			if ( 'plugin' === $type ) {
				uninstall_plugin( $file );
				$del = delete_plugins( array( $file ) );
			} elseif ( 'theme' === $type ) {
				$del = delete_theme( $file );
			}
			$ret = is_wp_error( $del ) ? $del : $ret;
```
quite golf

eri-trabiccolo (changes requested):

includes/models/model.llms.add-on.php
thomasplevy: fixed

thomasplevy:
@eri-trabiccolo I noticed a problem (we both missed) when reviewing your suggestions. Quick re-review when you can please <3

includes/models/model.llms.add-on.php
```diff
@@ -452,35 +452,58 @@ private function lookup_add_on( $lookup_key, $lookup_val ) {
...
+		$del  = null;
+		if ( 'plugin' === $type ) {
+			uninstall_plugin( $file );
+			$del = delete_plugins( array( $file ) );
+		} elseif ( 'theme' === $type ) {
+			$del = delete_theme( $file );
 		} else {
+			// Translators: %s = add-on type.
+			return new WP_Error( 'uninstall-invalid-type', sprintf( __( 'Cannot uninstall "%s" type add-ons.', 'lifterlms' ), $type ) );
+		}
```
eri-trabiccolo: Alright, I thought about that but I kinda trusted that you didn't take care of this 3rd case (invalid type) because it couldn't happen.
About the code itself, what do you think if we early bail when the type is not a valid one, to avoid if/elseif/else that are a bit ugly to me. I mean make it like:
```php
		$type = $this->get_type();
		if ( ! in_array( $type, array( 'theme', 'plugin' ), true ) ) {
			// Translators: %s = add-on type.
			return new WP_Error( 'uninstall-invalid-type', sprintf( __( 'Cannot uninstall "%s" type add-ons.', 'lifterlms' ), $type ) );
		}

		$file = $this->get( 'update_file' );

		if ( 'plugin' === $type ) {
			uninstall_plugin( $file );
			$del = delete_plugins( array( $file ) );
		} else {
			$del = delete_theme( $file );
		} 

```

eri-trabiccolo (changes requested):
Fine for me, proposed a maybe simplified code?

PR #1719, opened 2021-07-28
Fixed get_the_excerpt filter hook missing second param

eri-trabiccolo:
## Description
We need to pass a second parameter to `get_the_excerpt` filter hook according to the [WP core filter definition.](https://developer.wordpress.org/reference/hooks/get_the_excerpt/)
The second parameter (the `$post`) has been added since wp 4.5.0.
Not passing it can cause fatal errors, as reported [here](https://github.com/gocodebox/lifterlms-rest/issues/233#issuecomment-888303723) - a conflict with the latest version of the Divi theme 4.10.0
So this...
Fixes https://github.com/gocodebox/lifterlms-rest/issues/233

## How has this been tested?
Manually

## Types of changes

## Checklist:

### DIFF
--- a/includes/abstracts/abstract.llms.post.model.php
+++ b/includes/abstracts/abstract.llms.post.model.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Abstracts/Classes
  *
  * @since 3.0.0
- * @since [version]
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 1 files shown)


includes/abstracts/abstract.llms.post.model.php
```diff
@@ -5,7 +5,7 @@
  * @package LifterLMS/Abstracts/Classes
  *
  * @since 3.0.0
- * @version 4.10.0
+ * @since [version]
```
```suggestion
 * @version [version]
```

thomasplevy (changes requested):

PR #1725, opened 2021-08-04
Adds a filter on the list of included libraries

thomasplevy:
## Description

Required by https://github.com/gocodebox/lifterlms-helper/pull/29

This fixes an issue encountered in phpunit test suites for our library add-ons (REST, Blocks, and Helper). Since the LifterLMS core is required by the add-ons and the PHP files are included slightly differently in the test suites than how it would be included in an actual WP site the tests that evaluate whether or not the lib should be loaded (because the plugin doesn't exist) will always show the library should be loaded in the phpunit environment.

This adds a filter which will allow us to force the library to not be loaded during testing.

The filter can also be leveraged by 3rd parties to disable the libraries in a different way (we currently have constants to allow this also so, slightly redundant but...)

## How has this been tested?

+ Manually
+ Unit tests (in the helper)

## Types of changes

+ Bugfix

## Checklist:


eri-trabiccolo (changes requested):
The actual change is good to me but the PR itself is a little bit messy meaning that it includes commits which do not pertain to it.

thomasplevy:
woops, i realized i hand't pushed this at the very end of the day and I had staged commits on dev building the next release. I'll fix it up

PR #1728, opened 2021-08-05
Database upgrade refactor

thomasplevy:
## Description

Refactors the database upgrade / migration routine functionality found in `LLMS_Install`.

+ Moves update functionality into a separate class: `LLMS_DB_Upgrader`
+ Moves the update callback list from a class property to a list found in the `includes/schemas` directory.
+ Provides the ability for each update to be flagged as "manual" or "auto".
    + A manual update will work as updates have worked since their inception in LifterLMS: an admin notice is displayed and the user must press a button to start the update. 
    + An auto update will happen silently in the background without user interaction.

The goal is to reduce the *fear* instilled in end users by telling them their database is going to be modified when minor updates take place. For example the 5.0.0 upgrade changes 3 legacy options from autoload -> yes to autoload -> no. The impact of this update failing is *very low* and there's really no need to instill any amount of worry in a user for this kind of update. Similarly we'll be adding an update for a forthcoming release which will allow new users to have a new feature enabled by default but existing users (who upgrade) will have it disabled by default and can opt-in as desired. This is technically a "database update" but it isn't an upgrade that requires user's to backup their database since we'll just be running `update_option( ... )`

## How has this been tested?

+ New phpunit tests
+ Manually (partially) 

~I'm opening the PR but I'm going to continue testing and probably will write some more tests today.~

Edit: looks good to me with more testing.

To test real world updates, I reset the DB and LLMS version via CLI and then visited the admin panel:

`wp option set lifterlms_db_version 4.99.0 && wp option set lifterlms_current_version 4.99.0`

The 5.0.0 update is pretty low-effect so running that multiple times to test is pretty "safe"

## Types of changes

+ Update / feature

## Checklist:


includes/class-llms-db-ugrader.php
```diff
@@ -0,0 +1,313 @@
+<?php
+/**
+ * LLMS_DB_Upgrader class file
+ *
+ * @package LifterLMS/Classes
+ *
+ * @since 4.4.0
+ * @version 4.9.0
```
eri-trabiccolo:
```suggestion
 * @since [version]
 * @version [version]
```

includes/class-llms-db-ugrader.php
```diff
@@ -0,0 +1,313 @@
...
+	 *
+	 * If the update group cannot be auto-updated the following admin notices will be included:
+	 * + The "update started" notice will be immediately displayed/added.
+	 * + The "update complete" notice will be added to the end of the queue (and then displayed when the update is complete).
+	 *
+	 * @since [version]
+	 *
+	 * @see [Reference]
+	 * @link [URL]
+	 *
```
eri-trabiccolo:
```suggestion
```

includes/class-llms-db-ugrader.php
```diff
@@ -0,0 +1,313 @@
...
+	 * @return [type] [description]
```
eri-trabiccolo:
```suggestion
	 * @return void
```

includes/class-llms-db-ugrader.php
```diff
@@ -0,0 +1,313 @@
...
+			'bg-db-update-complete',
+			__( 'The LifterLMS database update is complete.', 'lifterlms' ),
+			array(
+				'dismissible'      => true,
+				'dismiss_for_days' => 0,
+			)
+		);
+
+	}
+	/**
```
eri-trabiccolo:
```suggestion
	}
	
	/**
```

includes/class.llms.install.php
```diff
@@ -613,12 +496,13 @@ private static function get_schema() {
 	 *
 	 * @since 3.4.3
 	 * @since 3.6.0 Unknown.
+	 * @since [version] Use `LLMS_PLUGIN_DIR` to include required class file.
 	 *
 	 * @return void
 	 */
 	public static function init_background_updater() {
 
-		include_once dirname( __FILE__ ) . '/class.llms.background.updater.php';
```
eri-trabiccolo, line 621: and btw `dirname( __FILE__ )` is equivalent to `__DIR__` :D
https://www.php.net/manual/en/language.constants.magic.php

includes/schemas/llms-db-updates.php
```diff
@@ -0,0 +1,142 @@
...
+	'5.0.0'  => array(
+		'type'    => 'manual',
+		'updates' => array(
+			'llms_update_500_legacy_options_autoload_off',
+			'llms_update_500_update_db_version',
+			'llms_update_500_add_admin_notice',
+		),
+	),
+
+);
```
eri-trabiccolo, line 148: if we only used namespaces from the start...
can we start from now anyways so that we'll have:
```
	'5.2.0'  => array(
		'type'    => 'manual',
		'updates' => array(
			'llms_update_something',
			'llms_update_update_db_version',
		),
	),
```
The logging then could take the key + the callable name like:
5.2.0 - llms_update_something

eri-trabiccolo (changes requested):
docs.
I also manually tested it.
Looks good.

includes/schemas/llms-db-updates.php
thomasplevy: I'm not sure what you mean exactly... You mean namespace the update functions so that we don't have to prefix them with a version number?

Eg

```
namespace LifterLMS\Updates\520
// does a db update of some kind
function something() { ... }

// update db version
function update_db_version() {
   update_option( ..., '5.2.0' );
}
```

??
`

eri-trabiccolo: yeah I mean this.

includes/class.llms.install.php
thomasplevy: you're giving me golf advice on code I deleted?

includes/schemas/llms-db-updates.php
thomasplevy: I don't want to rewrite and retest all this. Merging as is.

includes/class.llms.install.php
eri-trabiccolo: I bumped into it and wanted to _share_ this "thought" as we use the longer version (and I bet slower) in pretty much all our repo.

Issue #1561, opened 2021-03-12
Events: client and server event's dates are not saved with the same timezone.


eri-trabiccolo:
### Reproduction Steps

+ Set WP timezone to something different than php local time.
+ Produce a server side event, like sign-in/sign-out -> check the db entry date: set in the wp timezone
+ Now produce a client side event - e.g. start a video (with the advanced-videos add-on) - and persist it in the db (e.g.   exit the page).

### Expected Behavior
+ Both client and server side dates should be saved using the same time zone.

### Actual Behavior
+ Client side events date is saved using php local time (php [`date()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/class-llms-events.php#L152) to which we pass a [GMT timestamp](https://github.com/gocodebox/lifterlms/blob/4.19.0/assets/js/app/llms-tracking.js#L141)) while server side events are saved using [wp local time](https://github.com/gocodebox/lifterlms/blob/4.19.0/includes/abstracts/llms.abstract.database.store.php#L115)
The difference can be quite important as the first is usually a server setting the users are not going to change.

At the moment this impacts on the [idle sessions](https://github.com/gocodebox/lifterlms/blob/4.19.0/includes/class-llms-sessions.php#L214) (that need to be closed) only - as far as I can see.

I think we should definitely avoid storing the date using php local time, we need to decide whether we want to store it in GMT (`gmdate()`) or using local wp timezone. 
Historically we save dates using wp timezone, but I'd rather prefer storing dates in GMT.
We should probably build a db update routine too to fix server side stored event dates...(?)

### Error Messages / Logs

```
n/a
```

### System and Environment Information

```
LifterLMS 4.19.0
WordPress 5.7.0
```

### This issue has been recreated:
+ [x] Locally
+ [ ] On a staging site
+ [x] On a production website
+ [x] With only LifterLMS and a default theme

### Browser, Device, and Operating System Information

n/a

### Related User Information

HS-158540

thomasplevy:
@eri-trabiccolo is this and #1559 the same issue?

PR #1739, opened 2021-08-11
Use the local time zone instead of UTC for `last_seen` in `get_data`.

### LINKED ISSUE #1738: "Last Seen" date uses UTC instead of local time zone
### Reproduction Steps

1. Set the WordPress time zone in the general settings to UTC+12.
1. View the LifterLMS reporting page that lists students. `/wp-admin/admin.php?page=llms-reporting`
1. Set the WordPress time zone in the general settings to UTC-12.
1. View the LifterLMS reporting page that lists students. `/wp-admin/admin.php?page=llms-reporting`

### Expected Behavior

The "Last Seen" date should be different between the two page views.

### Actual Behavior

The "Last Seen" date is the same between the two page views.

This issue has been recreated:
+ [x] Locally
+ [ ] On a staging site
+ [ ] On a production website

pondermatic:
## Description
Changed `LLMS_Table_Students::get_data()` to use `wp_date()` instead of `date_i18n()` so that `last_seen` is returned in the server's local time zone instead of UTC.

Fixes #1738.

## How has this been tested?
Added `LLMS_Test_Table_Students::test_get_data_last_seen()`, which tests:
1. user has never been seen
2. user was seen yesterday
3. user was seen today

## Types of changes
This function is only used when displaying the "Last Seen" column in the LifterLMS students reporting page.

## Checklist:
- [ ] My code passes all existing automated tests. Unrelated tests are failing. 

### DIFF
--- a/includes/admin/reporting/tables/llms.table.students.php
+++ b/includes/admin/reporting/tables/llms.table.students.php
@@ -19,6 +19,7 @@ defined( 'ABSPATH' ) || exit;
  * @since 3.36.0 Add "Last Seen" column.
  * @since 3.36.1 Fixed "Last Seen" column displaying wrong date when the student last login date was saved as timestamp.
  * @since 3.37.2 The post filter on the students table now limits post results based on instructor access.
+ * @since [version] Use the local time zone instead of UTC for `last_seen` in `get_data`.
  */
 class LLMS_Table_Students extends LLMS_Admin_Table {
 
@@ -108,6 +109,7 @@ class LLMS_Table_Students extends LLMS_Admin_Table {
 	 * @since 3.36.0 Added "Last Seen" column.
 	 * @since 3.36.1 Fixed "Last Seen" column displaying wrong date when the student last login date was saved as timestamp.
 	 * @since 4.7.0 Speed up the query used to retrieve the last seen column by avoiding the found rows calculation.
+	 * @since [version] Use the local time zone instead of UTC for `last_seen`.
 	 *
 	 * @param string       $key     The column id / key.
 	 * @param LLMS_Student $student Instance of the LLMS_Student.
@@ -186,7 +188,12 @@ class LLMS_Table_Students extends LLMS_Admin_Table {
 					$value = $student->get( 'last_login' );
 				}
 
-				$value = $value ? date_i18n( get_option( 'date_format' ), is_numeric( $value ) ? $value : strtotime( $value ) ) : '&ndash;';
+				if ( $value ) {
+					$timestamp = is_numeric( $value ) ? $value : strtotime( $value );
+					$value     = wp_date( get_option( 'date_format' ), $timestamp );
+				} else {
+					$value = '&ndash;';
+				}
 
 				break;
 
--- a/tests/phpunit/unit-tests/tables/class-llms-test-table-students.php
+++ b/tests/phpunit/unit-tests/tables/class-llms-test-table-students.php
@@ -9,6 +9,7 @@
  * @since 3.28.0
  * @since 3.36.0 Add "last_seen" col.
  * @since 3.37.15 Add additional tests for new conditions in `generate_export_file()` method.
+ * @since [version] Add `test_get_data_last_seen`.
  */
 class LLMS_Test_Table_Students extends LLMS_UnitTestCase {
 
@@ -25,6 +26,40 @@ class LLMS_Test_Table_Students extends LLMS_UnitTestCase {
 
 	}
 
+	/**
+	 * Test `LLMS_Table_Students::get_data( 'last_seen' )` and the effect of different WordPress time zones.
+	 *
+	 * @since [version]
+	 * @todo  Test using the last row in the `lifterlms_events` table.
+	 */
+	public function test_get_data_last_seen() {
+		# Set the current date and time in the UTC time zone.
+		llms_tests_mock_current_time( '2021-01-20 00:00:00' );
+
+		$table = new LLMS_Table_Students();
+		update_option( 'date_format', 'Y-m-d' );
+
+		# Register a new student.
+		$user_id = $this->factory->user->create( array( 'role' => 'student' ) );
+		$student = llms_get_student( $user_id );
+
+		# The student has never been seen.
+		$last_seen = $table->get_data( 'last_seen', $student );
+		$this->assertEquals( '&ndash;', $last_seen, 'User has never been seen.' );
+
+		# Log the user in.
+		llms_set_user_login_time( $student->get( 'user_login' ), $student->get_user() );
+
+		# Set the server's local time zone to before UTC.
+		update_option( 'timezone_string', 'America/Chicago' );
+		$last_seen = $table->get_data( 'last_seen', $student );
+		$this->assertEquals( '2021-01-19', $last_seen, 'User was seen yesterday.' );
+
+		# Set the server's local time zone to after UTC.
+		update_option( 'timezone_string', 'Asia/Tokyo' );
+		$last_seen = $table->get_data( 'last_seen', $student );
+		$this->assertEquals( '2021-01-20', $last_seen, 'User was seen today.' );
+	}
 
 	/**
 	 * test the get_export() method.


includes/admin/reporting/tables/llms.table.students.php
```diff
@@ -186,7 +188,12 @@ public function get_data( $key, $student ) {
 					$value = $student->get( 'last_login' );
 				}
 
-				$value = $value ? date_i18n( get_option( 'date_format' ), is_numeric( $value ) ? $value : strtotime( $value ) ) : '&ndash;';
+				if ( $value ) {
+					$timestamp = is_numeric( $value ) ? $value : strtotime( $value );
+					$value     = wp_date( get_option( 'date_format' ), $timestamp );
```
eri-trabiccolo, line 193: @pondermatic 
I'm not convinced about this.

There is a problem, but I'm not convinced it's the one you're describing and solving here.

The thing is that that column basically shows this value:
https://github.com/gocodebox/lifterlms/blob/5.2.0/includes/abstracts/llms.abstract.database.store.php#L115
which is basically `current_time()` in `'mysql'` format  plus the wp timezone offset:
https://github.com/gocodebox/lifterlms/blob/dae9300e19b5cb2cd48458e88cd8348ca1a2274b/includes/llms.functions.core.php#L94
So here we're replacing [date_i18n()](https://developer.wordpress.org/reference/functions/date_i18n/) wit [wp_date()](https://developer.wordpress.org/reference/functions/wp_date/).
The first expects a timestamp with offset (timezone), while the second expects a timestamp without timezone.
The thing is that our saved value is a timestamp with offset, but the offset of course is calculated with the wp timezone at the moment the entry is created.
So using `wp_date` here is not an option, I think...

I have to think a little more about this but I'm not sure there's a solution, if not converting (with a db update) all those saved dates to UTC (with a guess) and then go with `wp_date`

pondermatic: You are correct @eri-trabiccolo, [`LLMS_Event` dates](https://github.com/gocodebox/lifterlms/blob/5.2.0/includes/abstracts/llms.abstract.database.store.php#L115), stored in the `wp_lifterlms_events` table, are relative to the local time zone. The [`last_login`](https://github.com/gocodebox/lifterlms/blob/5.2.0/includes/functions/llms.functions.person.php#L487) time, stored in `wp_usermeta` with the `llms_last_login` key, is relative to the UTC time zone.

I'll change `LLMS_Table_Students::get_data()` to handle both values and also add a test using an event.

eri-trabiccolo: @pondermatic 
but actually the `last_login` time is in local time too, no?
https://github.com/gocodebox/lifterlms/blob/5.2.0/includes/abstracts/llms.abstract.database.store.php#L115
https://github.com/gocodebox/lifterlms/blob/5.2.0/includes/functions/llms.functions.person.php#L487
both are using `llms_current_time('mysql');`

pondermatic:
It turns out @eri-trabiccolo that **some** LifterLMS events use local time while others use UTC.

* When `LLMS_Event` is [constructed](https://github.com/gocodebox/lifterlms/blob/5.2.0/includes/abstracts/llms.abstract.database.store.php#L115), `llms_current_time()` is used to get the local time.
* When this flow is used, the event date is set with a Unix timestamp, which is always relative to UTC.
  1. [`LLMS_Events->store_cookie()`](https://github.com/gocodebox/lifterlms/blob/f1dcd79d813e1996b819f9f044ccb0f1ac67a8db/includes/class-llms-events.php#L435) or [`LLMS_AJAX_Handler->persist_tracking_events()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/class.llms.ajax.handler.php#L1435)
  1. [`LLMS_Events->store_tracking_events()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/class-llms-events.php#L460)
  1. [`LLMS_Events->prepare_event()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/class-llms-events.php#L152)
  1. [`LLMS_Events->record()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/class-llms-events.php#L227)
  1. [`LLMS_Event->setup()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/abstracts/llms.abstract.database.store.php#L231)

@thomasplevy, should we:
1. Change [`LLMS_Abstract_Database_Store->__construct()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/abstracts/llms.abstract.database.store.php#L115) to use UTC time. This would be consistent with how WordPress tends to store dates with UTC times. However, this would have a large ripple effect and might break something else.  
  **or**
2. Change [`LLMS_Events->prepare_event()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/class-llms-events.php#L152) to set the date in local time. This would make all (or at least more) event dates in local time.

eri-trabiccolo:
@pondermatic 

> It turns out @eri-trabiccolo that **some** LifterLMS events use local time while others use UTC.

Yup, you're right, sorry, I forgot that the "last seen" takes the last tracked event date, not only the last login event date.

> * When `LLMS_Event` is [constructed](https://github.com/gocodebox/lifterlms/blob/5.2.0/includes/abstracts/llms.abstract.database.store.php#L115), `llms_current_time()` is used to get the local time.

Yes, you're right I forgot that the 

> * When this flow is used, the event date is set with a Unix timestamp, which is always relative to UTC.
>   1. [`LLMS_Events->store_cookie()`](https://github.com/gocodebox/lifterlms/blob/f1dcd79d813e1996b819f9f044ccb0f1ac67a8db/includes/class-llms-events.php#L435) or [`LLMS_AJAX_Handler->persist_tracking_events()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/class.llms.ajax.handler.php#L1435)
>   2. [`LLMS_Events->store_tracking_events()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/class-llms-events.php#L460)
>   3. [`LLMS_Events->prepare_event()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/class-llms-events.php#L152)
>   4. [`LLMS_Events->record()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/class-llms-events.php#L227)
>   5. [`LLMS_Event->setup()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/abstracts/llms.abstract.database.store.php#L231)

Quite, client side generated events are referring to the php local time (`date()` depends on the PHP default timezone).
Basically you bumped into this https://github.com/gocodebox/lifterlms/issues/1561

> @thomasplevy, should we:
> 
> 1. Change [`LLMS_Abstract_Database_Store->__construct()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/abstracts/llms.abstract.database.store.php#L115) to use UTC time. This would be consistent with how WordPress tends to store dates with UTC times. However, this would have a large ripple effect and might break something else.

my 2cents: we should progressively, if we can, migrate all our stored dates to UTC.

>    **or**
> 2. Change [`LLMS_Events->prepare_event()`](https://github.com/gocodebox/lifterlms/blob/2d86dd2c67042b1f3bfb9a63d0b7b2089e8eaa28/includes/class-llms-events.php#L152) to set the date in local time. This would make all (or at least more) event dates in local time.

thomasplevy:
In light of the ripple effect and the very large undertaking that I don't have the mental capacity to oversee with how dates are stored and handled we're going to reprioritize this for future development when we have more internal capacity to properly determine the date overhaul that needs to take place across LifterLMS. There's a lot to consider and think about and I personally cannot manage to figure it all out right now.

I'm going to start putting together a "master" issue organzing all the date issues so we can prioritize this as a larger project sometime in the future.

PR #1744, opened 2021-08-12
Adjust methodology for determining when a limited length recurring order ends

thomasplevy:
## Description

Fixes https://github.com/gocodebox/private-issues/issues/37

## How has this been tested?

+ New tests

## Types of changes

+ Bug fix

## Checklist:


thomasplevy:
@eri-trabiccolo Opening this for your thoughts on my strategy before I get too deep in the weeds.

Essentially the issue in https://github.com/gocodebox/private-issues/issues/37 is that we determine when a limited-length order ends based on a calculated end dates. This works well in *most* scenarios but it creates an issue if payments are missed. When a payment (or two or three) is/are missed the plan will end before all of the intended recurring payments are made.

My goal would be to remove the used of the `billing_end_date` property entirely and instead "end" an order like this after the required number of payments are successful.

The next step would be to remove the logic related to the billing end date

https://github.com/gocodebox/lifterlms/blob/ba7c9a00115103ee1d401a24fe4114b3050b1a8d/includes/models/model.llms.order.php#L358-L360

And replace it with a check like this:

```php
$remaining = $this->get_remaining_payments();
if ( 0 === $remaining ) {
   return '';
}
```

We could likely put that up at the top of the `calculate_next_payment_date()` method, saving the rest of the calculations for orders that require it.

This does feel like a relatively *large* change but realistically it's just a minor improvement to handle the scenario encountered in https://github.com/gocodebox/private-issues/issues/37

Also to consider: I'm including refunded transactions to count towards the number of payments processed. My logic here is that a refunded payment *was made* and was refunded for one reason or another but that doesn't mean that we add more payments to the schedule, right? EG: 5 payments are due, I make all 5 and the plan ends and then we refund the last payment that doesn't mean that we should start the plan back up and draw another payment... Right?

eri-trabiccolo:
It sounds good .

> Also to consider: I'm including refunded transactions to count towards the number of payments processed. My logic here is that a refunded payment was made and was refunded for one reason or another but that doesn't mean that we add more payments to the schedule, right? EG: 5 payments are due, I make all 5 and the plan ends and then we refund the last payment that doesn't mean that we should start the plan back up and draw another payment... Right?

Ideally you're right, but in practice in the past we had to deal with (edge cases) unwanted charges that had to be refunded, and those should not be counted. I think it's basically impossible for us to determine whether a refunded transaction has to be counted or not. I think we can assume it has to be counted by default but also allow to remove it from the count.
Like with a checkbox next to a refunded transaction in the order's transactions table?

thomasplevy:
@eri-trabiccolo yea... managing this in this way feels very complicated though, especially since this only applies to payment plans with a limited subscription...

Mostly my concern is that the interface for displaying transactions is already a bit "cramped" as there's a lot of information, adding a new checkbox means adding more and then, additionally, trying to explain *why that checkbox exists and what it does* in that limited amount of space feels like we'll end up just creating confusion...

Alternatively, I'm wondering if we could adjust the interface in the "Payment Information" column to add output of the number of remaining payments and allow that to be editable. 

Quick mockup:

Screenshot_2021-08-13_10-17-00

~Clicking the edit pencil will convert the the "3" to a text input and allow them to increase (or decrease) the number of remaining payments.~

Maybe it should actually be two buttons (a plus and a minus) -- clicking either will pop a modal prompting the user to enter an (optional) note as to the reason why. On save the orders "plan_length" will be adjusted and payments will be cancelled / rescehulded as needed. It should *not* show a remove payments if they've already made all the payments on the plan

This would let *refunds* always count towards the number of remaining payments but if they decide a refund shouldn't count they can add an extra payment onto the plan

This is a lot more code than your solution but I think it'll be easier to explain what's going on than with a checkbox...

What do you think?

thomasplevy:
I'm also thinking that we should create a migration that will attempt to identify orders that may have "missed" payments and provide users with an option to restart those subscriptions if they want... I'm thinking it'll be an automigration that will run automatically and then pop a notice with a list of affected orders for the user to review... Thoughts on this?

eri-trabiccolo:
> I'm also thinking that we should create a migration that will attempt to identify orders that may have "missed" payments and provide users with an option to restart those subscriptions if they want... I'm thinking it'll be an automigration that will run automatically and then pop a notice with a list of affected orders for the user to review... Thoughts on this?

Like looking for expired orders with expired payment plans which haven't reached the expected number of payments?
Mmm and what would "restart subscription" mean? Can this be abused?

> Maybe it should actually be two buttons (a plus and a minus) -- clicking either will pop a modal prompting the user to enter an (optional) note as to the reason why. On save the orders "plan_length" will be adjusted and payments will be cancelled / rescehulded as needed. It should not show a remove payments if they've already made all the payments on the plan
> This would let refunds always count towards the number of remaining payments but if they decide a refund shouldn't count they can add an extra payment onto the plan
> 
> This is a lot more code than your solution but I think it'll be easier to explain what's going on than with a checkbox...
> 
> What do you think?

You're probably right on that checkbox.
Let's go with your idea, both will create problems, but yours is easier to explain, you're right.

thomasplevy:
Alright... I'm cleaining up my local and writing some tests but this is what I've arrived at:

https://user-images.githubusercontent.com/1290739/129620903-e8a1c391-208c-4302-bc2a-f01381ae9426.mp4

How's this look?

Hiccup I've run into: I set it so you cannot reduce it past 1 remaining payment... If you wish to *end* a payment plan it won't be possible with this (as it currently isn't possible anyway). The method for doing this exposed the `billing_length` property and adjusts that property according to the number of remaining payments. If we reduce the billing length to `0` that effectively turns the order into a recurring plan with no end date instead of "ending" the plan.

I'm working out an alternate method to end a plan and I haven't come up with a low-impact method of doing so as of yet...

thomasplevy:
@eri-trabiccolo this is ready for review

includes/functions/llms-functions-locale.php
```diff
@@ -99,6 +99,53 @@ function llms_get_states() {
...
+
+	/**
+	 * Filter the translated name for a given time period string.
+	 *
+	 * @since [version]
+	 *
+	 * @param string $period Translated period name.
+	 * @param int    $length Period length, used for pluralization.
+	 */
+	return apply_filters( 'llmst_ime_period_l10n', $period, $length );
```
eri-trabiccolo:
```suggestion
	return apply_filters( 'llms_time_period_l10n', $period, $length );
```

includes/admin/post-types/meta-boxes/class.llms.meta.box.order.details.php
```diff
@@ -110,6 +115,70 @@ public function save( $post_id ) {
...
+			return -1;
+		}
+
+		$payments  = llms_filter_input( INPUT_POST, '_llms_remaining_payments', FILTER_SANITIZE_NUMBER_INT );
+		$remaining = $order->get_remaining_payments();
+
+		// Nothing to save.
+		if ( null === $payments || absint( $payments ) === $remaining || absint( $payments ) < 1 ) {
+			return 0;
+		}
```
eri-trabiccolo:
```suggestion
		$payments  = (int) llms_filter_input( INPUT_POST, '_llms_remaining_payments', FILTER_SANITIZE_NUMBER_INT );
		$remaining = $order->get_remaining_payments();

		// Nothing to save.
		if ( $payments < 1 || $payments === $remaining ) {
			return 0;
		}
```

eri-trabiccolo (changes requested):
Looks good, but I need to think a little bit more about this...

includes/models/model.llms.order.php
```diff
@@ -238,13 +238,13 @@ protected function after_create() {
...
+	 * Calculate the date when a recurring order should stop being billed
 	 *
 	 * Applicable to orders created from plans with a set # of billing intervals.
 	 *
 	 * @since 3.10.0
 	 *
-	 * @return int
+	 * @return int Calculated billing end date as a Unix timestamp.
 	 */
 	private function calculate_billing_end_date() {
```
eri-trabiccolo: We don't use this anymore, and ti returns something unreliable, because the "start" is not correct.
RIght?
So we can remove this.

eri-trabiccolo (changes requested):
I think I'm good with this. <3

includes/models/model.llms.order.php
thomasplevy: @eri-trabiccolo yes you're right!

Issue #1463, opened 2020-12-15
Possible Fatal when deleting a product (course/membership) related to a (recurring) manual payment


eri-trabiccolo:
### Reproduction Steps
(a way to reproduce it)
+ as admin
+ create a course with an access plan with recurring payments
+ buy the access plan with manual payment gateway
+ go to the backend and delete (permanently) the course
+ edit the pending order and record a manual transaction so that the next recurring payment is scheduled
+ go to LifterLMS->Status->Scheduled Actions
+ run the pending action llms_charge_recurring_payment related to the just generated order

### Expected Behavior

+ no errors

### Actual Behavior

+ error generated/logged
+ you cannot see the list of the orders or any other post type, so basically you cannot even easily delete the order, at least not with that user (admin and buyer).

This is kinda an edge case but raises some questions:
1) what should we do with recurring payments related to a deleted product?
2) is it normal that the basic notification is instantiated in admin (edit.php)?
3) why do we "parse" merge codes that are not part of the content of the notification? The basic notification for the manual payment, by default, doesn't contain the merge code that produces that error, so I wouldn't expect it to be parsed at all.

### Error Messages / Logs

```
[15-Dec-2020 18:36:04 UTC] PHP Fatal error:  Uncaught Error: Call to a member function get_customer_name() on null in /shared/httpd/mywp/wordpress.git/wp-content/plugins/lifterlms/includes/notifications/views/class.llms.notification.view.manual.payment.due.php:193
Stack trace:
#0 /shared/httpd/mywp/wordpress.git/wp-content/plugins/lifterlms/includes/abstracts/llms.abstract.notification.view.php(584): LLMS_Notification_View_Manual_Payment_Due->set_merge_data('{{CUSTOMER_NAME...')
#1 /shared/httpd/mywp/wordpress.git/wp-content/plugins/lifterlms/includes/abstracts/llms.abstract.notification.view.php(627): LLMS_Abstract_Notification_View->get_merged_string('A payment is du...')
#2 /shared/httpd/mywp/wordpress.git/wp-content/plugins/lifterlms/includes/notifications/processors/class.llms.notification.processor.email.php(62): LLMS_Abstract_Notification_View->get_subject()
#3 /shared/httpd/mywp/wordpress.git/wp-content/plugins/lifterlms/includes/libraries/wp-background-processing/wp-background-process.php(270): LLMS_Notification_Processor_Email->task(38)
[...]

PR #1748, opened 2021-08-13
Notifications: Only populate effectively used merged data

### LINKED ISSUE #1465: Merge codes: only parse actually used merge codes.
Originally posted here:
https://github.com/gocodebox/lifterlms/issues/1463#issuecomment-746726678

@eri-trabiccolo said:
> why do we "parse" merge codes that are not part of the content of the notification? The basic notification for the manual payment, by default, doesn't contain the merge code that produces that error, so I wouldn't expect it to be parsed at all.
> It never occurred to me to reduce the list of potentially used merge codes down to a list of actually used merge codes.

@thomasplevy replied

> A find & replace is run on the content of the merge code, the find is populated with the list of possible merge codes and the replace array with all the merged data.
> 
> You're correct that we can optimize this to prevent the merged data that is never used from being populated and instead search for all the merge codes actually used in the notification content and then run the find an replace requiring us to only ever populate what is needed.

eri-trabiccolo:
## Description
Fixes #1465 

## How has this been tested?
Manually and new unit tests

## Types of changes
Improvement.

## Checklist:

### DIFF
--- a/.changelogs/issue_1465.yml
+++ b/.changelogs/issue_1465.yml
@@ -1,0 +1,5 @@
+significance: patch
+type: changed
+links:
+  - "#1465"
+entry: When building a notification content, only parse actually used merge codes.
--- a/.changelogs/issue_1465.yml
+++ b/.changelogs/issue_1465.yml
@@ -2,4 +2,4 @@ significance: patch
 type: changed
 links:
   - "#1465"
-entry: When building a notification content, only parse actually used merge codes.
+entry: When building notification content, only parse merge codes used in the notification.
--- a/includes/abstracts/llms.abstract.notification.view.php
+++ b/includes/abstracts/llms.abstract.notification.view.php
@@ -11,7 +11,7 @@
 defined( 'ABSPATH' ) || exit;
 
 /**
- * LLMS_Abstract_Notification_View class
+ * LLMS_Abstract_Notification_View class.
  *
  * @since 3.8.0
  * @since 3.30.3 Explicitly define undefined properties.
@@ -549,7 +549,7 @@ abstract class LLMS_Abstract_Notification_View extends LLMS_Abstract_Options_Dat
 	}
 
 	/**
-	 * Get available merge codes for the current notification
+	 * Get available merge codes for the current notification.
 	 *
 	 * @since 3.8.0
 	 * @since 3.11.0 Unknown.
@@ -558,12 +558,15 @@ abstract class LLMS_Abstract_Notification_View extends LLMS_Abstract_Options_Dat
 	 * @return array
 	 */
 	public function get_merge_codes() {
+
 		if ( ! isset( $this->merge_codes ) ) {
 			$codes = array_merge( $this->get_merge_code_defaults(), $this->set_merge_codes() );
 			asort( $codes );
 			$this->merge_codes = $codes;
 		}
+
 		return apply_filters( $this->get_filter( 'get_merge_codes' ), $this->merge_codes, $this );
+
 	}
 
 	/**
@@ -592,7 +595,7 @@ abstract class LLMS_Abstract_Notification_View extends LLMS_Abstract_Options_Dat
 	}
 
 	/**
-	 * Merge a string
+	 * Merge a string.
 	 *
 	 * @since 3.8.0
 	 * @since 3.37.19 Use `in_array` with strict comparison.
@@ -603,17 +606,17 @@ abstract class LLMS_Abstract_Notification_View extends LLMS_Abstract_Options_Dat
 	 */
 	private function get_merged_string( $string ) {
 
-		// only merge if there's codes in the string.
+		// Only merge if there's codes in the string.
 		if ( false !== strpos( $string, '{{' ) ) {
 
 			foreach ( $this->get_used_merge_codes( $string ) as $code ) {
 
-				// set defaults.
+				// Set defaults.
 				if ( in_array( $code, array_keys( $this->get_merge_code_defaults() ), true ) ) {
 
 					$func = 'set_merge_data_default';
 
-					// set customs with extended class func.
+					// Set customs with extended class func.
 				} else {
 
 					$func = 'set_merge_data';
@@ -639,14 +642,16 @@ abstract class LLMS_Abstract_Notification_View extends LLMS_Abstract_Options_Dat
 	 */
 	private function get_used_merge_codes( $string ) {
 
-		$merged_codes = array_keys( $this->get_merge_codes() );
-
-		return array_filter(
-			$merged_codes,
-			function ( $code ) use ( $string ) {
-				return false !== strpos( $string, $code );
-			}
+		return array_keys(
+			array_filter(
+				$this->get_merge_codes(),
+				function ( $code ) use ( $string ) {
+					return false !== strpos( $string, $code );
+				},
+				ARRAY_FILTER_USE_KEY
+			)
 		);
+
 	}
 
 	/**
--- a/includes/abstracts/llms.abstract.notification.view.php
+++ b/includes/abstracts/llms.abstract.notification.view.php
@@ -609,10 +609,12 @@ abstract class LLMS_Abstract_Notification_View extends LLMS_Abstract_Options_Dat
 		// Only merge if there are codes in the string.
 		if ( false !== strpos( $string, '{{' ) ) {
 
+			$merge_code_defaults = $this->get_merge_code_defaults();
+
 			foreach ( $this->get_used_merge_codes( $string ) as $code ) {
 
 				// Set defaults.
-				if ( in_array( $code, array_keys( $this->get_merge_code_defaults() ), true ) ) {
+				if ( array_key_exists( $code, $merge_code_defaults ) ) {
 
 					$func = 'set_merge_data_default';
 
--- a/includes/abstracts/llms.abstract.notification.view.php
+++ b/includes/abstracts/llms.abstract.notification.view.php
@@ -88,9 +88,9 @@ abstract class LLMS_Abstract_Notification_View extends LLMS_Abstract_Options_Dat
 	protected $user;
 
 	/**
-	 * Merge codes
+	 * Merge codes.
 	 *
-	 * @var $merge_codes
+	 * @var string[]
 	 */
 	protected $merge_codes;
 
--- a/includes/abstracts/llms.abstract.notification.view.php
+++ b/includes/abstracts/llms.abstract.notification.view.php
@@ -606,7 +606,7 @@ abstract class LLMS_Abstract_Notification_View extends LLMS_Abstract_Options_Dat
 	 */
 	private function get_merged_string( $string ) {
 
-		// Only merge if there's codes in the string.
+		// Only merge if there are codes in the string.
 		if ( false !== strpos( $string, '{{' ) ) {
 
 			foreach ( $this->get_used_merge_codes( $string ) as $code ) {
--- a/includes/abstracts/llms.abstract.notification.view.php
+++ b/includes/abstracts/llms.abstract.notification.view.php
@@ -88,6 +88,13 @@ abstract class LLMS_Abstract_Notification_View extends LLMS_Abstract_Options_Dat
 	protected $user;
 
 	/**
+	 * Merge codes
+	 *
+	 * @var $merge_codes
+	 */
+	protected $merge_codes;
+
+	/**
 	 * Replace merge codes with actual values
 	 *
 	 * @since 3.8.0
@@ -546,13 +553,17 @@ abstract class LLMS_Abstract_Notification_View extends LLMS_Abstract_Options_Dat
 	 *
 	 * @since 3.8.0
 	 * @since 3.11.0 Unknown.
+	 * @since [version] Cache merge codes.
 	 *
 	 * @return array
 	 */
 	public function get_merge_codes() {
-		$codes = array_merge( $this->get_merge_code_defaults(), $this->set_merge_codes() );
-		asort( $codes );
-		return apply_filters( $this->get_filter( 'get_merge_codes' ), $codes, $this );
+		if ( ! isset( $this->merge_codes ) ) {
+			$codes = array_merge( $this->get_merge_code_defaults(), $this->set_merge_codes() );
+			asort( $codes );
+			$this->merge_codes = $codes;
+		}
+		return apply_filters( $this->get_filter( 'get_merge_codes' ), $this->merge_codes, $this );
 	}
 
 	/**
@@ -595,7 +606,7 @@ abstract class LLMS_Abstract_Notification_View extends LLMS_Abstract_Options_Dat
 		// only merge if there's codes in the string.
 		if ( false !== strpos( $string, '{{' ) ) {
 
-			foreach ( array_keys( $this->get_merge_codes() ) as $code ) {
+			foreach ( $this->get_used_merge_codes( $string ) as $code ) {
 
 				// set defaults.
 				if ( in_array( $code, array_keys( $this->get_merge_code_defaults() ), true ) ) {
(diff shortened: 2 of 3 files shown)


eri-trabiccolo:
I was working on #1463 but this looked a low hanging fruit so I'm pushing this incomplete PR before I forget about it.
Anything against this approach?
It looks so simple that I'm afraid I'm missing something obvious.

thomasplevy:
If it works, it works <3

includes/abstracts/llms.abstract.notification.view.php
```diff
@@ -608,7 +609,13 @@ private function get_merged_string( $string ) {
 
 				}
 
-				$string = str_replace( $code, $this->$func( $code ), $string );
+				$string = preg_replace_callback(
+					'#' . preg_quote( $code, '#' ) . '#',
```
pondermatic: Why use '#' delimiters instead of the traditional '/'?

pondermatic:
I'd like to see a unit test to see how much of a difference this optimization makes.

includes/abstracts/llms.abstract.notification.view.php
eri-trabiccolo: Actually for some reason I always use `#`.
But you made me think that in this case, since `#` are escaped by default, I don't need to pass the delimiter param, I guess...

eri-trabiccolo:
> I'd like to see a unit test to see how much of a difference this optimization makes.

Do you mean real benchmarks?
The reason for this PR is just to execute `$func` only on found "codes" instead of for each possible `$code`. The downside for this is that the callback will be executed for each occurrence of the same `$code`, which is unusual but not impossible. Did you mean this?

pondermatic:
I was concerned about the performance of running a regular expression for each code. It's probably faster than calling `set_merge_data()` for each code though.

eri-trabiccolo:
> I was concerned about the performance of running a regular expression for each code. It's probably faster than calling `set_merge_data()` for each code though.

I think you might be right, but it depends on several factors. I agree though that generally using regexp might be overkilling, especially for what we want to do here. Maybe a double for (the first to search for actually present `$code`s in the body and the second over them to expand them) should be faster. I'll go with that. Thanks.

eri-trabiccolo:
@pondermatic 
what do you think of the "new implementation"? Aside for missing unit tests (that still won't be benchmarks :)).

pondermatic (dismissed):
I like this technique of "caching" the merge codes.

thomasplevy (approved):
Tentative approval but I haven't physically tested. It looks good.

I need to clean up my local and I don't want to stash everything to checkout this PR right this moment.

@pondermatic if you have some time to review and also run some physical tests I'd appreciate it.

.changelogs/issue_1465.yml
```diff
@@ -0,0 +1,5 @@
+significance: patch
+type: changed
+links:
+  - "#1465"
+entry: When building a notification content, only parse actually used merge codes.
```
thomasplevy:
```suggestion
entry: When building notification content, only parse merge codes used in the notification.
```

includes/abstracts/llms.abstract.notification.view.php
```diff
@@ -87,6 +87,13 @@ abstract class LLMS_Abstract_Notification_View extends LLMS_Abstract_Options_Dat
 	 */
 	protected $user;
 
+	/**
+	 * Merge codes
+	 *
+	 * @var $merge_codes
+	 */
+	protected $merge_codes;
```
pondermatic:
```suggestion
	/**
	 * Merge codes.
	 *
	 * @var string[]
	 */
	protected $merge_codes = array();
```

includes/abstracts/llms.abstract.notification.view.php
```diff
@@ -542,17 +549,24 @@ public function get_icon_src() {
...
+	 * @since [version] Cache merge codes.
 	 *
 	 * @return array
 	 */
 	public function get_merge_codes() {
-		$codes = array_merge( $this->get_merge_code_defaults(), $this->set_merge_codes() );
-		asort( $codes );
-		return apply_filters( $this->get_filter( 'get_merge_codes' ), $codes, $this );
+
+		if ( ! isset( $this->merge_codes ) ) {
```
pondermatic, line 562:
```suggestion
		if ( empty( $this->merge_codes ) ) {
```
This change will allow extended classes to empty the cache without them having to call `get_merge_codes()` to refill it.

includes/abstracts/llms.abstract.notification.view.php
```diff
@@ -581,27 +595,28 @@ protected function get_merge_code_defaults() {
...
 	 * @since 3.37.19 Use `in_array` with strict comparison.
+	 * @since [version] Only populate effectively used merged data.
 	 *
 	 * @param string $string An unmerged string.
 	 * @return string
 	 */
 	private function get_merged_string( $string ) {
 
-		// only merge if there's codes in the string.
+		// Only merge if there's codes in the string.
```
pondermatic:
```suggestion
		// Only merge if there are codes in the string.
```

includes/abstracts/llms.abstract.notification.view.php
```diff
@@ -617,6 +632,28 @@ private function get_merged_string( $string ) {
...
+	 * @param string $string Text string whereto look for merge codes.
+	 * @return array Returns a list of merge codes actually used in the passed string.
+	 */
+	private function get_used_merge_codes( $string ) {
+
+		return array_keys(
+			array_filter(
+				$this->get_merge_codes(),
+				function ( $code ) use ( $string ) {
+					return false !== strpos( $string, $code );
```
pondermatic, line 651: Are merge codes case sensitive? If not, change `strpos()` to `stripos()`.

tests/phpunit/unit-tests/abstracts/class-llms-test-abstract-notification-view.php
```diff
@@ -0,0 +1,195 @@
+<?php
+/**
+ * Tests for the LLMS_Abstract_Notification_View
```
pondermatic:
```suggestion
 * Tests for the LLMS_Abstract_Notification_View.
```

tests/phpunit/unit-tests/abstracts/class-llms-test-abstract-notification-view.php
```diff
@@ -0,0 +1,195 @@
...
+ *
+ * @group abstracts
+ * @group notifications
+ *
+ * @since [version]
+ */
+class LLMS_Test_Abstract_Notification_View extends LLMS_UnitTestCase {
+
+	/**
+	 * Setup the test case
```
pondermatic:
```suggestion
	 * Setup the test case.
```

tests/phpunit/unit-tests/abstracts/class-llms-test-abstract-notification-view.php
```diff
@@ -0,0 +1,195 @@
...
+	 *
+	 * @return void
+	 */
+	public function set_up() {
+
+		parent::set_up();
+
+		global $wpdb;
+
+		$this->main   = $this->get_stub(
```
pondermatic:
```suggestion
		$this->main = $this->get_stub(
```

includes/abstracts/llms.abstract.notification.view.php
```diff
@@ -581,27 +595,28 @@ protected function get_merge_code_defaults() {
...
-		// only merge if there's codes in the string.
+		// Only merge if there's codes in the string.
 		if ( false !== strpos( $string, '{{' ) ) {
 
-			foreach ( array_keys( $this->get_merge_codes() ) as $code ) {
+			foreach ( $this->get_used_merge_codes( $string ) as $code ) {
 
-				// set defaults.
+				// Set defaults.
 				if ( in_array( $code, array_keys( $this->get_merge_code_defaults() ), true ) ) {
```
pondermatic: I know that this line was written prior to this PR, but while we are improving performance, `array_keys( $this->get_merge_code_defaults() )` should be called once before the `foreach` loop and assigned to a variable, instead of being called for each used merge code.

includes/abstracts/llms.abstract.notification.view.php
```diff
@@ -617,6 +632,28 @@ private function get_merged_string( $string ) {
...
+	 * Retrieve merge codes used in a given string.
+	 *
+	 * @since [version]
+	 *
+	 * @param string $string Text string whereto look for merge codes.
+	 * @return array Returns a list of merge codes actually used in the passed string.
+	 */
+	private function get_used_merge_codes( $string ) {
+
+		return array_keys(
```
pondermatic, line 647: I'll probably never need to debug this method, but in the future, please keep in mind that it is difficult to step debug complex functions when they are not assigned to a variable. I can't see the result of this series of function calls because it is immediately returned. Plus, this method uses the return value directly in a `foreach` loop, so I can't see it there either.

pondermatic (changes requested):

includes/abstracts/llms.abstract.notification.view.php
eri-trabiccolo: I don't want this, I want the `$this->merge_codes` to be even empty, if it's empty, but set, the code in this if block must not be executed every time...

includes/abstracts/llms.abstract.notification.view.php
eri-trabiccolo: I'm with you, and I wanted to do that as well, BUT we always treated merge codes as case sensitive.
See the use of str_replace and not str_ireplace.

includes/abstracts/llms.abstract.notification.view.php
eri-trabiccolo: You're right

includes/abstracts/llms.abstract.notification.view.php
thomasplevy: for the purposes of not making things stupid confusing they are case sensitive and that's okay.

includes/abstracts/llms.abstract.notification.view.php
eri-trabiccolo: I see, I will try to make stuff more readable in the future but I really think this one is not complex to read.

pondermatic: It's difficult to use Xdebug to step through live running code and see the contents of variables. The readability is OK.

eri-trabiccolo: Oh I see.
Is it better to declare several _useless_ variables for "maybe troubleshooting" purpose only?
I do understand your concern, but as long as something is _readable_ (which I understand is personal) we can always unpack when needed.

The following code would be ok for Xdebug?
```php
	private function get_used_merge_codes( $string ) {
		
		$merge_codes      = $this->get_merge_codes();
		$used_merge_codes = array_filter(
			$merge_codes,
			function ( $code ) use ( $string ) {
				return false !== strpos( $string, $code );
			},
			ARRAY_FILTER_USE_KEY
		);
		$used_merge_codes = array_keys( $used_merge_codes );

		return $used_merge_codes;

	}
```

includes/abstracts/llms.abstract.notification.view.php
eri-trabiccolo: solved in https://github.com/gocodebox/lifterlms/pull/1748/commits/8054c22f721d883455affc857636ede1ea67892d

includes/abstracts/llms.abstract.notification.view.php
pondermatic: Yes, that code would be easy to step through and see the data as it flows through functions.

Assigning the result of a function to a variable before returning the value is most useful when I'm not sure what the result will be, such as when `apply_filters()` is used. Sometimes it's faster for me to step debug through the code than to unpack multiple function calls in my head.

pondermatic (approved):

includes/abstracts/llms.abstract.notification.view.php
eri-trabiccolo: Do you think using variables when not needed would have any "micro" performance impact though?

pondermatic: Yes, there is a performance impact. From what I know about the internal Zend engine, it's probably **very** small. When a function returns data, the data is already taking up memory. To assign it to a variable, Zend allocates a small block of memory for the name, type, and stores the pointer to the memory that holds the data.

I woke up too dang early this morning, so I think just for fun I'll benchmark it.

pondermatic: @eri-trabiccolo, my benchmark results were surprising.

I wrote two functions that both call `str_repeat()`. One returns it directly, the other assigns it to a variable and then returns the variable. Each function is called 10,000 times. The number of times that the string should be repeated increments with the number of times the function is called.

**Sometimes function 1 is faster than function 2. The other times it is reversed.** The difference can be as great as 9 milliseconds.

It seems that there is no performance impact of assigning a function's result to a variable and then returning the variable.

```php
function return_function_result( $times ) {
	return str_repeat( 'A', $times );
}
function return_function_variable( $times ) {
	$result = str_repeat( 'A', $times );
	return $result;
}

$loopLimit = 10000;

$start = microtime(true);
for ( $i = 0; $i < $loopLimit; $i++ ) {
	$string = return_function_result( $i );
}
$stop = microtime(true);
$return_function_result_duration = $stop - $start;
echo "<pre>return_function_result()   = " . sprintf('%.2f', $return_function_result_duration * 1000) . "\n";

$start = microtime(true);
for ( $i = 0; $i < $loopLimit; $i++ ) {
	$string = return_function_variable( $i );
}
$stop = microtime(true);
$return_function_variable_duration = $stop - $start;
echo "return_function_variable() = " . sprintf('%.2f', $return_function_variable_duration * 1000) . "\n";

$difference = $return_function_variable_duration - $return_function_result_duration;
echo "                DIFFERENCE = " . sprintf('%.2f', $difference * 1000 ) . "</pre>";
```

eri-trabiccolo: I see,
well I think in this case performance should be referred to memory usage maybe rather than execution speed, and in a little bit more complex functions, where there are several declared variables, I mean something more similar to the method I wrote and the method Xdebug would like the best. :)

pondermatic: The memory used by a variable is basically the length of the variable's string name plus a pointer to the  [zval structure](https://www.phpinternalsbook.com/php7/zvals/basic_structure.html#the-zval-struct). The size of the zval is the same whether assigned to a variable or used/returned immediately.

Unless our methods are assigning a large number of variables, or the variable names are really long, the memory used and the execution time to allocate the memory are insignificant. When the method returns, the memory used by variables is reclaimed by PHP's garbage collector.

For a deeper dive, see https://github.com/php/php-src/blob/master/main/php_variables.c.

eri-trabiccolo: Thanks.
> When the method returns, the memory used by variables is reclaimed by PHP's garbage collector.

Which means that the memory used by variables is retained at least until the method returns.
Would be interesting to know what happens instead to the memory used by the zval structure of a function used immediately, how long it lasts I mean.

pondermatic: When returning from a function or method, the reference counter of zvals in that scope is decremented if they aren't used in the return. When a zval's reference counter reaches 0, the garage collector is allowed to reclaim the memory.

eri-trabiccolo: So we're saying that:
```php
		return array_keys(
			array_filter(
				$this->get_merge_codes(),
				function ( $code ) use ( $string ) {
					return false !== strpos( $string, $code );
				},
				ARRAY_FILTER_USE_KEY
			)
		);
```
is better than:
```php
		$merge_codes      = $this->get_merge_codes();
		$used_merge_codes = array_filter(
			$merge_codes,
			function ( $code ) use ( $string ) {
				return false !== strpos( $string, $code );
			},
			ARRAY_FILTER_USE_KEY
		);
		$used_merge_codes = array_keys( $used_merge_codes );

		return $used_merge_codes;
```

in that _very_ respect?
Am I understanding it correctly?

pondermatic: Yes. Assigning the results of function calls to locally scoped variables allows the value to be examined during [step debugging](https://xdebug.org/docs/step_debug). The memory used is insignificant and is reclaimed if needed.

Using function calls in arguments to other function calls is hard to debug. It can also be difficult to read if there are many nested function calls.

PR #1751, opened 2021-08-16
Don't output "You" for student name merge code in notifications

### LINKED ISSUE #1557: {{student_name}} merge code for achievements says "you" instead of student name when printed
### Reproduction Steps
This may not necessarily be a bug so much as a point of confusion.

+ Trigger any of the basic notifications that have the {{student_name}} merge code
+ go into the dashboard and look at the notification

image

### Expected Behavior

+ It should read student name but instead says you:
image

### Actual Behavior

+ This makes sense with our default layout! However, when users customize this there is an expectation that student_name should be the student's name. So it makes less sense when they say Congratulations student_name, you successfully enrolled in x. It reads on the front end as Congratulations You, you successfully enrolled in x.

### Error Messages / Logs

```

```

<details>

```


```

</details>

This issue has be recreated:
+ [X] Locally
+ [ ] On a staging site
+ [X] On a production website
+ [X] With only LifterLMS and a default theme

+ Browser name and version

thomasplevy:
## Description

Fixes #1557

## How has this been tested?

+ Manually

## Types of changes

+ Bugfix

## Checklist:

### DIFF
--- a/includes/notifications/views/class.llms.notification.view.achievement.earned.php
+++ b/includes/notifications/views/class.llms.notification.view.achievement.earned.php
@@ -14,8 +14,6 @@ defined( 'ABSPATH' ) || exit;
  * Notification View: Achievement Earned
  *
  * @since 3.8.0
- * @since 3.17.6 Unknown.
- * @since 3.30.3 Fixed spelling errors.
  */
 class LLMS_Notification_View_Achievement_Earned extends LLMS_Abstract_Notification_View {
 
--- a/includes/notifications/views/class.llms.notification.view.achievement.earned.php
+++ b/includes/notifications/views/class.llms.notification.view.achievement.earned.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Notifications/Views/Classes
  *
  * @since 3.8.0
- * @version 4.7.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -105,7 +105,8 @@ class LLMS_Notification_View_Achievement_Earned extends LLMS_Abstract_Notificati
 	 *
 	 * @since 3.8.0
 	 * @since 3.8.2 Unknown.
-	 * @since 4.7.0 Use `achievement_title` in favor of `title` for the {{ACHIEVEMENT_TITLE}} merge code
+	 * @since 4.7.0 Use `achievement_title` in favor of `title` for the {{ACHIEVEMENT_TITLE}} merge code.
+	 * @since [version] Remove output of "you" when displaying notification to the receiving student.
 	 *
 	 * @param string $code The merge code to get merged data for.
 	 * @return string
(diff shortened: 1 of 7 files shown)


eri-trabiccolo (dismissed):

pondermatic (dismissed):

thomasplevy:
@gocodebox/engineering 

Alright... so on closer inspection of this and reviewing the failing e2e tests (they're failing because the merge code content changed) I'm still feeling very frustrated with this issue...

The notification after this change *does not make a bit of sense* and feels like we're introducing a bug:

https://app.travis-ci.com/github/gocodebox/lifterlms/jobs/532079848#L449-L450

We get reports of the current behavior as a bug infrequently but I haven't wanted to change it because I think that *changing it* introduces a new bug...

So... if we're to merge this in and accept this as the new behavior would you agree that we should:

1. Update the default content of notifications that use the merge code to replace "{{STUDENT_NAME}}" with "you" (so that the defaults notifications make sense)
2. Migrate existing stored notifications -- however this is basically impossible to do because we cannot guarantee that users are expecting "You" or not and we also have to consider that many sites may have translated this so we'd need to replace the merge code with a translation and depending on the language it might not be so simple to do that...

After thinking about this I'm feeling that we should, possibly, *not* make this change and leave it as is and everyone involved (the support team) needs to be aware that *this is the way it works and we're not going to change it. 

I still hold that there is absolutely no reason to tell the student their own name in a notification displayed to themselves... again, look at any notifications you receive on your phone, on twitter, facebook, whatever... IT doesn't say "Thomas Levy received a new message" it says "*You* received a new message", "So and so liked *your* post" etc...

pondermatic:
This may not be a good idea, but I'll throw it out there. How about we create a new tag, like {{STUDENT_NAME_OR_SELF}}? It would work like the current {{STUDENT_NAME}} does, while {{STUDENT_NAME}} would be changed to always be the student's name.

eri-trabiccolo:
@thomasplevy, @pondermatic 
I don't think we have a simple solution for this.

I understand that now with the default message this change introduces a problem, but on the other hand without this change, the notifications messages, which use this `{{STUDENT_NAME}}`, are basically forced to follow the structure of the default message.
Also consider that while the default message makes sense in english for all the recipients, it doesn' t  make too much sense in Italian. E.g. at the moment:
`Congratulations! {{STUDENT_NAME}} enrolled in {{TITLE}}`
In english:
the student receives: `Congratulations! You enrolled in LifterLMS Quickstart Course`
the author receives: `Congratulations! Jamie Doe enrolled in LifterLMS Quickstart Course`
which makes perfectly sense.
Although in Italian it'll be, something like:
1)
student: `Complimenti! Tu è stato iscritto a  LifterLMS Quickstart Course`
author: `Complimenti! Jamie Doe è stato iscritto a LifterLMS Quickstart Course`
or
2)
student: `Complimenti! Tu sei stato iscritto a  LifterLMS Quickstart Course`
author: `Complimenti! Jamie Doe sei stato iscritto a LifterLMS Quickstart Course`

Non of them are good for both student and author pairs. 1) is good for the author but not for the student, 2) is good for the student but not for the author. So one, depending on the locale, must find some impersonal way to phrase that makes sense in  their locale, and it might even be impossible.

The real solution here would be to have two different messages, one for the "trigger" (the one with 'you') and another for the others. I don't see any other solution.

PR #1756, opened 2021-08-19
Issue 1746 traits

pondermatic:
## Description
1. Added 3 traits.
    1. `LLMS_Trait_Audio_Video_Embed`
    2. `LLMS_Trait_Sales_Page`
    3. `LLMS_Trait_Singleton`.
1. Updated `LLMS_Loader->autoload()` to handle traits.
1. Moved the loading of the `LLMS_Loader` class from the `LifterLMS` class to the main `lifterlms.php` file.
1. Added unit tests for the new traits.
1. Added traits to the coding standards document.

Fixes #1746

## How has this been tested?
* Used Xdebug to step through the code.
* Ran the full suite of unit tests.
* Monitored the PHP error log.

## Types of changes
What types of changes does your code introduce?  Structural.

## Checklist:
- [x] My code has been tested.
- [x] My code passes all existing automated tests. 
- [x] My code follows the LifterLMS Coding & Documentation Standards.

PR #1758, opened 2021-08-23
Implement traits

### LINKED ISSUE #1746: Implement usage of traits
LifterLMS is long overdue to use PHP traits. I often wish to add a trait for something but don't want to include the required "prework" in the related bugfix / feature PR.

+ Trait classes should follow the naming convention `LLMS_Trait_{$trait}`, eg: `LLMS_Trait_Singleton`.
+ Trait files should follow the naming convention `llms-trait-{$trait}.php`: eg: `llms-trait-singleton.php`
+ Trait classes should be stored in their own directory at `includes/traits`
+ Traits should be autoloaded via the `LLMS_Loader::autoload()` method

An initial singleton trait can be included and singleton classes in the core can remove their singleton methods and properties and be replaced with the new trait.

+ A singleton trait exists in the lifterlms-rest library: https://github.com/gocodebox/lifterlms-rest/blob/trunk/includes/traits/class-llms-rest-trait-singleton.php. I believe this can be lifted into the core (and ultimately replaced in the rest lib as the core's singleton could then be used).

+ the sales page redirect methods "shared" (duplicated) across courses and memberships (see #1743) can be converted into a "sales page" type trait containing these methods that can then be mixed into the course and membership models.

+ Audio and video embed code is similarly copied between courses and memberships

This would be a good start. There's lots of other potential areas where we could replace duplicated code with traits too but these jump out at me immediately.

+ Existing tests should continue to pass if traits are implemented properly.
+ When creating new traits I feel the best practice would be to instantiate a mock (anonymous?) class that uses the trait and write tests against the mock. However, for the purposes of ensuring the traits "integrate" properly having redundant tests for some traits *might* make sense. The sales page / redirect trait might benefit from having course/membership model tests AND having tests against an anonymous class that uses the trait out of the "intended" context

thomasplevy:
+ Fixes #1746
+ Update to #1756 

Includes latest changes to `lifterlms/lifterlms-tests`


thomasplevy:
@eri-trabiccolo this is a big change, I'm signing off on it already (from the previous PR @pondermatic opened) but I'd like to have your thoughts on this too.

Thanks

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -44,6 +44,7 @@
  * @since 3.34.0 Add `set_bulk()` method that will allow to update an object at once given an array of properties.
  * @since 3.34.0 Refresh the whole $post property with the just updated instance of WP_Post after updating it.
  * @since 3.36.1 In `set_bulk()` method, use WP_Error::$errors in place of WP_Error::has_errors() to support WordPress version prior to 5.1.
+ * @since [version] Move `get_embed()` to `LLMS_Trait_Sales_Page`.
```
```suggestion
 * @since [version] Move `get_embed()` to `LLMS_Trait_Audio_Video_Embed`.
```

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -812,51 +813,6 @@ protected function get_creation_args( $args = null ) {
...
-	 *
-	 * @since 3.17.0
-	 * @since 3.17.5 Unknown.
-	 *
-	 * @param string $type Optional. Embed type [video|audio]. Default is 'video'.
-	 * @param string $prop Optional. Postmeta property name. Default is empty string.
-	 *                     If not supplied it will default to {$type}_embed.
-	 * @return string
-	 */
-	protected function get_embed( $type = 'video', $prop = '' ) {
```
eri-trabiccolo: You might have already discussed this... but I don't think we can just remove this method here.
This is a protected method that a 3rd party extending class might rely on.
So we have to keep this and deprecate it?
Not a big of a deal since trait's method has precedence on the inherited one.
What do you think?

includes/models/model.llms.course.php
```diff
@@ -17,8 +17,9 @@
  * @since 3.30.3 Explicitly define class properties.
  * @since 4.0.0 Remove previously deprecated class methods.
  * @since 5.2.1 Check for an empty sales page URL or ID.
+ * @since [version] Move audio and video embed properties and methods to `LLMS_Trait_Audio_Video_Embed`.
+ *              Move sales page properties and methods to `LLMS_Trait_Sales_Page`.
  *
- * @property string $audio_embed                URL to an oEmbed enable audio URL.
```
eri-trabiccolo, line 21: Are properties defined in the used traits automatically "scanned" and added to the documenation?
I mean, will they appear here? https://developer.lifterlms.com/reference/classes/llms_course/#properties
Same thing with the methods?
If not, I was wondering if we could keep here, at least, the documentation of the properties we moved in traits, adding a note that they're part of this and this other trait.
I found this doc pretty useful when I had to create the rest-api related resources.
What do you think?

includes/traits/llms-trait-sales-page.php
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * @property int    $sales_page_content_page_id WP Post ID of the WP page to redirect to when $sales_page_content_type is 'page'.
+ * @property string $sales_page_content_type    Sales page behavior [none,content,page,url].
+ * @property string $sales_page_content_url     Redirect URL for a sales page, when $sales_page_content_type is 'url'.
+ */
+trait LLMS_Trait_Sales_Page {
+	/**
+	 * @inheritdoc
+	 */
+	abstract protected function add_properties( $props = array() );
```
eri-trabiccolo, line 29: Mmm, I see we want to require that exhibiting classes implement `add_properties` and `get` methods because we use these methods in the trait but then these two abstracts should be added to the `LLMS_Trait_Audio_Video_Embed` as well (same reason).
This is formally correct, but at the same time we don't do any requirement on the properties we access directly, like `$this->model_post_type`. Should we implement something of the sort as well?
Or we just get rid of these two abstract methods here and rely on the fact that exhibiting classes which do not implement these two methods will just throw fatals on execution? :D

eri-trabiccolo (changes requested):
Some considerations added. A couple of changes requested.
Overall looks good to me.

includes/abstracts/abstract.llms.post.model.php
pondermatic: A 3rd party extending class can still call `get_embed()` even if it's in a trait instead of in this class. To the extending class, there is no difference. It can even override the method if it needs to.

includes/models/model.llms.course.php
pondermatic: Great question! I'll test it in the doc parser.

includes/traits/llms-trait-sales-page.php
pondermatic: Thank you for catching that `LLMS_Trait_Audio_Video_Embed` needs to have abstract `add_properties()` and `get()` methods. Having an [abstract trait method](https://www.php.net/manual/en/language.oop5.traits.php#language.oop5.traits.abstract) imposes a requirement on the exhibiting class to implement the method, else a fatal error is triggered. It's also good documentation.

Unfortunately, properties can not be declared as abstract and can not be [redefined](https://www.php.net/manual/en/language.oop5.traits.php#language.oop5.traits.properties) without triggering a fatal error. I did not find a simple way to make PHP enforce the requirement. Options:
1. Use `property_exists()` and then `trigger_error()` if `$model_post_type` does not exist.
2. Add notes in the DocBlock and hope for the best.

@eri-trabiccolo and @thomasplevy, what are your thoughts on properties that the trait depends on?

includes/abstracts/abstract.llms.post.model.php
eri-trabiccolo: @pondermatic 
I don't think we understood each other on this. :D
We're removing the method `get_embed()` from the `LLMS_Post_Model` abstract class, moving it to the trait `LLMS_Trait_Audio_Video_Embed`, but `LLMS_Post_Model` doesn't use the trait `LLMS_Trait_Audio_Video_Embed`.
So if a 3rd party is for any reason extending `LLMS_Post_Model` and referring to `$this->get_embed();` what happens?

includes/traits/llms-trait-sales-page.php
eri-trabiccolo: >  Having an abstract trait method imposes a requirement on the exhibiting class to implement the method, else a fatal error is triggered.


Yeah sure, otherwise a fatal error will be triggered as well when trying to execute a non existing method. Right?
I mean maybe you're right and it's better to have that fatal triggering asap and not depending on whether that method will be ever executed?!
I see it's better for documentation in any case.
 

> Unfortunately, properties can not be declared as abstract and can not be redefined without triggering a fatal error. I did not find a simple way to make PHP enforce the requirement. Options:
> 
> Use property_exists() and then trigger_error() if $model_post_type does not exist.
> Add notes in the DocBlock and hope for the best.


Not sure, also because thinking more about it, in that case we don't only need that `$model_post_type` exists, but that it's also defined (`isset`) and actually even not empty.
Let's go with DocBlock only. IMHO.

includes/models/model.llms.course.php
pondermatic: Unfortunately, when a trait defines a "magic" property using the [`@property`](https://docs.phpdoc.org/3.0/guide/references/phpdoc/tags/property.html) DocBlock tag, the property is not included in the generated documentation. We'll need to add the tags to classes that use the traits.

includes/abstracts/abstract.llms.post.model.php
pondermatic: Oh! Now I see. That's a very important point!

I think I need to move `get_embed()` back to the `LLMS_Post_Model` class. `LLMS_Trait_Audio_Video_Embed` will just define `get_audio()` and `get_video()`.

includes/traits/llms-trait-sales-page.php
thomasplevy: Docblock only... if a dev uses the trait in a class they're going to see PHP's native error if they're trying to use properties that don't exist.

pondermatic (dismissed):

includes/abstracts/abstract.llms.post.model.php
eri-trabiccolo: @pondermatic this is still not resolved, right?

eri-trabiccolo (changes requested):

pondermatic:
I forgot to push the change regarding `get_embed()`.

includes/abstracts/abstract.llms.post.model.php
pondermatic: I fixed it but forgot to push the commit.

eri-trabiccolo (approved):

PR #1760, opened 2021-08-26
BUild a tool to help identify orders potentially affected by payment plan changes in 5.3.0

thomasplevy:
## Description

Building on #1744

Adds a (conditionally loaded) tool to identify orders potentially affected by the changes to how an limited-billing recurring order is ended

Prior to #1744 these orders ended by a calculated date, now they end based on the number of successful (and refunded) payments.

This tool generates a CSV of affected active and on-hold orders with a billing length of at least one which meet one of the following criteria:

1. The order contains at least one refund
2. The order is marked as "ended" and has fewer successful transactions than the expected number (the billing length)

This tool does not take any actual action on the orders, it just generates a report that an admin should review and do what they like with. The main concern would be orders that ended incorrectly and need additional billing to be added. Admins can visit the order screen and update the remaining payments via the new interface should they disagree with the new rules.

This PR additionally introduces a pluggable wrapper for `exit()` that has been redefined in `lifterlms/lifterlms-tests` so that we can easily test methods that `exit()`.

It also updates other admin tool testing classes to consolidate redundant tests (mainly testing methods like `get_text()`...

And I moved all the pluggable wrappers from `includes/llms.functions.core.php` to their own function file.

## TODO

- [x] A piece of documentation needs to be written about this and then the tool description's link should be updated once the link exists. 

## How has this been tested?

+ Manually
+ New tests

## Types of changes
+ New feature

## Checklist:

### DIFF
--- a/includes/admin/tools/class-llms-admin-tool-limited-billing-order-locator.php
+++ b/includes/admin/tools/class-llms-admin-tool-limited-billing-order-locator.php
@@ -1,0 +1,280 @@
+<?php
+/**
+ * LLMS_Admin_Tool_Limited_Billing_Order_Locator class
+ *
+ * @package LifterLMS/Admin/Tools/Classes
+ *
+ * @since [version]
+ * @version [version]
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Admin tool which generates a report of limited billing orders affected by order end changes
+ *
+ * @since [version]
+ * 
+ * @link https://github.com/gocodebox/lifterlms/pull/1744
+ */
+class LLMS_Admin_Tool_Limited_Billing_Order_Locator extends LLMS_Abstract_Admin_Tool {
+
+	/**
+	 * Tool ID.
+	 *
+	 * @var string
+	 */
+	protected $id = 'limited-billing-order-locator';
+
+	/**
+	 * Query the database for orders that may be affected by the change
+	 *
+	 * @since [version]
+	 *
+	 * @return array[] Returns an array of arrays where each array represents a line in the generated CSV file.
+	 */
+	protected function generate_csv() {
+
+		$csv = array();
+
+		$orders = new WP_Query( array(
+			'post_type'      => 'llms_order',
+			'post_status'    => array( 'llms-active', 'llms-on-hold' ),
+			'posts_per_page' => -1,
+			'meta_query'     => array(
+				array(
+					'key'     => '_llms_billing_length',
+					'value'   => 1,
+					'compare' => '>=',
+				),
+			),
+		) );
+
+		foreach ( $orders->posts as $order ) {
+
+			$order = llms_get_post( $order );
+			if ( ! $order || ! is_a( $order, 'LLMS_Order' ) ) {
+				continue;
+			}
+
+			$csv[] = $this->get_order_csv( $order );
+
+		}
+
+		return array_filter( $csv );
+
+	}
+
+	/**
+	 * Create a csv "file" via output buffering and return it as a string
+	 *
+	 * @since [version]
+	 *
+	 * @return string
+	 */
+	protected function get_csv_file() {
+
+		$csv = $this->get_csv();
+
+		// Add header row.
+		array_unshift( $csv, array(
+			'Order ID',
+			'Expected Payments',
+			'Total Payments',
+			'Successful Payments',
+			'Refunded Payments',
+			'Edit Link',
+		) );
+
+		// Create the CSV file.
+		ob_start();
+		$fh = fopen( 'php://output', 'w' );
+		foreach ( $csv as $line ) {
+			fputcsv( $fh, $line );
+		}
+		fclose( $fh );
+
+		return ob_get_clean();
+
+	}
+
+	/**
+	 * Retrieve a description of the tool
+	 *
+	 * This is displayed on the right side of the tool's list before the button.
+	 *
+	 * @since [version]
+	 *
+	 * @return string
+	 */
+	protected function get_description() {
+
+		$count = count( $this->get_csv() );
+
+		$desc  = sprintf(
+			// Translators: %1$s = opening anchor link to documentation; %2$s = closing anchor link.
+			__( 'The method used to determine when a limited-billing recurring order has completed its payment plan changed during version 5.3.0. This tool provides a report of orders which may been affected by this change. %1$sRead more%2$s about this change.', 'lifterlms' ),
+			'<a href="https://lifterlms.com/docs/###" target="_blank">',
+			'</a>'
+		);
+		$desc .= ' ';
+		// Translators: %d = the number of pending batches.
+		$desc .= sprintf(
+			_n(
+				'There is %d order that should be reviewed.',
+				'There are %d orders that should be reviewed.',
+				$count,
+				'lifterlms'
+			),
+			$count
+		);
+
+		return $desc;
+
+	}
+
+	/**
+	 * Retrieve the tool's label
+	 *
+	 * The label is the tool's title. It's displayed in the left column on the tool's list.
+	 *
+	 * @since [version]
+	 *
+	 * @return string
+	 */
+	protected function get_label() {
+		return __( 'Limited Billing Orders', 'lifterlms' );
+	}
+
+	/**
+	 * Retrieves an array representing a line in generated CSV for the given order
+	 *
+	 * An order is considered to be affected by the change if either of the following conditions are true:
+	 *
+	 *   + Any number of refunds exist for the order. Since we are now counting refunds towards the billing limit
+	 *     an active order with any number of refunds should be reviewed by the admin.
+	 *
+	 *   + The plan is marked as "ended" and the total number of successful payments is not equal to the billing length.
+	 *     In this scenario admins will likely want to add additional payments and start the order again.
+	 *
+	 * @since [version]
+	 *
+	 * @param LLMS_Order $order The order object.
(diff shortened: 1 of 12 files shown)


includes/admin/tools/class-llms-admin-tool-limited-billing-order-locator.php
```diff
@@ -0,0 +1,285 @@
...
+				),
+			)
+		);
+
+		foreach ( $orders->posts as $order ) {
+
+			$order = llms_get_post( $order );
+			if ( ! $order || ! is_a( $order, 'LLMS_Order' ) ) {
+				continue;
+			}
```
eri-trabiccolo, line 64:
```suggestion
```
Don't see how could `$order` be a falsy or not an LLMS_Order, given how `llms_get_post()` works

includes/admin/tools/class-llms-admin-tool-limited-billing-order-locator.php
```diff
@@ -0,0 +1,285 @@
...
+		return $csv;
+
+	}
+
+	/**
+	 * Schedules payments and expiration for an order
+	 *
+	 * Retrieves orders from the `get_orders()` method and schedules a recurring payment
+	 * and expiration action based on its existing calculated order data.
+	 *
```
eri-trabiccolo: copypasta

includes/functions/llms-functions-wrappers.php
```diff
@@ -0,0 +1,158 @@
...
+		$options = wp_parse_args(
+			$options,
+			array(
+				'status' => 302,
+				'safe'   => true,
+			)
+		);
+
+		$func = $options['safe'] ? 'wp_safe_redirect' : 'wp_redirect';
+		call_user_func( $func, $location, $options['status'] );
```
eri-trabiccolo:
```suggestion
		$func( $location, $options['status'] );
```
:D

includes/functions/llms-functions-wrappers.php
```diff
@@ -0,0 +1,158 @@
...
+	 * Wrapper for WP core redirects which automatically calls `exit();`.
+	 *
+	 * This function is redefined when running phpunit tests to make testing code that redirects (and exits).
+	 *
+	 * @since 3.19.4
+	 * @since [version] Moved location from `includes/llms.functions.core.php`.
+	 *
+	 * @link https://github.com/gocodebox/lifterlms-tests/blob/472c5a286e9f65e2be0c1d6b7edd8d5340d052ed/framework/functions-llms-tests.php#L178-L199
+	 *
+	 * @param string $location Full URL to redirect to
```
eri-trabiccolo:
```suggestion
	 * @param string $location Full URL to redirect to.
```

includes/functions/llms-functions-wrappers.php
```diff
@@ -0,0 +1,158 @@
...
+	 * @param array  $options  {
+	 *     Optional. Array of options. Default is empty array.
+	 *     @type int  $status HTTP status code of the redirect. Default: `302`.
+	 *     @type bool $safe   If true, use `wp_safe_redirect()` otherwise use `wp_redirect()`. Default: `true`.
```
eri-trabiccolo:
```suggestion
	 * @param array  $options  {
	 *     Optional. Array of options. Default is empty array.
	 *
	 *     @type int  $status HTTP status code of the redirect. Default: `302`.
	 *     @type bool $safe   If true, use `wp_safe_redirect()` otherwise use `wp_redirect()`. Default: `true`.
	 * }
```

eri-trabiccolo (changes requested):

includes/admin/tools/class-llms-admin-tool-limited-billing-order-locator.php
```diff
@@ -0,0 +1,285 @@
...
+		$successes = $this->get_txn_count_by_status( $order, 'llms-txn-succeeded' );
+		$total     = $refunds + $successes;
+
+		$ended    = llms_parse_bool( $order->get( 'plan_ended' ) );
+		$expected = $order->get( 'billing_length' );
+
+		if ( $refunds >= 1 || ( $ended && $total !== $expected ) ) {
+
+			$id   = $order->get( 'id' );
+			$link = get_edit_post_link( $id );
```
eri-trabiccolo:
```suggestion
			$link = get_edit_post_link( $id, 'raw' );
```
or anything different from 'display', which is the default value, which will turn & into &amp; generating something like:
https://mywp.loc/wp-admin/post.php?post=1389&amp;action=edit
which is a "bad" link.
That said, the WordPress doc says:
Optional. How to output the '&' character. Default '&amp;'.
https://core.trac.wordpress.org/browser/tags/5.8/src/wp-includes/link-template.php#L1433
but the generated html doc:
(string) (Optional) How to output the '&' character. Default '&'.
so I guess it's not your fault :P
cc @pondermatic maybe something you want to "consider" for your work on the docblocks?

includes/admin/tools/class-llms-admin-tool-limited-billing-order-locator.php
thomasplevy: yea, you're right. Habit from using the function when we don't exactly know what the input is (as we do in this case)

includes/admin/tools/class-llms-admin-tool-limited-billing-order-locator.php
thomasplevy: Yes, you're correct about this. Nice.

I also noticed the unit tests weren't actually testing anything with relation to this due to having no logged in user. Whoop. Fixed that too!

includes/admin/tools/class-llms-admin-tool-limited-billing-order-locator.php
thomasplevy: always <3

thomasplevy:
@eri-trabiccolo noticed that the lookup for possibly affected orders would also include orders made *after* the upgrade. I've added an EXISTS check on the removed meta property (`_llms_date_billing_end`) to ensure that the tool doesn't pick up orders created *after* the upgrade. These newer orders shouldn't be included because A) the bug is fixed and payments won't be missed and B) If someone refunds the order moves to *refunded* and then they have to switch it back to active at which point they'd be aware of the number of remaining payments and change accordingly. 

Does my logic seem sound on this to you? This is primarily why I'm requesting a re-review.

includes/admin/tools/class-llms-admin-tool-limited-billing-order-locator.php
```diff
@@ -0,0 +1,286 @@
...
+			$csv = $this->generate_csv();
+			wp_cache_set( $this->id, $csv, 'llms_tool_data' );
+		}
+
+		return $csv;
+
+	}
+
+	/**
+	 * Generate the CSV file an server it as a downloadable attachment
```
eri-trabiccolo:
```suggestion
	 * Generate the CSV file an serve it as a downloadable attachment
```

eri-trabiccolo (changes requested):
Looks good.
The only "missing" thing is a way to exclude already reviewed orders.
Meaning that I run the tool and it finds 10 orders to be reviewed, then I review them and decide they're all fine. Back on Tools, and run it again - it'll find the same orders.
Not a big deal I guess.

thomasplevy:
@eri-trabiccolo any thoughts on the ideal workflow for removing an order from a report... You're right that there's no way to do this and I hadn't thought about that... Other than building a UI to "mark as okay" somewhere on the view order admin ui I don't really have any thoughts on what to do here... Ideally a user would review this once only and never need to again but I don't know how to get rid of it once they're happy in that case...

Any thoughts?

eri-trabiccolo:
> @eri-trabiccolo any thoughts on the ideal workflow for removing an order from a report... You're right that there's no way to do this and I hadn't thought about that... Other than building a UI to "mark as okay" somewhere on the view order admin ui I don't really have any thoughts on what to do here... Ideally a user would review this once only and never need to again but I don't know how to get rid of it once they're happy in that case...
> 
> Any thoughts?

I thought about:
1) save the "is reviewed" in a order post meta, like a checkbox, which updates via ajax - so it doesn't require a submit button to be clicked. But this should appear only on the ones which need to be reviewed...
Maybe clumsy.
2) buttons next to the tool:
- mark all as reviewed: adds a not autoloaded option
- once this option is set ideally you won't see the tool again (we can think about having a constant to make the tool appear again...). Maybe still clumsy...
What do you think?

thomasplevy:
I think this means writing a lot of code and tests and further delaying the release of a bugfix that should have been shipped a week ago at least...

What if we update language of the tool to say "You can only perform this action once" and then set a non-autoloading option after the CSV is downloaded which hides the tool (forever)?

This probably isn't a good solution because someone will inevitably lose the CSV or something will go wrong and the option will get set but the CSV won't be downloaded or they'll download it and have incorrect file permissions set on their PC and they won't be able to access it or something...

If we don't touch this at all... just the leave the tool as is and ship, what happens? Users review and they say "Everything's good" but the tool still shows up... does anyone go there and get upset about this? If they do we tell them "You already reviewed it, it's okay" and they say "But I don't want to the see the tool anymore, it's confusing" and we say "We've explained what happened, it doesn't go away, you'll just have to ignore it"

Is this acceptable? I don't know.... I really don't want to write and test anymore code for this though. I want to ship the bugfix

eri-trabiccolo:
@thomasplevy 
None of the solutions is the best, so yeah let's go with the current implementation, and yeah we'll tell people it's fine it shows up.

PR #1761, opened 2021-08-26
Add `LLMS_Abstract_Options_Data` to `LLMS_Payment_Gateway` abstract  

thomasplevy:
## Description

Improves a gateway's ability to interact with it's own options primarily by adding the `set_option()` method.

## How has this been tested?

+ Manually
+ New tests

## Types of changes

## Checklist:

### DIFF
--- a/includes/abstracts/abstract.llms.payment.gateway.php
+++ b/includes/abstracts/abstract.llms.payment.gateway.php
@@ -375,7 +375,7 @@ abstract class LLMS_Payment_Gateway extends LLMS_Abstract_Options_Data {
 	 *
 	 * @since 3.30.0
 	 *
-	 * @param LLMS_Order $order The order object
+	 * @param LLMS_Order $order The order object.
 	 * @return string
 	 */
 	protected function get_complete_transaction_redirect_url( $order ) {
--- a/includes/abstracts/abstract.llms.payment.gateway.php
+++ b/includes/abstracts/abstract.llms.payment.gateway.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Abstracts/Classes
  *
  * @since 3.0.0
- * @version 4.0.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -18,8 +18,9 @@ defined( 'ABSPATH' ) || exit;
  * @since 3.34.3 During order completion, use `llms_redirect_and_exit()` instead of `wp_redirect()` and `exit()`.
  * @since 3.37.18 Allow redirection to external domains by disabling "safe" redirects.
  * @since 4.0.0 Removed deprecated completed transaction message parameter output.
+ * @since [version] Extend LLMS_Abstract_Options_Data for improved options interactions.
  */
-abstract class LLMS_Payment_Gateway {
+abstract class LLMS_Payment_Gateway extends LLMS_Abstract_Options_Data {
 
 	/**
 	 * Optional gateway description for the admin panel
@@ -101,6 +102,13 @@ abstract class LLMS_Payment_Gateway {
 	public $logging_enabled = '';
 
 	/**
+	 * Option name prefix.
+	 *
+	 * @var string
+	 */
+	protected $option_prefix = 'llms_gateway_';
+
+	/**
 	 * Array of supported gateway features
 	 *
 	 * @var      array
@@ -150,6 +158,13 @@ abstract class LLMS_Payment_Gateway {
 	 * @since  3.0.0
 	 */
 	public $title = '';
+
+	/**
+	 * Option's data version
+	 *
+	 * @var integer
+	 */
+	protected $version = 2;
 
 	/**
 	 * This should be called by the gateway after verifying the transaction was completed successfully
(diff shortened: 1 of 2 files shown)


includes/abstracts/abstract.llms.payment.gateway.php
```diff
@@ -713,41 +728,76 @@ public function log() {
...
+	 * The behavior of this function differs slightly from the parent method in that the second argument
+	 * in this method allows lookup of the secure key value.
+	 *
+	 * Default options are autoloaded via the get_option_default_value() method.
+	 *
+	 * @since 3.0.0
+	 * @since 3.29.0 Added secure option lookup via option second parameter.
+	 *
+	 * @param string $key option Option name / key, eg: "title".
+	 * @param string $secure_key Secure option name / key, eg: "TITLE".
```
eri-trabiccolo, line 804:
```suggestion
	 * @param string $secure_key Optional. Secure option name / key, eg: "TITLE". Default is `false`.
```

includes/abstracts/abstract.llms.payment.gateway.php
```diff
@@ -713,41 +728,76 @@ public function log() {
...
+	 *
+	 * @param mixed  $default_value        The default value. If no value is passed to `get_option()`, this will be an empty string.
+	 *                                     Otherwise it will be the default value passed to the method.
+	 * @param string $full_option_name     The full (prefixed) option name.
+	 * @param bool   $passed_default_value Whether or not a default value was passed to `get_option()`.
+	 * @return mixed The default option value.
+	 */
+	public function get_option_default_value( $default_value, $full_option_name, $passed_default_value ) {
+		$unprefixed = str_replace( $this->get_option_prefix(), '', $full_option_name );
+		return isset( $this->$unprefixed ) ? $this->$unprefixed : '';
```
eri-trabiccolo, line 848: Does this mean that we always return an empty string if `$this->$unprefixed` is not set, whatever is the passed default value to `get_option()`? Is this intended?

eri-trabiccolo (changes requested):

includes/abstracts/abstract.llms.payment.gateway.php
thomasplevy: We've been doing this a bit recently (I think I started it) but I realized just now that we are creating redundant text on the developer site when we do this.

Have a look at this generated doc: https://developer.lifterlms.com/reference/functions/llms_get_post/

The "Optional" and "Default Value" are automatically added to the generated doc without the need for us to write it. It can be inferred and we don't need to write it.

If we write the same doc this way it'll end up saying optional and default value twice as a result of the automatically added stuff.

So... we should stop writing docs like this to prevent duplications on the doc site... I didn't realize this was auto-added...

includes/abstracts/abstract.llms.payment.gateway.php
thomasplevy: It's not exactly intentional but it is a result of poor foresight (by me) when I introduced `LLMS_Payment_Gateway::get_option()`.

The *current* version of the class *requires* every option to be defined as a class variable. If you want to have `llms_gateway_{id}_title` as an option, then you need to define `public $title = 'default val'` on the gateway class.

The current version will throw an error if the option isn't defined: 

https://github.com/gocodebox/lifterlms/blob/8c913be8492b8216f8657ae2864ac0411c8d8700/includes/abstracts/abstract.llms.payment.gateway.php#L724-L736

So the change I'm introducing is to return an empty string if the variable isn't set on the class.

If you call `get_option()` directly you can pass in a default value and this won't run as the callback is only hooked (and unhooked) when calling `LLMS_Abstract_Options_Data::get_option()` -- https://github.com/gocodebox/lifterlms/blob/8c913be8492b8216f8657ae2864ac0411c8d8700/includes/abstracts/llms.abstract.options.data.php#L60-L66

So... this *doesn't change* any existing gateways that are already working "as expected" by having a default value defined as a var on the class but enables gateways to remove these variables (if they want) and to define their own autloader of default values if they choose.

Ideally there's more changes to make here:

The default value should *automatically be loaded from the settings array* and the secure key option should be automatically used (instead of requiring it to be passed into get_option()` but I haven't figured out if that's truly necessary yet...

Mostly: I have built and worked on all our gateways and everytime I work on one I'm mad that in order to *update an option programatically* I have to write : `update_option( $this->get_option_name( ... ), val )`

thomasplevy: At least this was my intention... I think this works as I intended, right? Do you see a logical flaw or error here that I'm missing?

includes/abstracts/abstract.llms.payment.gateway.php
eri-trabiccolo: Riiight. <3
You know. WP does it here and there. Ok let's stop.

thomasplevy: I'm actually not sure I'm right about it creating duplicates... I'm digging through code and it's a waste of my time right now... I don't want duplicated content on the frontend but it looks the WP core doesn't have duplicates when they do this so I can't explain the source or reasoning here... it's possibly an upstream feature coming from phpdocumentor itself... maybe if it detects default and optional it doesn't include it somehow... I'm not sure... Either way I think we're fine and I'm not going to change anything at this moment...

includes/abstracts/abstract.llms.payment.gateway.php
eri-trabiccolo: I think it works as expected yeah. Thanks for the explanation.

eri-trabiccolo (approved):

PR #1770, opened 2021-08-31
Fix duplicate last-name block in reusable block schema

### LINKED ISSUE #1765: Default reusable User name block has two last-name block fields rather than a first-name and last-name block field.
https://github.com/gocodebox/lifterlms/blob/5.2.1/includes/schemas/llms-reusable-blocks.php#L57-L74

Schermata 2021-08-30 alle 18 54 48

This might be minor actually, since field id/name are actually correct.

```
LifterLMS 5.2.1
```

## This issue has been recreated:
+ [x] Locally
+ [x] On a staging site
+ [x] On a production website

n/a

thomasplevy:
## Description

Fixes #1765 

I'm opting not to write a database migration to change this. There's no functional bug as the correct IDs, names, and labels are being used and given how it is "installed" the actual block name / block being used doesn't matter in this case.

If someone *sees* it's the wrong block the work to change it is relatively small (delete, recreate, and reapply any language customizations applied to the block).

I think writing a migration to identify the block and update it is going to be time consuming and we introduce additional points of failure / technical debt that might outweigh the actual visual issue caused by this bug.

## How has this been tested?

+ Manually
+ Unit tests

## Types of changes
+ Bugfix

## Checklist:


pondermatic (approved):
Perhaps we should have a post in the LifterLMS Knowledge base about how to edit forms with the WordPress/Gutenberg code editor and change the block type from `wp:llms/form-field-user-last-name` to `wp:llms/form-field-user-first-name`.

PR #1771, opened 2021-09-01
Refactor `LLMS_Engagements`

### LINKED ISSUE #290: delayed engagements may still fire after engagement has been deleted
Recreation steps:

1. Create an engagement (eg email) with a delay of 1 or more days -- try lesson completion
2. Trigger engagement via a student (eg student completes the lesson)
3. Delete the engagement
4. Email will still send

Fix:

When the engagement post is deleted the scheduled actions should be removed

### LINKED ISSUE #859: Fix spelling error in LLMS_Database_Query abstract method `preprare_query()`
The abstract method `preprare_query` should be spelled `prepare_query`

We cannot simply rename the function without causing potential with add-ons (known and unknown).

We cannot deprecate it by replacing the abstract method with a concrete method that calls a replacement abstract method as per the comment here: https://github.com/gocodebox/lifterlms/pull/722#issuecomment-446750779

I'm not entirely certain how to fix this issue long term or even short term.

See discussion in closed PR #722

### LINKED ISSUE #1027: Add utility function for escaping and quoting a string.
This function is used, primary, as an `array_map()` call back function to escape/sanitize and then wrap quotes around strings which will usually be joined into a comma-separated list that can be dropped into an `IN()` condition in SQL queries.

I've found myself recreating this class method in several LifterLMS add-ons so it would be generally useful to convert this class method into a core function that can be used without having to instantiate a class to access it.

EG: `llms_esc_and_quote_str()`.

https://developer.lifterlms.com/reference/classes/llms_database_query/escape_and_quote_string/

### LINKED ISSUE #1081: Update default images

### LINKED ISSUE #1082: Reduce code redundancy of code shared between certificates and achievements
Since certificates and achievements are quite similar in how they are created, earned, generated, etc... we can reduce duplicate code by creating a shared abstract base class that can be utilized by both elements.

The class `LLMS_Certificate` is only ever used by `LLMS_Certificate_User` (which extends it)

The fact that this hierarchy exists was created to allow other types of certificates to exist but in almost 6 years we've never added any additional certificate types and I don't see us adding any other. Maybe we'll add group certificates to the group but lets not worry about that hypothetical.

I propose we deprecate *both* of these classes in favor of moving this functionality into a new model `LLMS_Certificate_Template` model. This will function like an `LLMS_Post_Model` class.

A good deal of the code in the base class `LLMS_Certificate` is the definition of public properties that aren't used or aren't that useful.

The code in `LLMS_Certificate_User` is largely centered around generating an earned certificate.

I propose the functionality around generating an earned certificate be moved into the existing `LLMS_Certificate_User` model.

Most of this is true for the relationship between `LLMS_Achievement` and `LLMS_Achievement_User`. Similar updates could be made to create a new model `LLMS_Achievement_Template` and moving functionality into `LLMS_User_Achievement` model

After a review of all of the code in these 6 classes there is a lot of code that could be shared between them. The best course of action, I think, would be to create a trait (or several traits) that can be mixed into the appropriate classes.

### LINKED ISSUE #1640: Certificate {first_name} merge code fallback to user nickname is confusing
### Reproduction Steps

+ Create a certificate using the {first_name} merge code
+ Award the cert on user registration
+ Enable open registration and make first / last name optional
+ Register a user without a first name

### Expected Behavior

+ the {first_name} would be blank on the certificate

### Actual Behavior

+ The user's "nickname", which defaults to the user's username is filled as a fallback in favor of the first name.

### More information

The actual behavior is actually *expected* based on how the certificate is programmed but this has been raised as a bug enough times in support that I think we should change the actual behavior to match the expected behavior.

See: https://github.com/gocodebox/private-issues/issues/25 & there's another issue somewhere about this same thing but I can't find it

thomasplevy:
## Description

Refactors a significant portion of certificate and engagement related code, deprecates a lot of code and simplifies the process of generating earned certificates and achievements.

+ Fixes #290
+ Fixes #859 
+ Closes #1027
+ Closes #1081
+ Closes #1082
+ Fixes #1640 
+ includes https://github.com/gocodebox/lifterlms/pull/1838
+ See also #1845, #1846, #1859

## TODO

+ [x] Write achievement tests (there aren't any or possibly any)
+ [x] Replace metadata with post thumbnail for both post types (deprecate the meta properties how certificate title is deprecated)
+ [x] Deprecate achievement title & content like certificate title is deprecated
+ [x] Add extra metadata to achievements (related, engagement, etc...)
+ [x] Is certificate content properly / fully transitioned from metadata to the post content?
+ [x] Rewrite queries on the `LLMS_Student` model to use a WP_Query. We might have to deprecate the existing methods and create new ones, I haven't looked at this too thoroughly yet. Possibly create custom class for this, like `LLMS_Engagements_Query`
+ [x] Deprecate the `LLMS_Achievments::trigger_engagement()` class (like certificates has been deprecated)
+ [ ] More tests... I wrote a lot, there's more to write.
+ [ ] Do realtime testing as a user
+ [ ] Did I break e2e tests completely?
+ [x] Merge this into #1838 or merge that into this
+ [ ] Gather deprecations and write to changelog/s
+ [x] Write and test database migrations (my_cert and my_achievement) post/meta data
   + [x] postmeta title & content => post title & content
   + [x] Featured image meta => thumbnail
   + [x] store user as post_author
   + [x] store template as post_parent
   + [x] maybe more?
+ [x] Update achievements metabox to remove (deprecated) image (@eri-trabiccolo)

## How has this been tested?

+ New tests
+ Existing tests
+ Manually (various tests with delayed & immediate engagments for emails, achievements, and certs)

## Types of changes

+ Bug fix
+ Refactor existing code

## Checklist:


includes/class.llms.engagements.php
```diff
@@ -46,11 +43,15 @@ class LLMS_Engagements {
...
 	 *
+	 * @since 2.3.0
+	 * @since [version] Added deprecation warning when using constant `LLMS_ENGAGEMENT_DEBUG`.
+	 *
 	 * @return void
 	 */
 	private function __construct() {
 
 		if ( defined( 'LLMS_ENGAGEMENT_DEBUG' ) && LLMS_ENGAGEMENT_DEBUG ) {
+			_deprecated_function( 'Constant: LLMS_ENGAGEMENT_DEBUG', '[version]' );
```
eri-trabiccolo, line 55: let's remember to modify this doc:
https://lifterlms.com/docs/enable-engagement-debug-logging/

includes/class.llms.engagements.php
```diff
@@ -65,15 +66,148 @@ private function __construct() {
...
+		foreach ( $this->get_trigger_hooks() as $action ) {
+			add_action( $action, array( $this, 'maybe_trigger_engagement' ), 777, 3 );
+		}
+
+		add_action( 'lifterlms_engagement_send_email', array( $this, 'handle_email' ), 10, 1 );
+		add_action( 'lifterlms_engagement_award_achievement', array( $this, 'handle_achievement' ), 10, 1 );
+		add_action( 'lifterlms_engagement_award_certificate', array( $this, 'handle_certificate' ), 10, 1 );
+
+		add_action( 'deleted_post', array( $this, 'unschedule_delayed_engagements' ), 20, 2 );
+		add_action( 'trashed_post', array( $this, 'unschedule_delayed_engagements' ), 20 );
```
eri-trabiccolo: what do we do when the related post (course/lesson/access plan etc.) is trashed/deleted?
I'd say that we should unschedule the engagement as well, although I guess we don't have a way to determine what's the scheduled action to unschedule though...

We could, when processing the delayed engagements, at least avoid the certificates/achievements to be earned if the related post doesn't exist anymore. What do you think?

includes/class.llms.engagements.php
```diff
@@ -229,280 +346,293 @@ public function handle_email( $args ) {
...
+	 * @param string     $action          Name of the triggering action hook.
+	 * @param int|string $related_post_id WP_Post ID of the related post or an empty string.
+	 * @return string
 	 */
-	public function maybe_trigger_engagement() {
+	private function parse_hook_find_trigger_type( $action, $related_post_id ) {
 
-		$action = current_filter();
-		$args   = func_get_args();
+		$trigger_type = false;
```
eri-trabiccolo:
```suggestion
		$trigger_type = '';
```

includes/class.llms.engagements.php
```diff
@@ -229,280 +346,293 @@ public function handle_email( $args ) {
...
 
-					case 'certificate':
-						/**
-						 * @todo Fix this
-						 * if there's no related post id we have to send one anyway for certs to work
-						 * this would only be for registration events @ version 2.3.0
-						 * we'll just send the engagement_id twice until we find a better solution
-						 */
-						$related_post_id = ( ! $related_post_id ) ? $e->engagement_id : $related_post_id;
+		if ( ! in_array( $engagement->event_type, array_keys( llms_get_engagement_types() ), true ) ) {
```
eri-trabiccolo:
```suggestion
		if ( ! array_key_exists( $engagement->event_type, llms_get_engagement_types() ) ) {
```

includes/class.llms.engagements.php
```diff
@@ -229,280 +346,293 @@ public function handle_email( $args ) {
...
+			 *
+			 * @param array $parsed {
+			 *     An associative array of parsed data used to trigger the engagement.
+			 *
+			 *     @type string $handler_action (Required) Hook name of the action that will handle awarding the sending the engagement.
+			 *     @type array  $handler_args   (Required) Arguments passed to the `$handler_action` callback.
+			 * }
+			 * @param object $engagement      The engagement object from the `get_engagements()` query.
+			 * @param int    $user_id         WP_User ID who will be awarded the engagement.
+			 * @param int    $related_Post_id WP_Post ID of the related post.
```
eri-trabiccolo:
```suggestion
			 * @param int    $related_post_id WP_Post ID of the related post.
```

includes/class.llms.engagements.php
eri-trabiccolo: On a second thought, we should NOT delete the scheduled action when an engagement is just trashed - b/c one could trash it by error and rapidly untrash/publish it again, but with this PR all the attached delayed actions will be just deleted with no way to restore them.
I'd propose to just check the status of the engagement when processing the delayed action, and skip it if the engagement is not `publish`. Makes sense, right?

eri-trabiccolo (changes requested):
I think it's overall good.
Although I think we should avoid sending delayed engagements when the related post has been trashed/deleted, and when the engagement status is not `publish`.

includes/class.llms.engagements.php
thomasplevy: You make a good point. Since I've already written the code to unschedule actions I'm going to leave it in form *deleted* posts (one that has been trashed and then emptied from the trash). This, I believe, makes sense -- there's no way to get it back so it's okay to unschedule the action completely.

For trashed posts, I've adjusted the handlers to check the status of the engagement and processing will be skipped for any status of the triggering engagement post other than `publish`. That means if the engagement has been set to `draft` or moved to the trash will be skipped when they're processed.

thomasplevy: Updated in https://github.com/gocodebox/lifterlms/pull/1771/commits/300dd8501f6b1b2fb66ca8b14bd2001240f245e8

eri-trabiccolo (changes requested):
Looks good, although, and sorry for being nitpick, I think we still miss this part:

> what do we do when the related post (course/lesson/access plan etc.) is trashed/deleted?
> I'd say that we should unschedule the engagement as well, although I guess we don't have a way to determine what's the scheduled action to unschedule though...
> 
> We could, when processing the delayed engagements, at least avoid the certificates/achievements to be earned if the related post doesn't exist anymore. What do you think?

Right?

And I think we should check also that the engagement type is `publish`, I mean the achievement/email/certificate template actually exist and is publish otherwise what do we "send" (we might end up producing a fatal btw)?

includes/class.llms.post-types.php
```diff
@@ -1040,6 +1041,21 @@ public static function register_post_types() {
...
+		 * Intermittent issues that I can't quite recreate seem to crop up as a result of these
+		 * post types not being registered.
+		 *
+		 * Additionally, it's semantically incorrect that these post types *don't exist* so
+		 */
+		$earned_engagement_args = array(
+			'public'             => false,
+			'publicly_queryable' => true,
+		);
+		self::register_post_type( 'llms_my_certificate', $earned_engagement_args );
```
eri-trabiccolo: this is already registered:
https://github.com/gocodebox/lifterlms/blob/300dd8501f6b1b2fb66ca8b14bd2001240f245e8/includes/class.llms.post-types.php#L800
And we use to have it `public`, to allow people see and share it.

thomasplevy:
@eri-trabiccolo, forgive me for what I have done.

eri-trabiccolo:
man, I think we should decide the standard nomenclature, at least for file/class/variable names:
award/awarded vs earn/earned.

includes/admin/views/metaboxes/view-award-engagement-submit.php
```diff
@@ -0,0 +1,159 @@
...
+				if ( $can_publish ) :
+					if ( ! empty( $engagement->post_date_gmt ) && llms_current_time( 'U', true ) < strtotime( $engagement->post_date_gmt . ' +0000' ) ) :
+						?>
+						<input name="original_publish" type="hidden" id="original_publish" value="<?php echo esc_attr_x( 'Schedule', 'post action/button label', 'lifterlms' ); ?>" />
+						<?php submit_button( _x( 'Schedule', 'post action/button label', 'lifterlms' ), 'primary large', 'publish', false ); ?>
+						<?php
+					else :
+						?>
+						<input name="original_publish" type="hidden" id="original_publish" value="<?php esc_attr_e( 'Publish', 'lifterlms' ); ?>" />
+						<?php submit_button( __( 'Publish', 'lifterlms' ), 'primary large', 'publish', false ); ?>
```
thomasplevy:
```suggestion
						<?php submit_button( __( 'Award', 'lifterlms' ), 'primary large', 'publish', false ); ?>
```

includes/admin/post-types/meta-boxes/class.llms.meta.box.award.engagement.php
```diff
@@ -0,0 +1,225 @@
+<?php
```
thomasplevy: I think this whole metabox could actually be removed and the student select / output could go into the "award" (submit) box. Kind of like how the WP core will have the "author" in that area.

Looking at the big full-width metabox for this field seems a bit silly in reality, right?

thomasplevy (commented):
@eri-trabiccolo some (small) stuff I noticed

thomasplevy:
@eri-trabiccolo added some tasks on the PR description for you

thomasplevy:
@eri-trabiccolo 

I'm thinking about some things around syncing earned certs from their original template...

I wrote the sync method already, we need to expose a button to allow syncing (when syncing is possible). Syncing is possible when the my_cert post has a published and still existing `llms_certificate` associated with it through the `_llms_certificate_template` meta field.

Here's the sync method: https://github.com/gocodebox/lifterlms/blob/5e9309620c885d2b8365f9d6449f8314abb2bdc6/includes/models/model.llms.user.certificate.php#L398-L416

I'm thinking this could be a button next to the "Update" button in the submit box. The button could probably pretty simply be a nonce link that runs the sync() method and reloads the page, I think...

I'm also thinking about bulk syncing from a template to all of the my_certs generated from that certificate. Let's say a user updates the template and they want that change to be reflected in all already existing certs... They could click a button from the certificate template editor to bulk update all generated certs. This button should probably push the syncing to an async background process. When pressing this button the user should be asked to confirm that X certificates will be regenerated and the process cannot be undone and etc...

~This last part has me thinking that possibly we need to version control the earned certificates somehow. Maybe the sync will pull a timestamp (the templates modified date) and store that as a meta value (`_llms_template_sync`). When running a sync we could query generated certs that haven't been updated since the last sync.......~ [This is too much... I'm not thinking through that any further] 

Thinking more about this this probably all applies to achievements too

What do you think? Do you want to get started on this?

includes/admin/views/metaboxes/view-award-engagement-submit.php
eri-trabiccolo: @thomasplevy 
I agree with you, I talked about this here:
https://github.com/gocodebox/lifterlms/pull/1845#issuecomment-961430826

Here's a video to see what I was talking about  https://www.loom.com/share/d496770d23f540e7914a37fff78b4aa5
Those strings are in the js:
https://github.com/WordPress/WordPress/blob/master/wp-admin/js/post.js#L742

I was planning to implement our own js basically just for this, but I wanted to know if you thought it was worthwhile.
Let me know.

Ohhh, I think we can leverage the js hooks:
https://github.com/WordPress/WordPress/blob/abde9678b49d9f9b9e7521efc9075ea938369c97/wp-includes/js/dist/i18n.js#L1336
I'd go this way... Do you agree?

includes/admin/post-types/meta-boxes/class.llms.meta.box.award.engagement.php
eri-trabiccolo: You're right, although the WP core author is in a separate meta box not that different from this one (with the classic editor).
 see #1851

eri-trabiccolo:
@thomasplevy 
Awarded engagements generated from templates with default background images don't have the featured image set (if you edit them you'll see an empty featured image field) - I guess we could extend this to `llms_my_[achievement|certificate]`:
https://github.com/gocodebox/lifterlms/pull/1771/files#diff-5bd5f18e4668bbde1bb76ec64f66a9cc59116193938bf7bfbba2a9e28ed8c1faR207

Although this means that changing the default image after an engagement (using it) has been awarded will affect them too. I guess this is something acceptable?

thomasplevy:
@eri-trabiccolo 

> @thomasplevy Awarded engagements generated from templates with default background images don't have the featured image set (if you edit them you'll see an empty featured image field) - I guess we could extend this to `llms_my_[achievement|certificate]`: https://github.com/gocodebox/lifterlms/pull/1771/files#diff-5bd5f18e4668bbde1bb76ec64f66a9cc59116193938bf7bfbba2a9e28ed8c1faR207
> 
> Although this means that changing the default image after an engagement (using it) has been awarded will affect them too. I guess this is something acceptable?

Good call, see https://github.com/gocodebox/lifterlms/pull/1771/commits/907c94f22f3839e3e1a1ff0321273de3e22c2a25

includes/traits/llms-trait-earned-engagement-reporting-table.php
```diff
@@ -0,0 +1,64 @@
...
+		$student = false;
+		if ( ! empty( $this->student ) ) {
+			$student = $this->student->get_id();
+		} elseif ( ! empty( $_GET['student_id'] ) ) { //phpcs:ignore -- Nonce verification not needed.
+			$student = llms_filter_input( INPUT_GET, 'student_id', FILTER_SANITIZE_NUMBER_INT );
+		}
+
+		$post_new_file  = "post-new.php?post_type=$post_type";
+		$post_new_url   = esc_url( add_query_arg( 'sid', $student, admin_url( $post_new_file ) ) );
+		$add_new_button = '<a style="display:inline-block;margin-bottom:20px" href="' . $post_new_url . '" class="page-title-action">' . esc_html( $post_type_object->labels->add_new ) . '</a>';
```
thomasplevy: This should be an llms button (like all the other buttons on reporting screens), eg

`<a class=llms-button-primary" ...`

thomasplevy:
@eri-trabiccolo 

While we're working on this two new things have occurred to me:

1. The slug `my_certificate` is a rather disappointing slug. I think it should be `certificate` but we currently use `certificate` for the template preview URLs (for admins). I don't think it's wise to change the slug on existing certificates (especially since adding the sharing options). But... now would be a good time to change the slugs. I think we could change `certificate` to `certificate-template` with absolutely no repercussions since it's highly unlikely that these admin links are bookmarked anywhere. I'd like to change `my_certificates` but I'm not sure the best way to change a slug in a backwards compatible matter without adding an option for it (which doesn't seem like it's necessary) so maybe we should just leave it alone. But do you have any thoughts or ideas that could help?

2. The automatically generated slug for earned certificates via [wp_unique_post_slug()](https://developer.wordpress.org/reference/functions/wp_unique_post_slug/) is arguably problematic since it results in a predictable url being generated. I'm thinking we could use the pre filter in the function (https://developer.wordpress.org/reference/hooks/pre_wp_unique_post_slug/) to create our own unique slug for earned certs. But... what method should be used to create this? Part of me thinks that we could disregard the sanitized title all together and create a pseudocrpyt hash using `LLMS_Hasher` and pass in the WP_Post ID so that it's guaranteed to be unique, can be unhashed, but not easily guessed (like sequential numbers). We do this currently for the quiz attempt key/ids. It's a number passed into the hasher and it spits out a alphanumeric string. This is what I'm thinking, what do you think?

eri-trabiccolo:
> @eri-trabiccolo
> 
> While we're working on this two new things have occurred to me:
> 
> 1. The slug `my_certificate` is a rather disappointing slug. I think it should be `certificate` but we currently use `certificate` for the template preview URLs (for admins). I don't think it's wise to change the slug on existing certificates (especially since adding the sharing options). But... now would be a good time to change the slugs. I think we could change `certificate` to `certificate-template` with absolutely no repercussions since it's highly unlikely that these admin links are bookmarked anywhere. I'd like to change `my_certificates` but I'm not sure the best way to change a slug in a backwards compatible matter without adding an option for it (which doesn't seem like it's necessary) so maybe we should just leave it alone. But do you have any thoughts or ideas that could help?

Ok for `certificate-template` I would really love to change `my_certificate` slug as well... What if we do something like:
1) a db update to insert an hidden option that stores if there are shared certificates (which is a post meta).
2) then
```php
/**
 * @deprecated - keep it for 1 year.
 */
add_action(
	'some_action_I_dont_know',
	function () {
		if ( option_my_certificates_using_old_slug_is_true() && url_contains( $url, 'my_certificate') ) {
			redirect_to_new_awarded_certificate_url( $url );
		}
	}
);
```
And in one year from now we'll remove that code above and the option.
We could even start saving the `allow_sharing` meta with another name and schedule a check (monthly?) on whether the former ` `allow_sharing` exists for at least one certificate, and if not remove that option and unschedule the check.

> 2. The automatically generated slug for earned certificates via [wp_unique_post_slug()](https://developer.wordpress.org/reference/functions/wp_unique_post_slug/) is arguably problematic since it results in a predictable url being generated. I'm thinking we could use the pre filter in the function (https://developer.wordpress.org/reference/hooks/pre_wp_unique_post_slug/) to create our own unique slug for earned certs. But... what method should be used to create this? Part of me thinks that we could disregard the sanitized title all together and create a pseudocrpyt hash using `LLMS_Hasher` and pass in the WP_Post ID so that it's guaranteed to be unique, can be unhashed, but not easily guessed (like sequential numbers). We do this currently for the quiz attempt key/ids. It's a number passed into the hasher and it spits out a alphanumeric string. This is what I'm thinking, what do you think?

Using `LLMS_Hasher` can be enough to me, but the question is... why?
Are you saying that we want to make **shared** certificates URLs less predictable, is this your concern?

includes/traits/llms-trait-earned-engagement-reporting-table.php
eri-trabiccolo: done in #1851 
I decided to keep it `small` too.

eri-trabiccolo:
> @eri-trabiccolo
> 
> I'm thinking about some things around syncing earned certs from their original template...
> 
> I wrote the sync method already, we need to expose a button to allow syncing (when syncing is possible). Syncing is possible when the my_cert post has a published and still existing `llms_certificate` associated with it through the `_llms_certificate_template` meta field.
> 
> Here's the sync method:
> 
> https://github.com/gocodebox/lifterlms/blob/5e9309620c885d2b8365f9d6449f8314abb2bdc6/includes/models/model.llms.user.certificate.php#L398-L416
> 
> I'm thinking this could be a button next to the "Update" button in the submit box. The button could probably pretty simply be a nonce link that runs the sync() method and reloads the page, I think...
> 
> I'm also thinking about bulk syncing from a template to all of the my_certs generated from that certificate. Let's say a user updates the template and they want that change to be reflected in all already existing certs... They could click a button from the certificate template editor to bulk update all generated certs. This button should probably push the syncing to an async background process. When pressing this button the user should be asked to confirm that X certificates will be regenerated and the process cannot be undone and etc...
> 
> ~This last part has me thinking that possibly we need to version control the earned certificates somehow. Maybe the sync will pull a timestamp (the templates modified date) and store that as a meta value (`_llms_template_sync`). When running a sync we could query generated certs that haven't been updated since the last sync.......~ [This is too much... I'm not thinking through that any further]
> 
> Thinking more about this this probably all applies to achievements too
> 
> What do you think? Do you want to get started on this?

I think this is something we can do, but I think that It should not be a button next to the Update one, also because there's already the Delete link there.
I would rather place it above:
Schermata 2021-11-11 alle 10 40 39

with an edit link (opening in a new tab) to the template, also it should open an alert that warns the user that the current content/title(/image) will be lost. Also the background image should be synced as well while the awarded date won't change.
What do you think?

thomasplevy:
> I think this is something we can do, but I think that It should not be a button next to the Update one, also because there's already the Delete link there. I would rather place it above: Schermata 2021-11-11 alle 10 40 39
> 
> with an edit link (opening in a new tab) to the template, also it should open an alert that warns the user that the current content/title(/image) will be lost. Also the background image should be synced as well while the awarded date won't change. What do you think?

Agree on all fronts :+1:

thomasplevy:
@eri-trabiccolo can you help me find deprecation warnings showing up in various places as a result of old meta values being used and help me update them.

I see some on reporting tables and probably they're showing up on the frontend student dashboard. I think theres probably still some showing up in notifications too.

(After you work on the other stuff we've already talked about)

thomasplevy:
@eri-trabiccolo I noticed that awarded achievements are showing the classic editor in favor of the textarea metabox

eri-trabiccolo:
@thomasplevy going to fix it, thanks <3

PR #1776, opened 2021-09-08
Delete empty notices and do not output them.

### LINKED ISSUE #1560: Delete empty (malformed) admin notices
### Reproduction Steps

This is something I've found when troubleshooting a customer's issue.
They needed to use the Reset Automatic Payments Status tool in order to re-enable the recurring payments feature.
The problem was that the notice which prompts the user to choose whether to re-enable them or not was not shown.
I've investigated and found that the reason was that the notice was already present in the db (so it was not re-added) but it was empty (I couldn't find the reason why it was empty, nor I could reproduce it on the customer's website after removing and re-generating it). This prevented it to be shown, but also to be re-added.

### Expected Behavior

+ I'd expect that an empty notice is deleted so it can be added again.

### Actual Behavior

+ The empty notice is not shown but not deleted so it cannot be re-added. This is particularly bad for the recurring payments feature as that notice is the only way (via UI, so excluding the definition of constants) to toggle the feature.

We already treat a similar case here:
https://github.com/gocodebox/lifterlms/blob/4.19.0/includes/admin/class.llms.admin.notices.php#L306-L308

but we don't account for a totally empty notice :(
Also, we should return after deleting the notice...

### Error Messages / Logs

```
n/a
```

```
LifterLMS 4.19.0
WordPress 5.7.0

```

+ [ ] Locally 
+ [ ] On a staging site
+ [x] On a production website

n/a

### Related User Information

HS-158371

pondermatic:
## Description
Combined the checks for an empty notice array, empty html, and empty template file into one block that deletes the notice and returns without outputting a notice.

Fixes #1560.

## How has this been tested?
Wrote a new unit test for `LLMS_Admin_Notices->output_notice()`.

## Types of changes

## Checklist:

### DIFF
--- a/includes/admin/class.llms.admin.notices.php
+++ b/includes/admin/class.llms.admin.notices.php
@@ -290,6 +290,7 @@ class LLMS_Admin_Notices {
 	 * @since 3.0.0
 	 * @since 3.7.4 Unknown.
 	 * @since 5.2.0 Ensure `template_path` and `default_path` are properly passed to `llms_get_template()`.
+	 * @since [version] Delete empty notices and do not display them.
 	 *
 	 * @param string $notice_id Notice id.
 	 * @return void
@@ -300,14 +301,13 @@ class LLMS_Admin_Notices {
 
 			$notice = self::get_notice( $notice_id );
 
-			if ( empty( $notice ) ) {
-				return;
-			}
-
 			// Don't output those rogue empty notices I can't find.
 			// @todo find the source.
-			if ( empty( $notice['template'] ) && empty( $notice['html'] ) ) {
+			if ( empty( $notice ) || ( empty( $notice['template'] ) && empty( $notice['html'] ) ) ) {
 				self::delete_notice( $notice_id );
+
+				/** @noinspection PhpUnnecessaryStopStatementInspection */
+				return;
 			}
 			?>
 			<div class="notice notice-<?php echo $notice['type']; ?> llms-admin-notice" id="llms-notice<?php echo $notice_id; ?>" style="position:relative;">
--- a/includes/admin/class.llms.admin.notices.php
+++ b/includes/admin/class.llms.admin.notices.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Admin/Classes
  *
  * @since 3.0.0
- * @version 5.2.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
--- a/includes/admin/class.llms.admin.notices.php
+++ b/includes/admin/class.llms.admin.notices.php
@@ -306,7 +306,6 @@ class LLMS_Admin_Notices {
 			if ( empty( $notice ) || ( empty( $notice['template'] ) && empty( $notice['html'] ) ) ) {
 				self::delete_notice( $notice_id );
 
-				/** @noinspection PhpUnnecessaryStopStatementInspection */
 				return;
 			}
 			?>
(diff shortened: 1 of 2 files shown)


tests/phpunit/unit-tests/admin/class-llms-test-admin-notices.php
```diff
@@ -408,6 +408,41 @@ public function test_has_notice() {
...
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function test_output_notice() {
+
+		LLMS_Admin_Notices::init();
+
+		# Create a normal notice.
+		$notice_html = 'Have you heard of the band 999 MB? They haven\'t got a gig yet.';
```
thomasplevy, line 423: :rofl:

includes/admin/class.llms.admin.notices.php
```diff
@@ -300,14 +301,13 @@ public static function output_notice( $notice_id ) {
...
-				return;
-			}
-
 			// Don't output those rogue empty notices I can't find.
 			// @todo find the source.
-			if ( empty( $notice['template'] ) && empty( $notice['html'] ) ) {
+			if ( empty( $notice ) || ( empty( $notice['template'] ) && empty( $notice['html'] ) ) ) {
 				self::delete_notice( $notice_id );
+
+				/** @noinspection PhpUnnecessaryStopStatementInspection */
```
thomasplevy: What's this? an IDE code sniffer ignore or something?

I'd prefer to not start adding IDE-specific attributes into the codebase. My thinking here is that if we start to add IDE-specific attributes for *you* we also need to consider that we might need to start adding IDE specific attributes for other developers. This doesn't *affect* anything greatly but imagine a team with 10 developers all using different IDEs (I know there maybe isn't that many robust / popular IDEs for PHP, but still). I can imagine that over time the codebase could become pretty "ugly"

Is there another way to satisfy whatever is causing your IDE to throw a warning / error / whatever here? Can the code be rewritten as to not throw this kind of warning / error / etc...?

tests/phpunit/unit-tests/admin/class-llms-test-admin-notices.php
```diff
@@ -408,6 +408,41 @@ public function test_has_notice() {
...
+		$notice_id   = 'test-output-notice-normal';
+		LLMS_Admin_Notices::add_notice( $notice_id, $notice_html );
+		LLMS_Admin_Notices::save_notices();
+
+		# Test where current user does not have the 'manage_options' capability.
+		$this->assertOutputEmpty( array( 'LLMS_Admin_Notices', 'output_notice' ), array( $notice_id ) );
+
+		# Test where current user does have the 'manage_options' capability.
+		wp_set_current_user( 1 );
```
thomasplevy: While this currently (and probably always) will work, I prefer to create a new user when testing admin-related content in favor of relying on user existing, being an admin, and etc...

```suggestion
		wp_set_current_user( $this->factory->user->create( array( 'role' => 'administrator' ) ) );
```

includes/admin/class.llms.admin.notices.php
```diff
@@ -290,6 +290,7 @@ function( $notice ) {
 	 * @since 3.0.0
 	 * @since 3.7.4 Unknown.
 	 * @since 5.2.0 Ensure `template_path` and `default_path` are properly passed to `llms_get_template()`.
+	 * @since [version] Delete empty notices and do not display them.
```
thomasplevy, line 293: File version header update too please

thomasplevy (changes requested):

includes/admin/class.llms.admin.notices.php
pondermatic: PhpStorm incorrectly issues a "`return` is unnecessary as the last statement in a method" warning. Apparently, the inspection does check if HTML can be output in the method.

I'm OK taking this inspection suppressor out because it's wrong and only a minor nuisance.

There are probably other IDE comment helpers that I might feel more strongly about, but not this one.

PR #1777, opened 2021-09-10
Update package dependencies

thomasplevy:
## Description

Upgrades the npm package dependencies for monorepo packages `@lifterlms/llms-e2e-test-utils` and `@lifterlms/scripts`

## How has this been tested?

+ Manually
+ Updated / existing e2e tests

## Types of changes

+ Removes unused core repo dependencies:    `all-contributors-cli` and `geonames.js`
+ Replaces usage of deprecated`page.waitFor()` with `page.waitForTimeout()` in all e2e tests and  `@lifterlms/llms-e2e-test-utils` functions. This will make the test util package incompatible with puppeteer versions earlier than 5.3.0. This necessitates a major upgrade of the package. Existing add-ons (or 3rd parties) requiring `waitFor()` will be unaffected unless they upgrade.
+ Upgrades `@wordpress/scripts` to 18.x
    + There's a number of breaking changes here which won't affect us when upgrading to the latest version of the package (this PR fixes the stuff that breaks). Here's the list: https://github.com/WordPress/gutenberg/blob/trunk/packages/scripts/CHANGELOG.md#breaking-changes-1
    + Most notably is webpack 4 -> 5 as a inherited dependency
+ Upgrades `@wordpress/e2e-test-utils` to 5.4.0
+ Removes core puppeteer requirement in favor of allowing `@wordpress/scripts` to define the puppeteer version
+ e2e env vars will be loaded from `.llmsenv` and fallback to `.llmsenv.dist` -- the former should not be commited to the vcs (to allow protecting secret secrets)
+ Removes the e2e test fail screenshot reporter in favor of the screenshot (and HTML) reporter provided via `@wordpress/scripts`. Screenshots are now saved in the `tmp/artifcats` directory.
+ Setup linting for JS packages (via `@wordpress/scripts` and a GitHub Action workflow)

## A first step towards removing reliance on `gulp` and `lifterlms-lib-tasks`

I've converted a single JS file to use the `@lifterlms/scripts` for compiliation. I chose the first script in the JS directory as sorted on my machine - `js/llms-admin-addons.js`. I upgraded this to import dependencies, use `wp.18n` in favor of our `LLMS.l10n` translation function. 

The biggest change here is that given the way `@wordpress/scripts` works, an unminified/human readable JS file (with sourcemaps) is only generated during development (when running `npm start`). Given changes to plugin guidelines we're no longer *required* to ship source JS and it's okay to ship uglified/obfuscated/minified JS only -- so long as the source files are available publicly/for free (they are, here in GitHub). 

Are there issues with excluding unminified sources from an end user / developer perspective?

+ If someone wants to try to debug / manually review code they can't do it on their local machine without cloning from GitHub and building the unminified version manually.
+ @eri-trabiccolo pointed out that as a translator it would be more difficult to translate when the context of the string isn't evident through the string, context data, or translators comment (I'd say they can still go to GitHub though, but it is an extra step).
+ The filename will change: `.min.js` will just become `.js` and the `.min.js` will cease to be included (source maps will not be included either). I don't consider this a backwards incompatible change as the asset ids (as passed to WP core asset registration / enqueue functions) will remain the same. Developers *should not* be directly referencing assets in that way and, if they are, maybe they'll have to be taught a lesson in WP asset usage...?

Advantages:

+ Smaller shipped filesize -- we'll be able to drop probably 25+ files (when not shipping the unminified JS/CSS)
+ Simpler build process: The `@wordpress/scripts` does not expect or easily accommodate building two versions of a source file, it expects to generate an unminified version with sourcemaps when in "development" (watching files via the `start`) OR a minified version with no sourcemaps when shipping for production (via the `build` script). We *can* use the functionality to build two versions but I don't think it's essential to do so unless we think the disadvantages above outweigh the advantages and that we *must* retain the unminified version.

One hiccup to consider is that most of the unminified versions *are* the source JS. EG: they don't use any ESNext syntax, they don't use `imports` etc... We will be converting them to use newer syntax (over time) and using more and more imports (most notably `@wordpress/i18n` and `@wordpress/hooks`. Therefore the unminified versions is actually *still* not technically source file. It's an unminified and transformed version of the minified script. This is a semantic point but it's something to consider. Reading a webpack-compiled file is a bit of a pain. That's why the sourcemaps are useful/important... So if we're going to ship unminified with sourcemaps we actually should also include the (now new) source files using new syntax and imports and etc... Which makes the end shipped product *even larger*.

Thoughts?

## Checklist:


thomasplevy:
@gocodebox/engineering I know this is a huge changeset but it finally all works...

The major changes are outlined in the PR. I don't need a line by line review but I need signoff on the strategy.  More than anything I need feedback on whether or not to include unminified stuff alongside this change.

eri-trabiccolo (dismissed):
I haven't tested this, I trust it works as you said, but if you want I'll give it a try.
I approve the strategy.

PR #1778, opened 2021-09-10
Start to remove reliance on gulp and lifterlms-lib-tasks

thomasplevy:
## Description

This PR begins transitioning off of gulp and lifterlms-lib-tasks in favor of an asset pipeline centered around webpack (via `@lifterlms/scripts` and `@wordpress/scripts`

This PR removes the custom building of the builder file/app within the core plugin.

This has been opened on top of `package-dep-upgrades` (#1777) because it relies on dependency upgrades introduced in that PR

## How has this been tested?

+ Manually

## Types of changes

+ Moves builder source scripts from `assets/js/builder` to `src/builder`
+ Removes gulp task for building the builder script
+ Removes reliance on almond.js
+ Defines and loads builder-related assets via the `LLMS_Assets` class
+ Adds the ability to load data from asset.php files (generated via the `@wordpress/dependency-extraction-webpack-plugin`) into an asset definition array in the `LLMS_Assets` class
+ Adds depedency auto-registration of assets in the `LLMS_Assets` class.

## Checklist:


includes/class-llms-assets.php
```diff
@@ -462,6 +468,34 @@ public function is_inline_enqueued( $handle ) {
 		return in_array( $handle, array_keys( $this->inline ), true );
```
eri-trabiccolo:
```suggestion
		return array_key_exists( $handle, $this->inline );
```

.gitignore
```diff
@@ -18,3 +18,7 @@ node_modules/
 # Non-distributable configs.
 phpunit.xml
 .llmsenv
+
```
eri-trabiccolo, line 21: is this temporary?
I mean, as far as I can see we're not generating those maps anymore (I mean, there's no reference to them in the and mins). Why we don't just delete those map files then?

eri-trabiccolo (changes requested):
Looks good, I haven't tested it yet though.

.gitignore
thomasplevy: When you "watch" scripts via `wp scripts start` it generates source maps that are useful when developing / testing new code. By default the `assets/` dir would be cleaned by `wp scripts` when compiling production assets *but* I've disabled that because we currently have the source and minified versions of many scripts in the assets folder. If we clean the whole directory we lose most of our scripts.

So my thinking was to .gitignore them and exclude them from the zip archives during distributions that way we don't have to worry about remembering to delete them (or writing a new script to delete them during builds)

eri-trabiccolo: makes sense. thanks.

PR #1786, opened 2021-09-17
Ensure "model" mce editor instances are de-initialized before cloning

thomasplevy:
## Description

Fixes https://github.com/gocodebox/lifterlms-private-areas/issues/56

## How has this been tested?

Manually

The affected code isn't widely used, it's only actual appearance is in Private Areas (when creating PA automations). Some of the code (but not the mceEditor stuff) is used by the legacy metabox for course/membership instructors (pre block editor).

I could write e2e tests for this (inside PA) but it's a long walk. I've manually tested this for now and that's the best we're going to get to fix this bug.

## Types of changes
Bug fix

## Checklist:


eri-trabiccolo:
@thomasplevy 
looks like we have the exact same problem with the Access Plans, and unfortunately this PR doesn't fix it...

thomasplevy:
@eri-trabiccolo the (new) issue with access plans may have similar symptoms but it's a different issue so I'm going to address that separately.

(It uses different code)

eri-trabiccolo (approved):

PR #1789, opened 2021-09-18
Fix return type in `preprare_query()` DocBlock.

pondermatic:
## Description
Fix return type in preprare_query() DocBlock.

## How has this been tested?
`composer check-cs-errors`

## Types of changes
Coding documentation improvements.

## Checklist:
- [X] My code follows the LifterLMS Coding & Documentation Standards.

### DIFF
--- a/includes/class-llms-events-query.php
+++ b/includes/class-llms-events-query.php
@@ -132,7 +132,7 @@ class LLMS_Events_Query extends LLMS_Database_Query {
 	}
 
 	/**
-	 * Prepare the SQL for the query
+	 * Prepare the SQL for the query.
 	 *
 	 * @since 3.36.0
 	 * @since 4.7.0 Use `$this->sql_select_columns({columns})` to determine the columns to select.
--- a/includes/class-llms-events-query.php
+++ b/includes/class-llms-events-query.php
@@ -136,6 +136,7 @@ class LLMS_Events_Query extends LLMS_Database_Query {
 	 *
 	 * @since 3.36.0
 	 * @since 4.7.0 Use `$this->sql_select_columns({columns})` to determine the columns to select.
+	 *
 	 * @return string
 	 */
 	protected function preprare_query() {
--- a/includes/class-llms-events-query.php
+++ b/includes/class-llms-events-query.php
@@ -136,7 +136,6 @@ class LLMS_Events_Query extends LLMS_Database_Query {
 	 *
 	 * @since 3.36.0
 	 * @since 4.7.0 Use `$this->sql_select_columns({columns})` to determine the columns to select.
-	 *
 	 * @return string
 	 */
 	protected function preprare_query() {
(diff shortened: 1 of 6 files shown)


includes/class-llms-events-query.php
```diff
@@ -132,11 +132,10 @@ protected function parse_args() {
 	}
 
 	/**
-	 * Prepare the SQL for the query
+	 * Prepare the SQL for the query.
 	 *
 	 * @since 3.36.0
 	 * @since 4.7.0 Use `$this->sql_select_columns({columns})` to determine the columns to select.
-	 *
```
eri-trabiccolo: This should stay, we probably don't have a cs rule for this but I think we should.
We always left on "blank" line before the `@param`, or before `@return` when there are no `@param`.
@thomasplevy what do you think?

includes/class.llms.query.quiz.attempt.php
```diff
@@ -133,10 +133,9 @@ protected function parse_args() {
 	}
 
 	/**
-	 * Prepare the SQL for the query
+	 * Prepare the SQL for the query.
 	 *
 	 * @since 3.16.0
-	 *
```
eri-trabiccolo, line 139: This should stay

includes/class.llms.query.user.postmeta.php
```diff
@@ -196,11 +196,10 @@ protected function parse_args() {
...
 
 	/**
-	 * Prepare the SQL for the query
+	 * Prepare the SQL for the query.
 	 *
-	 * @return   void
-	 * @since    3.15.0
-	 * @version  3.15.0
+	 * @since 3.15.0
+	 * @return string
```
eri-trabiccolo:
```suggestion
	 * @since 3.15.0
	 *
	 * @return string
```

includes/class.llms.student.query.php
```diff
@@ -150,13 +150,12 @@ protected function parse_args() {
 	}
 
 	/**
-	 * Prepare the SQL for the query
+	 * Prepare the SQL for the query.
 	 *
 	 * @since 3.4.0
 	 * @since 3.13.0 Unknown.
 	 * @since 4.10.2 Demands to `$this->sql_select()` to determine whether or not `SQL_CALC_FOUND_ROWS` statement is needed.
-	 *
```
eri-trabiccolo, line 158: This should stay

eri-trabiccolo (changes requested):

includes/class-llms-events-query.php
thomasplevy: I haven't taken the time to write a PHPCS rule for this but it is in our doc standards ( and I think the WP core's too )

thomasplevy: At least it *should* be in our doc standards. If it's not it should be since 99% of our code adheres to it (intentionally)

eri-trabiccolo (approved):

PR #1796, opened 2021-09-20
Require a blank line before a @return line.

pondermatic:
## Description
Documented the requirement of a blank line before a `@return` line in a DocBlock.

## Types of changes
Documentation standards.

## Checklist:

### DIFF
--- a/docs/documentation-standards.md
+++ b/docs/documentation-standards.md
@@ -195,14 +195,17 @@ When updating an existing element:
 #### 1. Parameters and Returns
 
 Functions and methods should define all parameter arguments and returns with the `@param` and `@return` tags.
+There should be a blank line before the first `@param` line and before the `@return` line.
 
 No HTML should be used in the descriptions for these tags, though limited Markdown can be used as necessary, such as for adding backticks around variables, e.g. `$variable`.
 
 All descriptions for any of these tags should be a full sentence ending with a full stop (a period, for example).
 
 ```
+ *
  * @param string $var1 Description of the argument.
  * @param bool $var2 Description of the argument.
+ *
  * @return string
  */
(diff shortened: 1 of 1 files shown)


eri-trabiccolo:
I'd add "...when not preceded by a `@param` line", or something similar, what do you think?

thomasplevy:
This is *not* the correct standard that we follow though and I frankly don't think this is a valuable change to make.

`@param` and `@return` *should* be grouped together

thomasplevy:
This isn't explicitly stated in the WP core standards but it is *followed* by the WP core (and for 99% of our code the existing standard is how we do it)

What's the motivation for this change?

pondermatic:
@thomasplevy wrote
> What's the motivation for this change?

In the code review for #1789, @eri-trabiccolo said that the blank line before the `@return` line should stay.

thomasplevy:
@pondermatic I think you misread:

https://github.com/gocodebox/lifterlms/pull/1789#discussion_r711941240

> We always left on "blank" line before the @param, or before @return when there are no @param.

There should be a blank line *following* the `@since` tag (before either the params list of the return tag). That is sometimes going to be a blank line *before* the `@return` tag but that doesn't mean *always* a blank line before the `@return` tag

The existing examples show a blank line between `@since` and no blank line between `@param` and `@return`

The existing docs *do not* explicitly state that there *should* be a blank line after `@since` tags but it's implied by the example.

This PR would introduce a change that affects 99% of our codebase and also differs from the WP coding standards.

thomasplevy:
I would opt to *not* introduce any changes to the spec and that your previous PR should just follow the existing standard

pondermatic:
Misunderstanding.

PR #1800, opened 2021-09-21
Async dependency waiter improvements

### LINKED ISSUE #1797: LLMS.wait_for should give more information on the dependency
Currently (LifterLMS 5.3.1) this util only prints to messages:
`waiting...` while waiting for a condition to be met
`could not load dependency...` when the condition is still not met after 30 seconds.

When this happens we have no information on what's missing.

The idea would be to add a third (optional) param (`dep`) to `LLMS.wait_for` which is a string representing the condition to be met.

what do you think @gocodebox/engineering

thomasplevy:
## Description

+ Closes #1797
+ Fixes https://github.com/gocodebox/private-issues/issues/43

Adds a 3rd parameter to the `LLMS.wait_for()` function. Uses "unnamed" if not passed.

Adds names to functions in the `LLMS` namespace that use this. I did not endeavor to track down every instance of this throughout the codebase but I found that the offending dependency identified in https://github.com/gocodebox/private-issues/issues/43 was related to the loading of `llmsSelect2` in the forms namespace. So I added it there.

Also fixed the issue causing this to happen on various pages as a result of the check for llmsSelect2 running on page load (everywhere) instead of only running when it's actually going to be used (for forms)

## How has this been tested?

+ Manually 
+ existing e2e tests for form related stuff

## Types of changes

Bug fix and debugging improvement

## Checklist:


eri-trabiccolo (approved):

pondermatic (approved):

thomasplevy:
WP core is in the midst of backporting phpunit polyfil stuff into the testing lib and it's causing tests to intermittently fail. I can't resolve it because the backport isn't complete (I think). Merging this despite the failing tests because they're not related and working okay local.

PR #1802, opened 2021-09-21
Release Version 5.3.3

thomasplevy:
+ Closes #1784 via #1801
+ Closes #1797 via #1800
+ Fixes #1803 via #1804
+ Fixes #1814 via #1815, thanks @chetansatasiya!
+ #1806
+ Fixes https://github.com/gocodebox/private-issues/issues/43 via #1800

PR #1805, opened 2021-09-24
Update tests in accordance with changes in the WP core (yoast/phpunit-polyfill)

thomasplevy:
Upgrades phpunit tests to utilize the yoast/phpunit-polyfill as required by the WP core tests lib.

This is generally available now and backported to previous versions of WP. This is why all our tests suites are failing everywhere now.

lifterlms/lifterlms-tests changes are staged for testing and review purposes before pushing them globally.

~I'm doing more testing here before ready for a true review / ship~

~There's a gotcha that this *doesn't* work on the latest WP tests, so running:~

~`WP_VERSION=5.8 composer run tests-install` will fix that (for now)~

~Using `WP_VERSION=latest` results in 5.8.1 being pulled and the backport was only onto the branches, not the tags. It's unclear to me whether or not a 5.8.2 (tag) will be released with these changes.~

~We might need to change `latest` to pull the X.X branch instead of the X.X.X tag. I'm pretty sure the X.X branch is always equal to the X.X.X tag. I need to verify this.~

### Removing the `composer.lock` file as per the precedent put forth by the WP Core, see: https://github.com/WordPress/wordpress-develop/commit/0e442c4615bdcdc1c2e4f8db6f88f57e6859c2ff

This might be controversial but we can pretty easily follow the logic here.

Non-dev deps (lifterlms-blocks, etc...) are controlled at a specific version number in `composer.json` so the lockfile isn't needed for these dependencies.

Dev dependencies are (currently) limited, and, in general if the yoast polyfill library does it's job, we shouldn't have to worry about the lack of a lock files breaking our tests. The CS lib is a little bit different but WPCS itself doesn't update that frequently so I think we're okay to follow them as they go and risk running into issues. We could create a tighter upstream lock in `lifterlms-tests` if we run into issues of frequent upgrades causing us headaches.

The benefit is that we can start using PHPUnit 9.x immediately. Version 7.x is still necessary to run tests against WP 5.4 (I've handled this condition in the test-phpunit.yml action workflow). By using PHPUnit 9.x immediately we can start running tests against WP 5.9 and PHP 8.1 (this is important to help us get advance signals for forthcoming changes so we're ahead of them before GA releases -- see the latest workflow run on PHP 8.1. There's quite a few deprecations we'll need to tackle which we NOW KNOW ABOUT!)

### I've updated the `composer run tests` script to pass `trunk` as the default value for `WP_TESTS_VERSION`. 

As of this moment the changes necessary to run phpunit with the WP core tests library are on trunk. The *branches* (not tags) for 5.8 etc have a different set of changes backported to them which would require us to stick with PHPUnit 7.5.

The downside to this change is that if new (breaking) changes are introduced into the core we're going to get them immediately and our tests will start failing (again). Sticking on 7.5 loses the above noted gains. My hope is that this is a *temporary* change that we can switch off after 5.9 is released (I hope).

If this proves to be untrue the alternate method I'm currently considering is creating our own repo that forks the content off trunk and including the content that way so we can "lock" at a specific version and upgrade as desired.

### GitHub Actions vs Travis

I've been wanting to switch to GH Actions for a bit. Testing all of this proves *much faster and easier* on GH and this introduces a new GH Action workflow for unit testsing and removes the main travis config used to trigger core builds on Travis. In a separate PR I'm going to pull in e2e tests on GH Actions and write a new workflow for running PHPCS.


eri-trabiccolo (dismissed):
Tests pass, with the suite installed via:
`WP_VERSION=5.8 composer run tests-install`
Tested with:
Runtime:       PHP 8.0.7
Runtime:       PHP 7.4.20

thomasplevy:
Alright... This is now really ready (I think). I've updated the description. Can I get some eyes on this @gocodebox/engineering and let me know if you find any issues / problems / etc...

eri-trabiccolo (dismissed):
Looks good to me.

pondermatic:
@thomasplevy
Other than the delta issues listed below, my tests ran successfully.
PHP 7.4.22 + WP trunk + PHPUnit 9.5.10 = success
PHP 8.0.6   + WP trunk + PHPUnit 9.5.10 = success

- [x] https://github.com/gocodebox/lifterlms/blob/dev/tests/phpunit/unit-tests/controllers/class-llms-test-controller-orders.php#L165
- [x] https://github.com/gocodebox/lifterlms/blob/dev/tests/phpunit/unit-tests/processors/class-llms-test-processor-course-data.php#L107
- [x] https://github.com/gocodebox/lifterlms/blob/dev/tests/phpunit/unit-tests/processors/class-llms-test-processor-course-data.php#355
- [x] https://github.com/gocodebox/lifterlms/blob/dev/tests/phpunit/unit-tests/processors/class-llms-test-processor-course-data.php#362
- [x] https://github.com/gocodebox/lifterlms/blob/dev/tests/phpunit/unit-tests/processors/class-llms-test-processor-course-data.php#389
- [x] https://github.com/gocodebox/lifterlms/blob/dev/tests/phpunit/unit-tests/processors/class-llms-test-processor-course-data.php#396
- [x] https://github.com/gocodebox/lifterlms/blob/dev/tests/phpunit/unit-tests/processors/class-llms-test-processor-course-data.php#422
- [x] https://github.com/gocodebox/lifterlms/blob/dev/tests/phpunit/unit-tests/processors/class-llms-test-processor-course-data.php#448
- [x] https://github.com/gocodebox/lifterlms/blob/dev/tests/phpunit/unit-tests/processors/class-llms-test-processor-course-data.php#481
- [x] https://github.com/gocodebox/lifterlms/blob/dev/tests/phpunit/unit-tests/processors/class-llms-test-processor-course-data.php#563
- [x] https://github.com/gocodebox/lifterlms/blob/dev/tests/phpunit/unit-tests/user/class-llms-test-student.php#L454
- [x] https://github.com/gocodebox/lifterlms/blob/dev/tests/phpunit/unit-tests/user/class-llms-test-student.php#L456

thomasplevy:
@pondermatic thanks for catching these, updated in https://github.com/gocodebox/lifterlms/pull/1802/commits/1d517576fc8b975b763976c0303dee1846522191

pondermatic:
Thanks to PhpStorm's [Code Inspection](https://www.jetbrains.com/help/phpstorm/2021.2/code-inspection.html) feature for helping find the remaining calls that used the deprecated argument.
image

PR #1807, opened 2021-09-28
Move PHPCS from Travis to GitHub Actions

thomasplevy:
Adds a GitHub actions for running PHPCS on pull requests

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -1,0 +1,37 @@
+name: Coding Standards
+
+on:
+  workflow_dispatch:
+  pull_request:
+
+concurrency:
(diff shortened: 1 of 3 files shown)


eri-trabiccolo (approved):
I guess this is only the gh actions creation,
misses the travis removal, right?

thomasplevy:
The travis removal is *complicated* because the Travis tests and CS all run on the same config (you can't do like multiple travis jobs from a single repo like you can with GH workflows).

So the "removal" of travis entirely is already started in #1805. I wanted to keep the addition of the CS job separate because it's easy enough to have separate and self-contained


Migration strategy is to get things working on GH actions and have both GH and travis running together until we're completely switched over.

PR #1808, opened 2021-09-29
Fix various issues when adding existing lessons/quizzes/assignments

eri-trabiccolo:
## Description

Fixes https://github.com/gocodebox/private-issues/issues/41

## How has this been tested?
manually

## Types of changes

## Checklist:

### DIFF
--- a/assets/js/builder/Models/Course.js
+++ b/assets/js/builder/Models/Course.js
@@ -67,7 +67,7 @@ define( [ 'Collections/Sections', 'Models/_Relationships', 'Models/_Utilities' ]
 		 * @since 3.16.0
 		 * @since 3.24.0 Unknown.
 		 * @since 3.37.11 Use the author id instead of the author object.
-		 * @since [version] On cloning, duplicate assignments too, if assignment add-on active and assignment attached.
+		 * @since [version] Added filter hook 'llms_cloning_lesson_data'.
 		 *
 		 * @param {Object} lesson Lesson data obj.
 		 * @return {Void}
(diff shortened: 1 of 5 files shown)


thomasplevy:
I know that I put forth the precedent of having assignments code in the core codebase but it's kind of a terrible precedent. I don't have a good rule for when code should or should not be in the assignments add-on (obviously) but this adds more logic for handling assignments into the core and further complicates the preexisting issue.

Did you consider any of this or try to add this code to the assignments addon using JS hooks? This might not be that hard to do.

Thoughts?

eri-trabiccolo:
> I know that I put forth the precedent of having assignments code in the core codebase but it's kind of a terrible precedent. I don't have a good rule for when code should or should not be in the assignments add-on (obviously) but this adds more logic for handling assignments into the core and further complicates the preexisting issue.
> 
> Did you consider any of this or try to add this code to the assignments addon using JS hooks? This might not be that hard to do.
> 
> Thoughts?

I could do that, but the whole Assignment view is in core, so ideally I had to do a small change in the view in core and a small (though bigger than the one in the view) change in the assignment add-on. I can.
Fine for you to keep the Assignment view's changes in core?

eri-trabiccolo:
@thomasplevy see https://github.com/gocodebox/lifterlms/pull/1808/commits/dde8ad5a9c33c1173e3796e783487702c74f16ba
and
https://github.com/gocodebox/lifterlms-assignments/pull/68
does this make more sense?
It seems a little bit clumsy to me tbh having the assignments view in core...

eri-trabiccolo:
So we decided to revert the moving assignment cloning logic in the assignment add-on, as most of the involved entities are already in core.

Issue #627, opened 2018-08-13
BuddyPress profile doesn't have a LifterLMS "Notifications" tab


thomasplevy:
![_62167_lifter_lms_notifications_to_be_listed_in_buddypress_area_-_science_cosmos](https://user-images.githubusercontent.com/1290739/44063452-5b7047e4-9f15-11e8-942a-7f2062afe4fc.jpg)

yojance:
Submitted a PR for this #693

PR #1816, opened 2021-10-04
BuddyPress integration: Display all registered dashboard tabs (enabled in the settings) automatically.

eri-trabiccolo:
## Description

- Fixes https://github.com/gocodebox/private-issues/issues/42
- Fixes https://github.com/gocodebox/lifterlms/issues/627

## How has this been tested?
manually and with new unit tests

## Types of changes

## Checklist:


includes/integrations/class.llms.integration.buddypress.php
```diff
@@ -1,36 +1,60 @@
...
 	 */
 	protected $priority = 5;
 
 	/**
-	 * Configure the integration
+	 * Current endpoint's key being processed.
+	 *
+	 * @var string
+	 */
+	private $current_endpoint_key;
```
pondermatic, line 35: I usually prefer to make properties `protected` so that a class that extends this class can change it. Another option is to add getter and setter methods.

includes/integrations/class.llms.integration.buddypress.php
```diff
@@ -54,79 +78,175 @@ protected function configure() {
...
+		$this->endpoints = apply_filters( 'llms_buddypress_profile_endpoints', $endpoints );
+
+	}
+
+	/**
+	 * Get a list of custom endpoints to add to BuddyPress profile page.
+	 *
+	 * @since [version]
+	 *
+	 * @param bool $active_only If true, returns only active endpoints.
```
pondermatic: Should it be noted that the parameter is optional?

eri-trabiccolo: Not sure...
We recently "discovered" that the generated html docs  automatically adds "Optional" and the default value. So we thought we could avoid the manual addition, furthermore the manual addition might create duplicates. We wanted to check actually what happens... Do you have the time to make a rapid check with the docs generator?

includes/integrations/class.llms.integration.buddypress.php
eri-trabiccolo: I recently started to make props and methods `private` as much as possible, mostly because if we want to turn a property/method to protected/public because a dev requested it, or because we need it to be so, we can do it without introduce bc (backward compatibility) issues, while the opposite, of course, it will introduce a bc issue.
I'd say `protected` would be fine as well, but given the logic here I doubt either making it `protected` or introducing `protected` getters/setters would be of any help for an extending class.
Use case?

includes/integrations/class.llms.integration.buddypress.php
pondermatic: Yes, the generated HTML docs do have "(Optional)" and "Default value:" added, so we don't need to indicate this in the description.

Good example.
https://developer.lifterlms.com/reference/classes/llms_database_query/get/

Redundant example.
https://developer.lifterlms.com/reference/classes/llms_date/db_date/

eri-trabiccolo: So I'd say we don't add that stuff anymore. At most, if possible, the docs parser should remove `Optional.` and `Default value [...] .` from the description, if we want to keep them in the codebase. But I don't think it's worthwhile. Let's stop using them and progressively remove them from the codebase - if @thomasplevy agrees.

includes/integrations/class.llms.integration.buddypress.php
pondermatic: It doesn't seem likely that a 3rd party would extend this class, so the visibility on these two properties probably don't matter.

However, I get frustrated a few times a year when the project I'm working on needs access to a variable that is private and does not have a getter/setter, or I am extending a class and can't override a private method.

I see you point about backward compatibility though. I need to read up on the pros and cons of private/protected to help balance future decisions.

https://stackoverflow.com/questions/4913025/reasons-to-use-private-instead-of-protected-for-fields-and-methods#answer-14201260

includes/integrations/class.llms.integration.buddypress.php
thomasplevy: I agree, no need to write things that are going to be automatically output.

includes/integrations/class.llms.integration.buddypress.php
thomasplevy: I go back and forth on this... however... as a maintainer of a codebase I more frequently bang my head against a wall trying to figure out *how to deprecate something gracefully* than someone complains about something being "private"

I have seen the argument about *never* using private and making everything protected all the time but backwards compat is such a pain and it's so nice when we can simply delete a method we don't need anymore.

eri-trabiccolo:
@thomasplevy 
I haven't written any real test that would require BuddyPress to be installed... 
Do you think we should download and activate bp and write specific tests?
I'd say we should, but I don't know whether this would be a priority or not.

includes/functions/llms.functions.page.php
```diff
@@ -77,6 +79,7 @@ function llms_get_endpoint_url( $endpoint, $value = '', $permalink = '' ) {
 
 	if ( ! $permalink ) {
 		$permalink = get_permalink();
+		$permalink = empty( $permalink ) ? wp_guess_url() : $permalink;
```
thomasplevy: This is a weird function that I've yet to encounter. Cool?

includes/functions/llms.functions.templates.dashboard.php
```diff
@@ -506,7 +507,7 @@ function lifterlms_template_student_dashboard_my_grades() {
 		}
 
 		global $wp_query, $wp_rewrite;
-		$slug = $wp_query->query['my-grades'];
+		$slug = $wp_query->query['my-grades'] ?? '';
```
thomasplevy, line 529: I see this, thank you.

thomasplevy (changes requested):
I'm sorry @eri-trabiccolo this is going to constitute a new feature and a minor bump to 5.5.0 :-(


I'd like to see tests added for `llms_get_endpoint_url()` I know we don't have any coverage on that test today so this is a good reason to cover the whole function.

While manually testing it I found a few issues and have a few thoughts:

**My Grades** Tab

+ Sorting the table doesn't work
+ Clicking into a course loads the course itself (instead of the drilldown into the course details like on the default student dashboard)

These are probably problems with how permalinks are generated in the core template not so much in the BP functionality but it should be addressed.

**Edit account & orders page both 404**

includes/functions/llms.functions.page.php
eri-trabiccolo: Yeah, I just discovered it.
Consider that in our user cases the `get_permalink()` was never empty, except in the bp profile.

thomasplevy:
> @thomasplevy I haven't written any real test that would require BuddyPress to be installed... Do you think we should download and activate bp and write specific tests? I'd say we should, but I don't know whether this would be a priority or not.

Some of the functions in the integration class don't require BP at all and can be tested without BP existing. 

Some of the used BP methods can be mocked (see how we accomplish this here: https://github.com/gocodebox/lifterlms/blob/trunk/tests/phpunit/unit-tests/integrations/class-llms-test-integration-bbpress.php) 

With a little refactoring and some mocks the class can probably be fully covered without needing to physically install BP to run the tests

eri-trabiccolo:
> I'm sorry @eri-trabiccolo this is going to constitute a new feature and a minor bump to 5.5.0 :-(

Fine.

> I'd like to see tests added for llms_get_endpoint_url() I know we don't have any coverage on that test today so this is a good reason to cover the whole function.

Ok.

> While manually testing it I found a few issues and have a few thoughts:
> 
> My Grades Tab
> 
> Sorting the table doesn't work

Tabs are not sortable, yeah:

> Clicking into a course loads the course itself (instead of the drilldown into the course details like on the default student dashboard)
> These are probably problems with how permalinks are generated in the core template not so much in the BP functionality but it should be addressed.

Going to check this, considering we use the same student dashboard template there must be something there that acts differently because we're not in a dashboard endpoint.

> Edit account & orders page both 404

I think I bumped into this, do you have WooCommerce enabled? It's like wc "steals" edit-account and "orders" slugs, so instead of reinventing the wheel (meaning, still demanding all the routing to buddypress), I'd suggest that WooCommerce user + BuddyPress users just change those acocunt endpoint slugs to avoid conflicts. What do you think?
We can write a documentation about that.

thomasplevy:
@eri-trabiccolo ah, yes, re the 404ing tabs: I did have WC enabled, disabling it fixes it. Yes, we can simply add this to documentation on BP knowledgebase as a "known" conflict when using LLMS + BP + WC... Your workaround seems *totally fine* to me... Especially since if using WC and LLMS you probably are selling through WC so you don't even need our orders endpoint. It's arguable whether you'd need account edit, but we can work around it as you've suggested. Cool.

eri-trabiccolo:
@thomasplevy 

> While manually testing it I found a few issues and have a few thoughts:
> 
> My Grades Tab
> 
> Sorting the table doesn't work
> 
> Tabs are not sortable, yeah:

I misread this, as you can see I replied with "tab are not sortable" while you were talking about the grades table :D... actually the table sorting does work here, I had tested it, the `wp_guess_url()` was basically introduced exactly for fixing that issue mmm. Can you double check?

> Clicking into a course loads the course itself (instead of the drilldown into the course details like on the default student dashboard)

I can't reproduce this either... once again the `wp_guess_url()` it's what should make this work...

thomasplevy:
https://user-images.githubusercontent.com/1290739/138729554-7422c300-9a07-46f5-8d0d-3657ab27a887.mp4

eri-trabiccolo:
@thomasplevy 
thanks.
Cannot reproduce :(
https://www.loom.com/share/c72e9ba002be45d1bb4505828672c258
and:
Schermata 2021-10-25 alle 18 49 00

Can you test - when you got time - with the latest?

Any chance I can test your site?

eri-trabiccolo:
@thomasplevy 
The reason why it doesn't work for you it's probably because of [`wp_guess_url()`](https://core.trac.wordpress.org/browser/tags/5.8/src/wp-includes/functions.php#L5719) not working as I thought. Inspecting the function it's clear that it doesn't do what I wanted when `( defined( 'WP_SITEURL' ) && '' !== WP_SITEURL )`. Sorry for having wasted your time.
fixed in https://github.com/gocodebox/lifterlms/pull/1816/commits/bbb8a467fdb25ed15c0d9479375adf793bcced6e

includes/functions/llms.functions.page.php
eri-trabiccolo: That's a function turned out to be totally useless. <3

thomasplevy:
I'm having a hard time reviewing your latest changes @eri-trabiccolo -- the updated `llms_get_endpoint_url()` function has gone from being *kind of dense and complicated* to now extremely complicated... I don't quite understand the problem that's being solved (as I haven't stepped through all the code you've been working on) so reading this to understand why these changes have been made is difficult. I'd feel a TON more confident if there were some tests illustrating the various scenarios now handled by the method.

eri-trabiccolo:
> I'm having a hard time reviewing your latest changes @eri-trabiccolo -- the updated `llms_get_endpoint_url()` function has gone from being _kind of dense and complicated_ to now extremely complicated... I don't quite understand the problem that's being solved (as I haven't stepped through all the code you've been working on) so reading this to understand why these changes have been made is difficult. I'd feel a TON more confident if there were some tests illustrating the various scenarios now handled by the method.

I do understand what you mean @thomasplevy,
I will add some comments too. To make it simple, though:
When you're in our dashboard `get_permalink()` works fine and returns the permalink of the dashboard page (`//example.com/dashboard/`), to which the current (stable) code adds the endpoint slug (`my-courses`), a slash, and possible query arguments (`//example.com/dashboard/my-courses/`).
BUT, when you are in the BuddyPress profile tabs, `get_permalink()` doesn't work, so we need to retrieve the endpoint url from something else, and we're going to do that with `filter_var( wp_unslash( $_SERVER['REQUEST_URI'] ), FILTER_SANITIZE_URL )` which returns the requested uri, which in the grades tab it would be something like:
`//example.com/members/admin/courses/my-grades/page/2/`, to which we would add the endpoint slug (and possible query vars): `//example.com/members/admin/courses/my-grades/page/2/my-grades/`.
So what the dense code does is removing:
a) pagination part, if any
b) endpoint slug part, if any
so that we get  `//example.com/members/admin/courses` to which we add the endpoint and the query args so to have:
`//example.com/members/admin/courses/my-grades/`

thomasplevy:
@eri-trabiccolo I don't know what was left here but what do you think about bumping this for scheduling after 6.0 release?

eri-trabiccolo:
Yes, I agree.
I just rebased it a couple of days ago and planned this to be shipped as 6.1.0, see https://github.com/gocodebox/lifterlms/pull/1816/files#diff-12a7caf8a379d32f2ee26b5fd80db84a84e1b78538f6f62c05fc1c1129b20d56R11

I'm not working on this right now. :D

thomasplevy:
@eri-trabiccolo perfect <3

thomasplevy:
@eri-trabiccolo moving this out of in progress until you get back to actively working on it

thomasplevy:
@eri-trabiccolo can you switch this to a draft until it's ready for review again?

eri-trabiccolo:
Nobody seems to love this!

PR #1838, opened 2021-10-27
Allow editing earned certificates and achievements

eri-trabiccolo:
## Description
Allow editing earned certificates and achievements.

~Pondering on whether to extend this to achievements too: they are not `public` and they don't have a template, so to actual see the changes one should be the one who earned it or using an user switcher plugin...~

A couple of thoughts about the "my" achievements:

a) They don't have a public permalink, and I don't think we should introduce it just for the purpose of allowing seeing the published changes in an easy way: even if I guess it would be rare, we could generate some conflict with 3rd party code.
b) They could be previewed (Preview Changes button) BUT since most of the actual, and relevant, achievement data to be displayed are post meta (achievement_title, achievement_content, achievement_image), and as we know not already saved post metas are not available in the preview, I decided to leave the preview not available
c) They don't have a "single" template, which because of a) and b) I decided to not implement - although its implementation would be trivial.

If you think otherwise let's see what we can do. But I think that this PR can be merged as it is, and possible further improvements could be set to be done in the future.

## How has this been tested?
manually

## Types of changes

## Checklist:

### DIFF
--- a/includes/admin/post-types/meta-boxes/fields/llms.class.meta.box.fields.php
+++ b/includes/admin/post-types/meta-boxes/fields/llms.class.meta.box.fields.php
@@ -11,7 +11,7 @@
 defined( 'ABSPATH' ) || exit;
 
 /**
- * Metabox_Field parent class
+ * Metabox_Field parent class.
  *
  * Contains base code for each of the Metabox Fields.
  *
@@ -21,14 +21,14 @@ defined( 'ABSPATH' ) || exit;
 abstract class LLMS_Metabox_Field {
 
 	/**
-	 * Global array used in class instance to store field information
+	 * Global array used in class instance to store field information.
 	 *
 	 * @var array
 	 */
 	public $field;
 
 	/**
-	 * Global variable to contain meta information about $field
+	 * Global variable to contain meta information about $field.
 	 *
 	 * @var object
 	 */
@@ -37,10 +37,13 @@ abstract class LLMS_Metabox_Field {
 	/**
 	 * Outputs the head for each of the field types.
 	 *
-	 * @todo  all the unset variables here should be defaulted somewhere else probably
+	 * @todo All the unset variables here should be defaulted somewhere else probably.
+	 *
 	 * @since unknown
 	 * @since 3.11.0 Unknown.
 	 * @since [version] Do not print empty labels; do not print the description block if both 'desc' and 'label' are empty.
+	 *
+	 * @return void
 	 */
 	public function output() {
 
@@ -77,19 +80,24 @@ abstract class LLMS_Metabox_Field {
 			<div class="description <?php echo $this->field['desc_class']; ?>">
 			<?php if ( ! empty( $this->field['label'] ) ) : ?>
 				<label for="<?php echo $this->field['id']; ?>"><?php echo $this->field['label']; ?></label>
-			<?php endif ?>
+			<?php endif; ?>
 				<?php echo $this->field['desc']; ?>
 				<?php
 				if ( isset( $this->field['required'] ) && $this->field['required'] ) :
 					?>
 					<em>(required)</em><?php endif; ?>
 			</div>
-		<?php endif; ?>
-		<?php
+			<?php
+			endif;
+
 	}
 
 	/**
-	 * outputs the tail for each of the field types
+	 * Outputs the tail for each of the field types
+	 *
+	 * @since unknown.
+	 *
+	 * @return void
 	 */
 	public function close_output() {
 
@@ -98,13 +106,14 @@ abstract class LLMS_Metabox_Field {
 	}
 
 	/**
-	 * Set the default meta value of a field
+	 * Set the default meta value of a field.
 	 *
-	 * @param    int    $post_id   WP Post ID
-	 * @param    string $field_id  ID/name of the field
-	 * @return   mixed
-	 * @since    1.0.0
-	 * @version  3.24.0
+	 * @since 1.0.0
+	 * @since 3.24.0 Unknown.
+	 *
+	 * @param int    $post_id  WP Post ID.
+	 * @param string $field_id ID/name of the field.
+	 * @return mixed
 	 */
 	public static function get_post_meta( $post_id, $field_id ) {
 
--- a/includes/admin/post-types/meta-boxes/fields/llms.class.meta.box.fields.php
+++ b/includes/admin/post-types/meta-boxes/fields/llms.class.meta.box.fields.php
@@ -1,11 +1,11 @@
 <?php
 /**
- * Abstract Metabox_Field
+ * Abstract Metabox_Field.
  *
  * @package LifterLMS/Admin/PostTypes/MetaBoxes/Fields/Classes
  *
- * @since ??
- * @version 3.24.0
+ * @since unknown
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -15,7 +15,7 @@ defined( 'ABSPATH' ) || exit;
  *
  * Contains base code for each of the Metabox Fields.
  *
- * @since ??
+ * @since unknown
  * @since 3.24.0 Unknown.
  */
 abstract class LLMS_Metabox_Field {
@@ -35,11 +35,12 @@ abstract class LLMS_Metabox_Field {
 	public $meta;
 
 	/**
-	 * outputs the head for each of the field types
+	 * Outputs the head for each of the field types.
 	 *
 	 * @todo  all the unset variables here should be defaulted somewhere else probably
-	 * @since    ??
-	 * @version  3.11.0
+	 * @since unknown
+	 * @since 3.11.0 Unknown.
+	 * @since [version] Do not print empty labels; do not print the description block if both 'desc' and 'label' are empty.
 	 */
 	public function output() {
 
@@ -72,15 +73,19 @@ abstract class LLMS_Metabox_Field {
 
 		?>
 		<li class="<?php echo implode( ' ', $wrapper_classes ); ?>"<?php echo $controller . $controller_value; ?>>
+		<?php if ( ! empty( $this->field['desc'] ) || ! empty( $this->field['label'] ) ) : ?>
 			<div class="description <?php echo $this->field['desc_class']; ?>">
+			<?php if ( ! empty( $this->field['label'] ) ) : ?>
 				<label for="<?php echo $this->field['id']; ?>"><?php echo $this->field['label']; ?></label>
+			<?php endif ?>
 				<?php echo $this->field['desc']; ?>
 				<?php
 				if ( isset( $this->field['required'] ) && $this->field['required'] ) :
 					?>
 					<em>(required)</em><?php endif; ?>
 			</div>
-			<?php
+		<?php endif; ?>
+		<?php
 	}
 
 	/**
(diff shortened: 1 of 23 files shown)


eri-trabiccolo:
@gocodebox/engineering 
your early thoughts are greatly appreciated!

thomasplevy:
@eri-trabiccolo I didn't test it but it looks good so conceptually

eri-trabiccolo:
doc fixes here:
https://github.com/eri-trabiccolo/lifterlms/tree/edit-certificates-docs
(going to open a PR once this is merged...)

includes/class.llms.post-types.php
```diff
@@ -752,6 +753,50 @@ public static function register_post_types() {
...
+				'public'              => false,
+				/**
+				 * Filters the needed capability to generate and allow a UI for managing `llms_my_achievement` post type in the admin.
+				 *
+				 * @since [version]
+				 *
+				 * @param bool $show_ui The needed capability to generate and allow a UI for managing `llms_my_achievement` post type in the admin.
+				 *                      Default is `manage_lifterlms`.
+				 */
+				'show_ui'             => ( current_user_can( apply_filters( 'lifterlms_admin_my_achievements_access', 'manage_lifterlms' ) ) ) ? true : false,
```
thomasplevy: I already chatted with you about this but I'm thinking we should use a new capability here:

`manage_earned_engagements`

This will not need to be filtered and will be a cap added to admins and lms managers by default

In the future we could add this to instructors conditionally by filtering `user_has_cap` to check if they can manage a specific earned cert based on the related post id of the certificate

includes/class.llms.post-types.php
```diff
@@ -818,7 +863,15 @@ public static function register_post_types() {
...
-					'show_ui'             => true,
+					/**
+					 * Filters the needed capability to generate and allow a UI for managing `llms_my_certificate` post type in the admin.
+					 *
+					 * @since [version]
+					 *
+					 * @param bool $show_ui The needed capability to generate and allow a UI for managing `llms_my_certificate` post type in the admin.
+					 *                      Default is `manage_lifterlms`.
+					 */
+					'show_ui'             => ( current_user_can( apply_filters( 'lifterlms_admin_my_certificates_access', 'manage_lifterlms' ) ) ) ? true : false,
```
thomasplevy: Same as my note with achievements

I think sharing the capability between the two makes sense.

thomasplevy (commented):
I need to review the changes to uncommenting and changing `achievement_title` and etc... This was a change I made a while ago and I tihnk I retained both when creating to not require a migration but maybe I missed something...

I want to take a closer look at this with some manual testing

Also:

Notes re making it possible to create things
Capabilities changes (talked about on slack)
When deleting via the UI I think we might need to do something else that happens when deleting the earned engagement via the button(s) on the reporting screens

phpcs.xml
```diff
@@ -83,6 +83,7 @@
 
 		<exclude-pattern>templates/*.php</exclude-pattern>
 		<exclude-pattern>templates/**/*.php</exclude-pattern>
+		<exclude-pattern>uninstall.php</exclude-pattern>
```
thomasplevy:
```suggestion
```

uninstall.php
```diff
@@ -1,11 +1,12 @@
 <?php
 /**
- * LifterLMS Uninstall
+ * LifterLMS Uninstall.
  *
  * @package LifterLMS/Main
  *
  * @since 1.0.0
- * @version 4.5.0
+ * @since [version] Delete `llms_my_achievement` and `llms_form` post type when removing all data.
```
thomasplevy:
```suggestion
```

phpcs.xml
thomasplevy: I don't think we should exclude this in order to add a changelog to the fileheader. In this case we don't have a good precedent for how to record the changelog on the uninstall.php and I think we can just *not do it*

includes/admin/post-types/meta-boxes/fields/llms.class.meta.box.fields.php
```diff
@@ -72,15 +73,19 @@ public function output() {
...
+			<?php endif ?>
 				<?php echo $this->field['desc']; ?>
 				<?php
 				if ( isset( $this->field['required'] ) && $this->field['required'] ) :
 					?>
 					<em>(required)</em><?php endif; ?>
 			</div>
-			<?php
+		<?php endif; ?>
+		<?php
```
thomasplevy:
```suggestion
		<?php endif; 
```

includes/traits/llms-trait-earned-engagement-meta-box.php
```diff
@@ -0,0 +1,177 @@
...
+ * @since [version]
+ * @version [version]
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * LifterLMS Eearned Engagements (Certificate/Achievement) Meta Box trait.
+ *
+ * **This trait should only be used by classes that extend from the {@see LLMS_Admin_Metabox} class.**
```
thomasplevy:
```suggestion
 * This trait should only be used by classes that extend from the {@see LLMS_Admin_Metabox} class.
```

Emphasis is unnecessary, I think.

includes/traits/llms-trait-earned-engagement-meta-box.php
```diff
@@ -0,0 +1,177 @@
...
+				array(
+					$user_id,
+					$post_id,
+				)
+			)
+		);
+	}
+
+	/**
+	 * Wheter the user has earned the engagement.
```
thomasplevy:
```suggestion
	 * Whether the user has earned the engagement.
```
Typo :muscle:

includes/traits/llms-trait-earned-engagement-meta-box.php
```diff
@@ -0,0 +1,177 @@
...
+	 *
+	 * @since [version]
+	 *
+	 * @param int $user_id The student's user id.
+	 * @param int $post_id The earned engagement id.
+	 */
+	private function has_user_earned( $user_id, $post_id ) {
```
thomasplevy: I'm thinking that this could be a "public" function, something like `llms_has_user_earned_engagement( $user_id, $engagement_type, $related_post_id )` and ultimately it could lookup the post via a query method designed to query engagements.

On my branch I've recorded a note about phasing out the usage of user postmeta for storing this information (though I think it should still be stored for now) and instead querying by post type and author.

I think we can merge these together and rewrite both sides to make everything more dry and also publicly available...

includes/traits/llms-trait-earned-engagement-meta-box.php
```diff
@@ -0,0 +1,177 @@
...
+	/**
+	 * Wheter the user has earned the engagement.
+	 *
+	 * @since [version]
+	 *
+	 * @param int $user_id The student's user id.
+	 * @param int $post_id The earned engagement id.
+	 * @return void
+	 */
+	private function log_earned_engament( $user_id, $post_id ) {
```
thomasplevy: On my branch `LLMS_Engagement_Handler::create()` (a private method) stores the user postmeta and triggers this action.

I think the storage of the postmeta and the triggering of the action can be moved into a public method, `LLMS_Engagement_Handler::create_actions( ... )` or something of the sort. You can use that method here and the `create()` method could use it as well.

includes/traits/llms-trait-earned-engagement-meta-box.php
```diff
@@ -0,0 +1,177 @@
...
+	protected function save_field( $post_id, $field ) {
+
+		if ( $this->prefix . 'student' === $field['id'] && isset( $_POST[ $field['id'] ] ) ) { //phpcs:ignore -- nonce already verified in `LLMS_Admin_Metabox::save`.
+			$this->log_earned_engament( llms_filter_input( INPUT_POST, $field['id'], FILTER_SANITIZE_NUMBER_INT ), $post_id );
+		}
+
+		/**
+		 * Skip saving _llms_student field, only used to award an engagement, it's not a post field.
+		 */
+		if ( isset( $field['id'] ) && $this->prefix . 'student' === $field['id'] ) {
```
thomasplevy: On my branch I'm switching to save the user id to the `post_author` field. This will make querying data more efficient in the long run (instead of relying on our congested user postmeta table. We need a data migration and I plan to continue storing the postmeta data for backwards compatiblity reasons though.

includes/models/model.llms.user.achievement.php
```diff
@@ -15,16 +15,17 @@
  *
  * @since 3.8.0
  * @since 3.18.0 Unknown.
+ * @since [version] Declare `achievement_title` and `achievement_content` properties.
```
thomasplevy: The only merge conflicts that are apparent with your branch and mine are with this file. I've recorded some todos on my branch related to these two fields. I think your changes can be discarded and we can figure out a better way to manage this after the merge.

thomasplevy (commented):
I haven't finished looking through all of this yet. This is just what I've noticed so far and isn't exactly a code review so much as notes on how your changes and my changes interact with each other.

includes/admin/post-types/meta-boxes/fields/llms.class.meta.box.fields.php
eri-trabiccolo: Cannot accept, because of phpcs, which is the reason why I wrote it that way :D
gonna find another way.

includes/traits/llms-trait-earned-engagement-meta-box.php
eri-trabiccolo: Alright, this means that I have to save that field value into the `post_author` on creation... ?

includes/traits/llms-trait-earned-engagement-meta-box.php
eri-trabiccolo: Actually I meant to make this method "temporary": just to make the current development working fine.
It's based on this: https://github.com/gocodebox/lifterlms/blob/5.4.1/includes/achievements/class.llms.achievement.user.php#L105
a more complex query which required the `*_template_id` which we don't have here.

includes/traits/llms-trait-earned-engagement-meta-box.php
eri-trabiccolo: I agree <3

includes/models/model.llms.user.achievement.php
eri-trabiccolo: Fixed...
Are you fine with me rebasing (not merging) my branch against yours and working on it?

includes/traits/llms-trait-earned-engagement-meta-box.php
eri-trabiccolo: I'm going to slip that code then, I guess I'll use a static public method.

includes/traits/llms-trait-earned-engagement-meta-box.php
thomasplevy: Yes :-/

includes/models/model.llms.user.achievement.php
thomasplevy: That's fine with me, I was planning on merging your branch into mine though so that I can review what you're working on more thoroughly and so that we can start to sync up

eri-trabiccolo: Alright, rebased and force pushed.

thomasplevy:
@eri-trabiccolo I know we have a lot more work to do yet, we'll keep working but I wanted to get us synced up

PR #1840, opened 2021-10-28
Unify the way we register our post types and filter their registration data

eri-trabiccolo:
## Description
Unify the way we register our post types and filter their registration data.

Also deprecate all those filter which used the post type name in place of the unprefixed (without the trailing `llms_`) one: of course when the post type name and the unprefixed name are not the same - e.g. for the `course` post type.

My approach is simple, I didn't want to add another param to determine whether the deprecation filter should be fired or not for a post type, meaning that while `lifterlms_register_post_type_llms_engagement` existed before and if used it'll fire a deprecation warning but will continue to work, `lifterlms_register_post_type_llms_membership` never existed, but now it'll start existing and working although it'll fire a deprecation warning.

## How has this been tested?
manually and unit test on filters deprecation

## Types of changes
code improvement (really minor refactoring).

## Checklist:

### DIFF
--- a/.changelogs/llms-register-post-type.yml
+++ b/.changelogs/llms-register-post-type.yml
@@ -1,0 +1,22 @@
+significance: minor
+type: dev
+entry: All the LifterLMS post types are now registered using the static method `LLMS_Post_Types::register_post_type()`.
+The following filters have been deprecated
+
+- `lifterlms_register_post_type_llms_engagement`
+- `lifterlms_register_post_type_llms_achievement`
+- `lifterlms_register_post_type_llms_certificate`
+- `lifterlms_register_post_type_llms_my_certificate`
+- `lifterlms_register_post_type_llms_email`
+- `lifterlms_register_post_type_llms_coupon`
+- `lifterlms_register_post_type_llms_voucher`
+
+and replaced respectively with
+
+- `lifterlms_register_post_type_engagement`
+- `lifterlms_register_post_type_achievement`
+- `lifterlms_register_post_type_certificate`
+- `lifterlms_register_post_type_my_certificate`
+- `lifterlms_register_post_type_email`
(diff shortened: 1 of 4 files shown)


eri-trabiccolo:
@thomasplevy I added a "convoluted" changelog entry... :D I don't think it's in the right format so let me know. <3

.changelogs/llms-register-post-type.yml
```diff
@@ -0,0 +1,22 @@
+significance: minor
+type: dev
```
thomasplevy, line 2:
```suggestion
type: deprecated
```

.changelogs/llms-register-post-type.yml
```diff
@@ -0,0 +1,22 @@
+significance: minor
```
thomasplevy, line 1: I'll have to check on the formatting of a multi-line entry. I didn't think to test that out this might break the log writer or it might technically be invalid yaml... For now this is fine I'll check that out later

includes/class.llms.post-types.php
```diff
@@ -319,6 +324,15 @@ public static function register_post_type( $name, $data ) {
 
 			$unprefixed_name = str_replace( 'llms_', '', $name );
 
+			if ( $unprefixed_name !== $name ) {
```
thomasplevy, line 327: This is a good way to handle this but... if someone uses a filter that *never existed*, for example `lifterlms_register_post_type_llms_membership` it'll trigger this...

As I type it I realize that *that doesn't matter* and *I don't care*

thomasplevy (approved):

includes/class.llms.post-types.php
eri-trabiccolo: yeah I mentioned it, in the PR description, that I didn't care about that either <3

PR #1846, opened 2021-11-04
Engagement images and more

thomasplevy:
Part of #1771 

Work on meta key switching and switching meta images to featured images


includes/class.llms.post-types.php
```diff
@@ -121,6 +122,7 @@ public static function deregister_sitemap_post_types( $post_types ) {
 			$post_types['lesson'],
 			$post_types['llms_quiz'],
 			$post_types['llms_certificate'],
+			$post_types['llms_my_achievment'],
```
eri-trabiccolo:
```suggestion
			$post_types['llms_my_achievement'],
```
:D

includes/class.llms.post-types.php
```diff
@@ -770,20 +773,24 @@ public static function register_post_types() {
...
+					'view_item'             => __( 'View Achievement Template', 'lifterlms' ),
+					'search_items'          => __( 'Search Achievement Templates', 'lifterlms' ),
+					'not_found'             => __( 'No Achievement Templates found', 'lifterlms' ),
+					'not_found_in_trash'    => __( 'No Achievement Templates found in trash', 'lifterlms' ),
+					'parent'                => __( 'Parent Achievement Template', 'lifterlms' ),
+					'menu_name'             => _x( 'Achievements', 'Admin menu name', 'lifterlms' ),
+					'featured_image'        => __( 'Achievement Image', 'lifterlms' ),
+					'set_featured_image'    => __( 'Set achievement  image', 'lifterlms' ),
+					'remove_featured_image' => __( 'Remove achievement image', 'lifterlms' ),
+					'use_featured_image'    => __( 'Use achievement image', 'lifterlms' ),
```
eri-trabiccolo, line 793: this is exactly why I hate this phpcs alignment rule, because of a single fucking line (the longer) you have to modify the whole array, potentially generating merge conflicts.

includes/functions/llms.functions.certificate.php
```diff
@@ -11,80 +11,125 @@
...
-		$cert->certs['LLMS_Certificate_User']->init( $id, get_current_user_id(), $id );
-		$certificate_content = $cert->certs['LLMS_Certificate_User']->get_content_html();
+		// If `$id` was empty to use the global, ensure an id is available in filter on the return.
+		$id = $certificate->get( 'id' );
 
-	} else {
-
-		$certificate_content = get_the_content();
+		// Get merged content for templates or the already-merged content of the earned cert, retrieve the raw because we filter it again below.
+		$content = 'llms_certificate' === get_post_type( $id ) ? $certificate->merge_content() : $certificate->get( 'content', false );
```
eri-trabiccolo:
```suggestion
		$content = 'llms_certificate' === get_post_type( $id ) ? $certificate->merge_content() : $certificate->get( 'content', true );
```
If we want the `raw` content, then we need to pass true:
https://github.com/gocodebox/lifterlms/blob/3f516b3f5301b1de0c1457be56fea11c556ac95c/includes/abstracts/abstract.llms.post.model.php#L517

includes/models/model.llms.user.certificate.php
```diff
@@ -46,6 +46,17 @@ class LLMS_User_Certificate extends LLMS_Abstract_User_Engagement {
 		'sequential_id'        => 'absint',
 	);
 
+	/**
+	 * Array of default property values
```
eri-trabiccolo:
```suggestion
	 * Array of default property values.
```

includes/functions/llms.functions.certificate.php
```diff
@@ -11,80 +11,125 @@
...
+/**
+ * Retrieve the default certificate background image for a given certificate.
+ *
+ * @since [version]
+ *
+ * @param int $certificate_id WP_Post ID of the earned certificate. This is passed so that anyone filtering the default image could
+ *                            provide a different default image based on the certificate.
+ * @return string The full image source url.
+ */
+function llms_get_certificate_default_image( $certificate_id ) {
```
eri-trabiccolo: Are we dropping the default certificate option in Settings->Engagements->Certificates?

includes/models/model.llms.user.certificate.php
```diff
@@ -114,6 +125,100 @@ public function can_user_view( $user_id = null ) {
...
+
+			/**
+			 * Filters the display width of the default certificate background image.
+			 *
+			 * @since 2.2.0
+			 *
+			 * @param int $width          Display width of the image, in pixels.
+			 * @param int $certificate_id WP_Post ID of the earned certificate.
+			 */
+			$width = apply_filters( 'lifterlms_certificate_background_image_placeholder_width', 800, $id );
```
eri-trabiccolo, line 177: Are we also dropping the default width/height Settings->Engagements->Certificates?

includes/models/model.llms.user.certificate.php
```diff
@@ -114,6 +125,100 @@ public function can_user_view( $user_id = null ) {
...
+	 *     @type int    $height     The image display height, in pixels.
+	 *     @type bool   $is_default Whether or not the default image was returned.
+	 * }
+	 */
+	public function get_background_image() {
+
+		$id = $this->get( 'id' );
+
+		// Don't retrieve a size if legacy mode is enabled.
+		$size = llms_parse_bool( get_option( 'lifterlms_certificate_legacy_image_size', 'yes' ) ) ? 'full' : 'lifterlms_certificate_background';
```
eri-trabiccolo, line 147: I wonder if it still makes sense to refer to that option (I mean the label and the description in Settings -> Engagements -> Certificates) as "legacy".
Let me explain:
a) if I use the same certificate background for all my certificates then either I upload something with the default size, or something with custom size and either I tweak the image width/height default settings or I check that option above. Same result.
b) if I use a different thumbnail for each certificate (certificate template) I guess I want exactly that image to be used, not a thumbnail version of it.
I mean why should I need a thumbnail? Lazyness?
I surely miss the history behind the introduction of this image size...

includes/models/model.llms.user.certificate.php
```diff
@@ -236,7 +341,7 @@ protected function get_merge_data() {
 	 *
 	 * @return void
```
eri-trabiccolo:
```suggestion
	 * @return string
```

eri-trabiccolo (changes requested):

includes/models/model.llms.user.certificate.php
```diff
@@ -114,6 +125,100 @@ public function can_user_view( $user_id = null ) {
...
+	 * @return array {
+	 *     Returns an associative array of information about the background image.
+	 *
+	 *     @type string $src        The image source url.
+	 *     @type int    $width      The image display width, in pixels.
+	 *     @type int    $height     The image display height, in pixels.
+	 *     @type bool   $is_default Whether or not the default image was returned.
+	 * }
+	 */
+	public function get_background_image() {
```
eri-trabiccolo, line 142: ~Ohhh just noticed this method is not backward compatible.
Meaning that if someone earned a certificate with the current stable and that certificate template has a custom background image, when viewing the earned certificate with this PR I see the default background.~
Oh stupid me, there will be a db update that'll migrate them.

includes/class.llms.post-types.php
thomasplevy: I don't care if you don't like coding standards. <3

includes/models/model.llms.user.certificate.php
thomasplevy: We used to have a lot of users *complain* that the certificate sizing was tied to the size of the uploaded image. It was constructed this way initially so that users *didn't have to worry* about what size to make their image. Instead they could upload their image at whatever size they'd prefer and it would automatically set the size based on the image size. But over time it resulted it a lot of users saying "I DON'T KNOW WHAT SIZE TO MAKE THE IMAGE" and I'd have to continually explain (and link to documentation) explaining how it worked.

My solution -- which does seem dumb today -- was conceived so that users could control the image size with an option in favor of being able to control it with the image.

It's all maybe a dumb conversation because what we actually should be doing is controlling the size of the canvas based on common print dimensions (and using their equivalent pixels on screen) and stretching the background image to fill the space at 100% width and height.

Possibly we could rewrite this (now) to have a default width (selected from a dropdown of predetermined print sizes (A4, etc...) and also having that same option available on a per-certificate setting. Default used if no override is present.

Perhaps converting the existing options to a "custom" size (enabled by default via a migration) and calling the "legacy" sizing "automatic" sizing (and enable that by default with a migration if it's enabled).

maybe... I don't know...

My memory is also terrible perhaps I'm remembering all of this wrong.

eri-trabiccolo: Alright, we can _fix_ this later, let's think about this a bit more...

includes/class.llms.post-types.php
eri-trabiccolo: I like them, I just don't like that part, also because contradicts the other rule that requires a comma after the last array item...

includes/functions/llms.functions.certificate.php
thomasplevy: Yea i'm just dumb.

includes/functions/llms.functions.certificate.php
thomasplevy: No, but it doesn't exist yet, did you write it and I've missed it?

includes/models/model.llms.user.certificate.php
thomasplevy: No. Those options are used at the end of this file for registering the certificate image size (which is what's requested as the image size when legacy mode is *not* enabled: lifterlms_certificate_background

includes/functions/llms.functions.certificate.php
eri-trabiccolo: I guess I made it up.

includes/models/model.llms.user.certificate.php
eri-trabiccolo: Ok, makes sense to have them static as we know the size of our placeholder.

includes/models/model.llms.user.certificate.php
thomasplevy: I got thinking about how to implement and it might not be so hard...

But the thing that was hurting my brain is something you addressed at some point and I can't find... The problem is that paper sizes don't equate well with pixels. Pixels are relative to the DPI whereas paper size is absolute.

So if we think about the certificate sizing option to be the size of the page / certificate canvas (rather than as the size of the background image) and we center the background image vertically and horizontally (and maybe stretch it to fit?) that's all cool... but we shouldn't use pixels, we should use the actual dimensions of the paper. Which we can actually do in CSS (though I don't have a lot of experiencing writing screens in inches / mm via CSS)...

So we could build a dropdown with some default common paper sizes (this would probably be good to start, we can filter it and if we get requests for other sizes just add them in) and then have another option to select orientation (portrait vs landscape).

The existing sizing option could map to the "CUSTOM" option which would be populated with the existing pixels stored in the DB. But then my head breaks... If we want to make our lives easier we should probably switch to mm/inches so then what do we have the user input their pixels? Or maybe we "break" and we don't allow them to define a custom size at all anymore... 

This array contains some common sizes we could use -- I've opted to add the dimensions as `[ mm, in ]` so that we don't have to prioritize US over the rest of the (sane) world (or vice versa). That further complicates user input if we allow them to define their own size, do they have to enter both? I guess we could have them select the unit (and generate the other unit for them automatically)?

Anyway... maybe this is something for next time....

```
	public function get_sizes() {

		return array(
			// ISO 216 sizes.
			'A3' => array(
				'name' => _x( 'A3', 'Paper size name', 'lifterlms' ),
				'width' => array( 297, 11.7 ),
				'height' => array( 420, 16.5 ),
			),
			'A4' => array(
				'name' => _x( 'A4', 'Paper size name', 'lifterlms' ),
				'width' => array( 210, 8.3 ),
				'height' => array( 297, 11.7 ),
			),
			'A5' => array(
				'name' => _x( 'A5', 'Paper size name', 'lifterlms' ),
				'width' => array( 148, 5.8 ),
				'height' => array( 210, 8.3 ),
			),
			// North American sizes.
			'LETTER' => array(
				'name' => _x( 'North American Letter', 'Paper size name', 'lifterlms' ),
				'width' => array( 216, 8.5 ),
				'height' => array( 279, 11 ),
			),
			'LEGAL' => array(
				'name' => _x( 'North American Legal', 'Paper size name', 'lifterlms' ),
				'width' => array( 216, 8.5 ),
				'height' => array( 356, 14 ),
			),
			'LEDGER' => array(
				'name' => _x( 'North American Ledger', 'Paper size name', 'lifterlms' ),
				'width' => array( 279, 11 ),
				'height' => array( 432, 17 ),
			),
			'CUSTOM' => array(
				'name' => _x( 'Custom', 'Paper size name', 'lifterlms' ),
				'width' => array( ?, ? ),
				'height' => array( ?, ? ),
			),

		);

	}
```

includes/functions/llms.functions.certificate.php
thomasplevy: It's pushed to this branch now

includes/models/model.llms.user.certificate.php
eri-trabiccolo: paper size <=> px
we included a link in our documentation: https://lifterlms.com/docs/certificate-background-image-sizes/
But yeah I see your point.
Ok, I think what we have to do is to provide something like, for the custom size:
- dpi select (common dpis)
defaults is 96
- width
input number (allows floats) - select (in|mm|px)
- height
input number (allows floats) - select (in|mm|px)

Then depending on their choice we'll be able to do our conversion, right?
Complicated? Yeah, but not our fault, instead of Custom we could call that Advanced :D

includes/functions/llms.functions.certificate.php
eri-trabiccolo: @thomasplevy what happened to this function?
it has been removed, but it's still referenced here:
https://github.com/gocodebox/lifterlms/blob/engagement-refactoring/includes/models/model.llms.user.certificate.php#L154

thomasplevy: I merged early. I shouldn't have been working on the weekend during naptime

PR #1853, opened 2021-11-11
Engagement refactoring fix failing tests

eri-trabiccolo:
fixes some failing tests.

There are a bunch of other tests that are erroring because of the now missing `create_attachment()` test method, not sure what you want to do with it though, so I haven't looked for a solution.

### DIFF
--- a/includes/class.llms.engagements.php
+++ b/includes/class.llms.engagements.php
@@ -647,7 +647,9 @@ class LLMS_Engagements {
 	 */
 	public function unschedule_delayed_engagements( $post_id, $post = null ) {
 
+		// @todo Remove compatibility with WP < 5.5 when bumping the minimum WP required version to 5.5+
 		$post_type = $post ? $post->post_type : get_post_type( $post_id );
+
 		if ( 'llms_engagement' === $post_type ) {
 			as_unschedule_all_actions( '', array(), $this->get_delayed_group_id( $post_id ) );
 		}
--- a/includes/class.llms.engagements.php
+++ b/includes/class.llms.engagements.php
@@ -637,15 +637,18 @@ class LLMS_Engagements {
 	 *
 	 * This is the callback function for deleted engagement posts.
 	 *
+	 * The `deleted_post` action param `$post` has been added since WordPress 5.5.0.
+	 *
 	 * @since [version]
 	 *
 	 * @param int          $post_id WP_Post ID.
 	 * @param WP_Post|null $post    Post object of the deleted post.
 	 * @return void
 	 */
-	public function unschedule_delayed_engagements( $post_id, $post ) {
+	public function unschedule_delayed_engagements( $post_id, $post = null ) {
 
-		if ( 'llms_engagement' === $post->post_type ) {
+		$post_type = $post ? $post->post_type : get_post_type( $post_id );
+		if ( 'llms_engagement' === $post_type ) {
 			as_unschedule_all_actions( '', array(), $this->get_delayed_group_id( $post_id ) );
 		}
 
--- a/tests/phpunit/framework/class-llms-settings-page-test-case.php
+++ b/tests/phpunit/framework/class-llms-settings-page-test-case.php
@@ -3,6 +3,7 @@
  * LifterLMS Unit Test Case Base class
  *
  * @since 3.37.3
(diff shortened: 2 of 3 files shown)


thomasplevy:
https://github.com/gocodebox/lifterlms-tests/blob/306a3189378c3a45e36edb1d1fb58171c4fea28b/framework/traits/trait-llms-unit-test-case-base.php#L74

thomasplevy:
Upgraded composer here: https://github.com/gocodebox/lifterlms/blob/a175fd7cb5da2f318cd74dcec899c21c2282fde6/composer.json#L39

eri-trabiccolo:
Oh alright locally I had those tests failing, my bad, there are other tests failing here, ok. <3

includes/class.llms.engagements.php
```diff
@@ -637,15 +637,18 @@ private function trigger_engagement( $data, $delay ) {
 	 *
 	 * This is the callback function for deleted engagement posts.
 	 *
+	 * The `deleted_post` action param `$post` has been added since WordPress 5.5.0.
```
thomasplevy, line 640: Good catch. I suggest this just get changed to work on < 5.5.0 only and not worry about whether or not `$post` is passed. We can just always rely on `get_post_type()` and drop the second parameter entirely.

tests/phpunit/framework/class-llms-settings-page-test-case.php
```diff
@@ -98,7 +100,7 @@ public function test_get_settings() {
 
 		$mock   = array_keys( $settings );
 		$actual = $this->get_settings_ids();
-		$this->assertEquals( $mock, $actual );
+		$this->assertEqualSets( $mock, $actual );
```
thomasplevy: WHat test is failing as a result of settings order?

The order kind of does matter here as the settings *should not* be being displayed in a random order on the frontend so they also should not be returned in a random order. The order is hardcoded into the arrays.

tests/phpunit/unit-tests/admin/settings/class-llms-test-settings-engagements.php
```diff
@@ -10,6 +10,7 @@
  *
  * @since 3.37.3
  * @since 3.40.0 Add tests for `get_settings_group_email_delivery()`.
+ * @version [version]
```
thomasplevy:
```suggestion
```

I know that I probably never updated our doc standards but we've talked about this before and we've stopped adding version tags in test class headers.

tests/phpunit/framework/class-llms-settings-page-test-case.php
eri-trabiccolo: alright, going to fix this.

tests/phpunit/unit-tests/admin/settings/class-llms-test-settings-engagements.php
eri-trabiccolo: Fine for me but I don't want to add a changelog entry there because it's already in the specific method doc.
Which is the reason why we stopped logging entries in the class doc blocks, do you agree?
So to me adding a version there was a good way to note the version of the tests.
Can I just remove that `version` and keep on not adding any entry in the file header?

includes/class.llms.engagements.php
eri-trabiccolo: we can of course, but if I already have an information ready, why should I call a function?
Why did they pass the `$post` parameter at all? To avoid calling `get_post( $post_id )` I guess, right?
I'd rather add a todo like "remove wp<5.5 compat code when we'll raise minimum wp version to 5.5+".
What do you think?

tests/phpunit/unit-tests/admin/settings/class-llms-test-settings-engagements.php
thomasplevy: yes!

includes/class.llms.engagements.php
thomasplevy: I was just thinking it seems like an bit of logic to worry about unnecessarily. Calling `get_post()` when the post has already been called loads it from memory and won't take that much longer than accessing the post directly. But you're right. Add a note.

PR #1855, opened 2021-11-11
Rename llms_certificate and llms_my_certificate post type slugs

thomasplevy:
## Description

Based on @eri-trabiccolo's proposed solution here: https://github.com/gocodebox/lifterlms/pull/1771#issuecomment-966107495

Instead of storing an option and putting a lifespan on the redirect I've opted to handle requests to the old URL and redirect them. If we store an option based on whether or not any shared certificates exist we run into an issue where bookmarks *might exist* for users that aren't sharing them (but have them saved somewhere for their own private purposes). 

This just performs a pretty safe redirect when a 404 is encountered, the URL contains `/my_certificate/` (or the equivalent transation), and an earned certificate exists for the slug (parsed from the url).

This should just do the job. In the scenario where someone has changed the slug via a filter this *will not work* but in that scenario we wouldn't want it to anyway since it's already been customized.

## How has this been tested?

+ Manually 
+ New PHPUnit tests

## Types of changes
+ New feature

## Checklist:

### DIFF
--- a/includes/class.llms.query.php
+++ b/includes/class.llms.query.php
@@ -300,11 +300,10 @@ class LLMS_Query {
 
 		global $wp, $wp_query;
 
-		$url = home_url( $wp->request );
-		$old = sprintf( '/%s/', _x( 'my_certificate', 'slug', 'lifterlms' ) );
-		if ( $wp_query->is_404() && false !== strpos( $url, $old ) ) {
-
-			$slug     = str_replace( $old, '', wp_parse_url( $url, PHP_URL_PATH ) );
+		$old  = sprintf( '/%s/', _x( 'my_certificate', 'slug', 'lifterlms' ) );
+		$path = wp_parse_url( home_url( $wp->request ), PHP_URL_PATH );
+		if ( $wp_query->is_404() && 0 === strpos( $path, $old ) ) {
+			$slug     = str_replace( $old, '', $path );
 			$new_post = get_page_by_path( $slug, 'OBJECT', 'llms_my_certificate' );
 			if ( $new_post ) {
 				llms_redirect_and_exit( get_permalink( $new_post->ID ) );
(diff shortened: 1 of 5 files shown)


eri-trabiccolo (commented):
I will better review this tomorrow, but this check on is_404 looks pretty smart... <3

includes/class.llms.query.php
```diff
@@ -280,6 +282,37 @@ public function maybe_404_certificate() {
...
+	 *
+	 * @return void
+	 */
+	public function maybe_redirect_certificate() {
+
+		global $wp, $wp_query;
+
+		$url = home_url( $wp->request );
+		$old = sprintf( '/%s/', _x( 'my_certificate', 'slug', 'lifterlms' ) );
+		if ( $wp_query->is_404() && str_contains( $url, $old ) ) {
```
eri-trabiccolo: `str_contains()` requires PHP 8+
```suggestion
		if ( $wp_query->is_404() && false !== strpos( $url, $old ) ) {
```

eri-trabiccolo (changes requested):
The only other thing to consider here is that potentially `maybe_redirect_certificate()` matches also `'https://www.example.com/something/my_certificate/something'`, and for that `$slug = str_replace( $old, '', wp_parse_url( $url, PHP_URL_PATH ) );` will return
`/somethingsomething`.
But I think you already considered this and evaluated it as too much of an edge case also because we already check for if `is_404` so we don't break anything. Right?

thomasplevy:
> The only other thing to consider here is that potentially `maybe_redirect_certificate()` matches also `'https://www.example.com/something/my_certificate/something'`, and for that `$slug = str_replace( $old, '', wp_parse_url( $url, PHP_URL_PATH ) );` will return `/somethingsomething`. But I think you already considered this and evaluated it as too much of an edge case also because we already check for if `is_404` so we don't break anything. Right?

This should handle this edge case: https://github.com/gocodebox/lifterlms/pull/1855/commits/b9156e46a2d783303eee8edad272c3bbcb891e1f

PR #1856, opened 2021-11-11
Create llms_my_certificates with a non-numeric unique post slug

thomasplevy:
## Description

Solves part 2 of https://github.com/gocodebox/lifterlms/pull/1771#issuecomment-965878426

Adds a function that generates a 3+ lowercase + numeric suffix to append to post slugs (in favor of `-1`, `-2`, etc...)

 For a certificate with the title of "Certificate of Completion" the generated slug might look like this:

 `certificate-of-completion-a3x` 
 `certificate-of-completion-10p`
 `certificate-of-completion-pya` 
etc... 

This only applies when a `llms_my_certificate` is awarded via an engagement trigger. If the certificate is created manually on the admin panel (or an existing certificate is edited) this will *not* run. The idea here is that the admin awarding the certificate manually should choose a unique slug (should they desire one) and then when updating a slug it should already be unique and if an admin chooses to change it that's on them.

`wp_unique_post_slug()` will still do it's thing, which could result in a slug like:  `certificate-of-completion-a3x-1` if the slug was updated manually to be a randomly generated slug that already exists.

The function will automatically increase the length of the random string as collisions are encountered while generating a string. 

There's a filter to customize the minimum length of the string. A very large site could conceivably run out of unique 3 digit strings and wish to speed things up (reduce DB queries found in wp_unique_post_slug()) by increasing the minimum length by reducing collisions.

Also there are 46,656 permutations of the dictionary when using 3 characters, 1,679,616 permutations when using 4. I think the default of 3 is good but I could be convinced to increase it to 4.

## How has this been tested?

+ Manually
+ New unit tests

## Types of changes

## Checklist:

### DIFF
--- a/tests/phpunit/unit-tests/class-llms-test-certificates.php
+++ b/tests/phpunit/unit-tests/class-llms-test-certificates.php
@@ -142,6 +142,62 @@ class LLMS_Test_Certificates extends LLMS_UnitTestCase {
 	}
 
 	/**
+	 * Test get_unique_slug()
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function test_get_unique_slug() {
+
+		$slugs = array();
+
+		$i = 0;
+		while ( $i < 250 ) {
+
+			$post = $this->factory->post->create_and_get( array(
+				'post_type' => 'llms_my_certificate',
+				'post_name' => llms()->certificates()->get_unique_slug( 'A Title' ),
+			) );
+
+			$slugs[] = $post->post_name;
+
+			$i++;
+
+		}
+
+		$this->assertEquals( 250, count( array_unique( $slugs ) ) );
+
+	}
+
+	/**
+	 * Test get_unique_slug() will increase the suffix string length after encountering collisions.
+	 *
+	 * @since [version]
+	 *
+	 * @return void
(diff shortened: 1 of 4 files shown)


eri-trabiccolo (approved):
Looks like it works :P

tests/phpunit/unit-tests/class-llms-test-certificates.php
```diff
@@ -141,6 +141,62 @@ public function test_get_export_with_cache() {
...
+
+		// Create a post with the '-aaa' suffix so we can have a real fake collision.
+		$post = $this->factory->post->create_and_get( array(
+			'post_type' => 'llms_my_certificate',
+			'post_name' => llms()->certificates()->get_unique_slug( 'A Title' ),
+		) );
+		$this->assertEquals( 11, strlen( $post->post_name ) );
+		$this->assertEquals( 'a-title-aaa', $post->post_name );
+
+		// This request will result find '-aaa' as a collision and then try 4 more times and then increase to 4 characters.s
```
eri-trabiccolo:
```suggestion
		// This request will result find '-aaa' as a collision and then try 4 more times and then increase to 4 characters.
```
💪🏻

PR #1864, opened 2021-11-19
Fix awarded achievement: must not show the full editor

eri-trabiccolo:
@thomasplevy 
fine with the strategy here?


thomasplevy:
@eri-trabiccolo I didn't do a test or full review but the strategy seems good

eri-trabiccolo:
@thomasplevy if you want to review and possibly merge this - it's ready, imho.

PR #1857, opened 2021-11-12
Syncing awarded certificates with their templates

eri-trabiccolo:
## Description
Initial work on syncing awarded certificates with their templates

per https://github.com/gocodebox/lifterlms/pull/1771#issuecomment-966158070

## How has this been tested?
manually (for now) - unit tests about to come

## Types of changes
New feature (non-breaking change which adds functionality)

## Checklist:
- [x] My code has been tested.
- [x] My code passes all existing automated tests. 
- [x] My code follows the LifterLMS Coding & Documentation Standards.

PR #1866, opened 2021-11-22
Sync awarded certificates with their templates

eri-trabiccolo:
## Description
Initial work on syncing awarded certificates with their templates

per https://github.com/gocodebox/lifterlms/pull/1771#issuecomment-966158070

## How has this been tested?
manually (for now) - unit tests about to come

## Types of changes

## Checklist:


thomasplevy:
@eri-trabiccolo this needs tests

eri-trabiccolo:
@thomasplevy 
yeah - meant to turn into a Draft, I just re-opened this https://github.com/gocodebox/lifterlms/pull/1857 which was closed because of the deleting of the `awards-query` branch.

includes/models/model.llms.user.certificate.php
```diff
@@ -689,23 +692,28 @@ public function sync( $context = 'update' ) {
 		}
 
 		$this->set( 'title', get_post_meta( $template_id, '_llms_certificate_title', true ) );
-		set_post_thumbnail( $this->get( 'id' ), get_post_thumbnail_id( $template_id ) );
+		if ( ! set_post_thumbnail( $this->get( 'post' ), get_post_thumbnail_id( $template_id ) ) ) {
+			delete_post_thumbnail( $this->get( 'post' ) );
+		}
```
thomasplevy: Good catch. Please update the existing tests to also test this new scenario that I had previously missed.

includes/admin/post-types/class.llms.meta.boxes.php
```diff
@@ -188,6 +189,68 @@ public function hide_meta_boxes() {
...
+				__( '%1$sSync%2$s %3$d awarded certificates', 'lifterlms' ),
+				sprintf(
+					'<a href="%1$s" onclick="%2$s">',
+					$sync_url,
+					$on_click
+				),
+				'</a>',
+				$awarded_certificates_number
+			)
+		);
```
thomasplevy: This string desires some `_n()` usage, eg "1 awarded certificates" on my site currently.

Also... we should load the post name strings from the post type as to avoid multiple translations unnecessarily.

Eg: `get_post_type_object()->labels` etc...

includes/controllers/class.llms.controller.certificates.php
```diff
@@ -131,6 +133,118 @@ public function maybe_handle_reporting_actions() {
...
+					__( 'You\'re trying to perform an invalid action.', 'lifterlms' )
+				);
+		}
+
+		$cert_id = llms_filter_input( INPUT_GET, 'post', FILTER_SANITIZE_NUMBER_INT );
+		if ( empty( $cert_id ) ) {
+			return $error;
+		}
+
+		return self::{$method}( $cert_id );
```
thomasplevy: I'm almost positive this fancy syntax is going to break the WP Doc parser like this method does:

https://github.com/gocodebox/developer.lifterlms.com/blob/b55f970683ac03bae16e3666ce3aa600470007ee/wp-content/themes/llms-dev/inc/parser.php#L38-L49

Please have mercy on me and just write this less fancy even though it's going to hurt you to do so:

The referenced function breaks because of `->{` -- I'm guessing the same issue will happen with `::{` (though I'm not positive)

The error, as far as I can tell, is upstream even from the PHP doc parser and lies in the php parser library they use. I haven't tried to report or debug further than skipping parsing in our dev docs. I don't want to have another exception <3

includes/controllers/class.llms.controller.certificates.php
```diff
@@ -131,6 +133,118 @@ public function maybe_handle_reporting_actions() {
...
+	 * Sync awarded certificate with its template.
+	 *
+	 * @since [version]
+	 *
+	 * @param int $cert_id Awarded certificate id.
+	 * @return void|WP_Error
+	 */
+	private static function sync_awarded_certificate( $cert_id ) {
+
+		if ( ! current_user_can( get_post_type_object( 'llms_my_certificate' )->cap->edit_post, $cert_id ) ) {
```
thomasplevy:
```suggestion
		if ( ! current_user_can( 'edit_post', $cert_id ) ) {
```

Doesn't WP take care of figuring out what the actual cap is automatically based on the id passed in?

includes/controllers/class.llms.controller.certificates.php
```diff
@@ -131,6 +133,118 @@ public function maybe_handle_reporting_actions() {
...
+					__( 'Sorry, you are not allowed to edit the awarded certificate #%d.', 'lifterlms' ),
+					$cert_id
+				),
+				compact( 'cert_id' )
+			);
+		}
+
+		$redirect_url = get_edit_post_link( $cert_id, 'raw' );
+
+		$sync = ( new LLMS_User_Certificate( $cert_id ) )->sync();
```
thomasplevy:
```suggestion
		$sync = llms_get_certificate( $cert_id )->sync();
```

includes/controllers/class.llms.controller.certificates.php
```diff
@@ -131,6 +133,118 @@ public function maybe_handle_reporting_actions() {
...
+
+		if ( ! current_user_can( get_post_type_object( 'llms_my_certificate' )->cap->edit_posts ) ) {
+			return new WP_Error(
+				'llms-sync-awarded-certificates-insufficient-permissions',
+				__( 'Sorry, you are not allowed to edit awarded certificates', 'lifterlms' )
+			);
+		}
+
+		// Trigger background sync.
+		do_action( 'llms_do_awarded_certificates_bulk_sync', $certificate_template_id );
```
thomasplevy, line 258: Needs action doc

thomasplevy (changes requested):
In the interest in not forcing a complete rewrite of all this code moving the general sync action into it's own metabox will enable us to use the same code in the block editor without having to rewrite everything. I think, generally, it should work without issue in the block ed, we just need it in it's own metabox so that we can load it in both places.

In that way I think the title should be "Sync Awarded Certificates" and the metabox content should be a button that says "Sync" (which pops on the onlick alert as it does now). Below the button a description could identify the number of certificates that will be synced.

Instead of returning early and displaying an empty metabox the button should be disabled if there are no children to sync.

I'm not done reviewing this, just my notes so far (it looks good overall though)

includes/controllers/class.llms.controller.certificates.php
eri-trabiccolo: ok (old)man

includes/controllers/class.llms.controller.certificates.php
eri-trabiccolo: I guess you're right, we're end up doing this, since `map_meta_cap` is `false` for awarded certificates.
https://github.com/WordPress/wordpress-develop/blob/5.8.1/src/wp-includes/capabilities.php#L171
I guess I have this habit of specify that that way... dunno.

includes/controllers/class.llms.controller.certificates.php
eri-trabiccolo: you lazy but ok

includes/admin/post-types/class.llms.meta.boxes.php
eri-trabiccolo: So I won't use _n since we're loading the post type labels...
Let's try this way.

PR #1867, opened 2021-11-22
Allow DB upgrade functions to be declared within a namespace

thomasplevy:
## Description

When creating a database migration we currently prefix any update routine functions with a version number (such as `llms_update_500_{$function_name()` for version 5.0.0). When we're developing and testing these functions it becomes a pain when we determine the version number of the release will change as we then need to update methods, their references in the schema, and so.

This PR introduces the ability to list functions without a prefix in the schema and then add a `namespace=true` to the schema. This will automatically load the callback functions with the namespace `LLMS\Updates\Version_X_X_X` where `X_X_X` is a namespace-safe version number (eg: 5.0.0 becomes 5_0_0)

When changing the version nubmer the migration will be included with we then have to update the filename, namespace declaration, and the schema's version number. There's still a bit to change but it's less cumbersome than our current system.

This PR is backwards compatible so that namespacing an upgrade routine is optional. Existing migration schemas will not need to be updated and will continue to work as expected.

Additionally this supports a custom namespace should a plugin want to extend our upgrader and use a different namespace.

Lastly, this also autoloads all files found in the `includes/functions/updates` directory so that we don't need to add or update a reference in the `llms.functions.update.php` file should we rename the file.

## How has this been tested?

## Types of changes
+ New feature

## Checklist:


includes/class-llms-db-ugrader.php
```diff
@@ -101,6 +101,37 @@ public function can_auto_update() {
 
 	}
 
+	/**
+	 * Retrieve the callback's prefix string based on the schema's namespace declaration.
+	 *
+	 * If `$namespace` is empty, no prefix will be added.
+	 * If `$namespace` is `true`, the namespace is assumed to be `LLMS\Updates`.
+	 * If `$namespace` is a string, that string will be used.
```
eri-trabiccolo:
```suggestion
	 * If `$info['namespace']` is empty, no prefix will be added.
	 * If `$info['namespace']` is `true`, the namespace is assumed to be `LLMS\Updates`.
	 * If `$info['namespace']` is a string, that string will be used.
```

includes/class-llms-db-ugrader.php
```diff
@@ -101,6 +101,37 @@ public function can_auto_update() {
...
+	 *
+	 * If a namespace is found, `\Version_X_X_X` will automatically be appended to the namespace. The
+	 * string `X_X_X` is the database version for the upgrade substituting underscores for dots.     *
```
eri-trabiccolo:
```suggestion
	 * string `X_X_X` is the database version for the upgrade substituting underscores for dots.
```

includes/class-llms-db-ugrader.php
```diff
@@ -101,6 +101,37 @@ public function can_auto_update() {
...
+	 *
+	 * @param array  $info    Upgrade schema array.
+	 * @param string $version Version string for the upgrade.
+	 * @return string
+	 */
+	protected function get_callback_prefix( $info, $version ) {
+
+		if ( ! empty( $info['namespace'] ) ) {
+
+			$ver = explode( '-', $version ); // Drop prerelease data.
```
eri-trabiccolo, line 124: So this "won't work" with pre-releases.
Like we can't use this for a db update from `1.0.0-beta1` to `1.0.0-beta2`.
Shall we note this in the method documentation?

eri-trabiccolo (changes requested):
Looks good, minor docs stuff.

PR #1868, opened 2021-11-23
Add LLMS_Media_Protector class.

thomasplevy:
@pondermatic I'm opening this on your behalf so I can record some notes and thoughts I'm having while I'm working through this


includes/class-llms-media-protector.php
```diff
@@ -0,0 +1,640 @@
...
+	 */
+	public function set_upload_subdirectory( string $upload_subdirectory ) {
+
+		// Add leading slash.
+		if ( strpos( $upload_subdirectory, '/' ) !== 0 ) {
+			$upload_subdirectory = '/' . $upload_subdirectory;
+		}
+
+		// Strip trailing slash.
+		$upload_subdirectory = rtrim( $upload_subdirectory, '/' );
```
thomasplevy: Nitpicking but I like using the WP api where possible: https://developer.wordpress.org/reference/functions/untrailingslashit/

```suggestion
		$upload_subdirectory = untrailingslashit( $upload_subdirectory );
```

includes/class-llms-media-protector.php
```diff
@@ -0,0 +1,640 @@
...
+	 *                          The defaults are set in {@see media_handle_upload()}.
+	 * @param array  $overrides Optional. Override the {@see wp_handle_upload()} behavior.
+	 * @return int|WP_Error Post ID of the media file or a WP_Error object on failure.
+	 */
+	public function handle_upload(
+		string $file_id,
+		int $post_id,
+		array $post_data = array(),
+		array $overrides = array( 'test_form' => false )
+	) {
```
thomasplevy: I'm thinking that for the sake of usability it might make sense to add an extra parameter here, probably after `$post_id` that would accept a string, default to `AUTHORIZATION_FILTER_KEY` and accept an empty string to disable authorization requirements.

This would load the value into `$post_data['meta_input']` making it a bit easier for developers to work with this function (reducing the lines of code necessary to upload a protected file, from your example in the class doc:

```
$post_data['meta_input'] = array( LLMS_Media_Protector::AUTHORIZATION_FILTER_KEY => 'llms_sl_authorize_media_view' );
$id = $media->handle_upload( 'image', 0, $post_data );
```

to:
```
$id = $media->handle_upload( 'image', 0, 'llms_sl_authorize_media_view' );
```

I'm suggesting this primarily because after reading through everything I had to actually read all the docs and examples (and then look up WP core `media_handle_upload()` before I realized how/where the auth filter meta value was being set.

includes/class-llms-media-protector.php
```diff
@@ -0,0 +1,640 @@
...
+	protected $upload_subdirectory = '/llms-uploads';
+
+	/**
+	 * Set up this class.
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function __construct() {
```
thomasplevy: I'm thinking that it might make sense to add parameters to the constructor to configure the upload directory.

The core defaults of `/llms-uploads` makes sense and I think that our add-ons (and probably 3rd parties) will want to put their own directory inside of that directory. But still guarding against a need to create an entirely separate directory makes sense.

I'm thinking that having a `$base_upload_subdirectory = '/llms-uploads'` and an additional `$upload_subdirectory = ''` would work. The final directory will be `$base_upload_subdirectory . $upload_subdirectory`. The constructor might look like this:

```suggestion
	public function __construct( string $upload_subdirectory = '', string $base_upload_subdirectory = '/llms-uploads' ) {
		// Clean input and set to protected class vars...
```

This would make it so that the example in the docs above could be shortened from:
```
 $media = new LLMS_Media_Protector();
 $media->set_upload_subdirectory( $media->get_upload_subdirectory() . '/social-learning' );
```

To:

```
 $media = new LLMS_Media_Protector(); // Core plugin.
 $media = new LLMS_Media_Protector( '/social-learning' ); // add-on.
 $media = new LLMS_Media_Protector( '', 'custom-dir' ); // Custom directory.
```

I suppose the input could also be an array so that in the last example we don't have to pass an empty string.

Thoughts?

includes/class-llms-media-protector.php
```diff
@@ -0,0 +1,640 @@
...
+	 *     @type string       $subdir  Subdirectory if uploads use year/month folders option is on.
+	 *     @type string       $basedir Path without subdirectory.
+	 *     @type string       $baseurl URL path without subdirectory.
+	 *     @type string|false $error   False or error message.
+	 * }
+	 * @return array
+	 */
+	public function upload_dir( array $uploads ) {
+
+		$uploads['path'] = $uploads['basedir'] . $this->upload_subdirectory . $uploads['subdir'];
```
thomasplevy: Just so I make sure I'm reading this right, the presence of `$uploads['subdir']` will maintain the year/month folder structure within the custom directory(s), right?

I think this makes sense but I want to make sure I'm reading it right.

includes/class-llms-media-protector.php
```diff
@@ -0,0 +1,640 @@
...
+		// Is the user authorized to view the file?
+		$is_authorized = $this->is_authorized_to_view( get_current_user_id(), $media_id );
+		if ( false === $is_authorized ) {
+			$content_type = $media_file->post_mime_type;
+			if ( 0 === stripos( $content_type, 'image/' ) ) {
+				// @todo Find or create an image to denote unauthorized access to an image file.
+				$file_name = LLMS_PLUGIN_DIR . 'assets/images/circle-with-a-slash.png';
+			} else {
+				header( 'HTTP/1.1 403 Forbidden' );
+				exit();
```
thomasplevy: This (and other occurrences) should use `llms_exit()`. The function is mocked in the test suite so that we can test the exit via an exception thrown.

https://github.com/gocodebox/lifterlms/blob/ae8dcae0be15abeabd032bdac0f7b199942dc59a/includes/functions/llms-functions-wrappers.php#L45-L62

```suggestion
				llms_exit();
```

includes/class-llms-media-protector.php
```diff
@@ -0,0 +1,640 @@
...
+		if ( false === file_exists( $file_name ) ) {
+			return;
+		}
+
+		// Is the user authorized to view the file?
+		$is_authorized = $this->is_authorized_to_view( get_current_user_id(), $media_id );
+		if ( false === $is_authorized ) {
+			$content_type = $media_file->post_mime_type;
+			if ( 0 === stripos( $content_type, 'image/' ) ) {
+				// @todo Find or create an image to denote unauthorized access to an image file.
```
thomasplevy: I'll get Chris to get our design team to make an asset for this. For now use some CC0 image as a placeholder.

includes/class-llms-media-protector.php
```diff
@@ -0,0 +1,640 @@
...
+		header( "Location: $url" );
+		exit();
+	}
+
+	/**
+	 * Serves the requested media file to the HTTP client.
+	 *
+	 * This method calls the {@see https://www.php.net/manual/function.exit.php `exit()`} function and does not return.
+	 *
+	 * @todo Maybe respond with 404 Not Found?
```
thomasplevy: I think we should serve a 404 if the file isn't found.

includes/class-llms-media-protector.php
```diff
@@ -0,0 +1,640 @@
...
+		// Optionally, use an alternate image size.
+		if ( ! is_null( $size ) || ! is_null( $icon ) ) {
+			$image     = wp_get_attachment_image_src( $media_id, $size, $icon );
+			$base_name = basename( $image[0] );
+			$file_name = dirname( $file_name ) . "/$base_name";
+		}
+
+		// Validate that the media file exists.
+		if ( false === file_exists( $file_name ) ) {
+			return;
```
thomasplevy: Probably should serve a 404 here, right?

includes/class-llms-media-protector.php
```diff
@@ -0,0 +1,640 @@
...
+		exit();
+	}
+
+	/**
+	 * Serves the requested media file to the HTTP client.
+	 *
+	 * This method calls the {@see https://www.php.net/manual/function.exit.php `exit()`} function and does not return.
+	 *
+	 * @todo Maybe respond with 404 Not Found?
+	 * @todo Maybe respond with 401 Unauthorized?
```
thomasplevy: The 403 response should respond with 401 if the user isn't logged in and 403 if the user is.

includes/class-llms-media-protector.php
```diff
@@ -0,0 +1,640 @@
...
+				exit();
+			}
+		}
+
+		/**
+		 * Determine how the media file should be served.
+		 *
+		 * @since [version]
+		 *
+		 * @param string $serve_method One of the {@see LLMS_Media_Protector::SERVE_REDIRECT LLMS_Media_Protector::SERVE_X} constants.
```
thomasplevy:
```suggestion
		 * @param string $serve_method One of the {@see LLMS_Media_Protector::SERVE_X} constants.
```

includes/class-llms-media-protector.php
pondermatic: I debated about the cumbersome way to pass the filter hook name vs. keeping the `LLMS_Media_Protector->handle_upload()` signature identical to `media_handle_upload()`. I was hoping that the documentation would be clear enough, but see that an extra parameter would be clearer.

I'll add the extra parameter.

thomasplevy: I figured you probably thought about this and intended to keep the signature's identical but... I don't think it *needs* to be identical, just compatible

pondermatic: For the social learning add-on, the original handler looked like this.
```php
$id = media_handle_upload(
	'image',
	0,
	array(
		'post_title' => llms_sl_get_student_image_title( $type, $student ),
	)
);
```
Without the extra parameter, it would look like this.
```php
$id = $media_protector->handle_upload(
	'image',
	0,
	array(
		'meta_input' => array( LLMS_Media_Protector::AUTHORIZATION_FILTER_KEY => 'llms_sl_authorize_media_view' ),
		'post_title' => llms_sl_get_student_image_title( $type, $student ),
	) );
```
With the extra parameter:
```php
$id = $media_protector->handle_upload(
	'image',
	0,
	'llms_sl_authorize_media_view',
	array(
		'post_title' => llms_sl_get_student_image_title( $type, $student ),
	) );
```
Having the extra parameter makes it easier to read and understand.

includes/class-llms-media-protector.php
pondermatic: Yes. In my dev environment, uploaded files are saved to `wp-content/uploads/llms-uploads/social-learning/2021/11/`.

includes/class-llms-media-protector.php
pondermatic: Thank you! I wasn't aware of that function.

includes/class-llms-media-protector.php
pondermatic: @thomasplevy, 

I like the idea of optionally setting the upload directory in the constructor.

Currently, the class is instantiated by `LifterLMS->init()` and the constructor sets up hooks. Perhaps the hooks should be setup in a new method that `LifterLMS->init()` calls. Then the constructor could be used to set the upload directory. Although it looks like setting up hooks multiple times wouldn't hurt (other than minor waste of time).

Another option would be to split the class into one that handles uploads and another that handles authorization, URLs and file serving.

Options:
1. Add setting of the upload directory to the constructor that also sets up hooks.
2. Add setting of the upload directory to the constructor and add a new method that sets up hooks.
3. Split the class into two.

includes/class-llms-media-protector.php
pondermatic: [401 Unauthorized](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) is supposed to be sent with a [WWW-Authenticate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate) header to allow the user to authenticate in the little web browser pop up window. I'm guessing that if the user is not logged in, we should send them to the log in page instead?

thomasplevy: I'm going to say I wasn't aware of this but I guess I am since I wrote this https://github.com/gocodebox/lifterlms-rest/blob/9215df26a9a589d5a5488f019339dd83c1b3c15d/includes/class-llms-rest-authentication.php#L274-L291

I don't believe WordPress's default authentication layer is going to work properly with Basic Auth in this case. What I've just read (briefly) recommends using 403 in the kind of scenario we're in. I don't feel like that's the perfectly valid way to do it but I also read that performing a redirect isn't the proper way to do it either. So... /shrug

pondermatic: After more research, it seems that 403 is the proper response code to use. 401 is meant to ask the user for credentials. We don't want to display a login page while displaying an element on the requested page. Basic authentication isn't normally used on WordPress sites.

includes/class-llms-media-protector.php
pondermatic: I wanted the `@see` tag to link to the default serve method constant, which has more information. `SERVE_X` doesn't link. How about this?
```suggestion
		 * @param string $serve_method One of the LLMS_Media_Protector::SERVE_X constants, {@see LLMS_Media_Protector::SERVE_REDIRECT}.
```

Another option would be to use a `@see` link to each of the serve constants.

The format for `@see` tags is `@see URL label`. It looks great when rendered, but not so great reading the raw text.

thomasplevy: I don't really care... I thought you had made a mistake with the initial version

includes/class-llms-media-protector.php
thomasplevy: It might make sense to split into two different classes, especially if someone wants to perform an upload later they'll need to create a new instance (which sets up the hooks again).

Although it might be better to keep it as it is now but just expose the instance via a class variable. Then `LLMS_Assets()` class works similarly to how I think the media class could work. In this scenario we could probably leave it more or less how it is now without much change and establish a pattern that add-ons could create their own instance within the plugin, setup on init, and then access itself later... Maybe this is what you were going for and then my only concern would be not requiring a new instance every time.

pondermatic: I'm leaning away from splitting the class up into a upload class and a download class because most (92%) of the code is for downloading. There are only 3 small methods used to set the authorization filter (per your suggestion above) and change the upload directory.

I had not considered multiple files being uploaded in one request. If different add-ons need to upload at the same time (is this likely?), this would currently require multiple instances of the media protector class and some minor time wasted re-registering hooks. I've implemented option 2 to solve this problem. Option 4 would be to use a singleton pattern.

pondermatic:
@thomasplevy, I've made a bunch of changes to this WIP based on the above code review.

eri-trabiccolo:
@thomasplevy @pondermatic 
haven't really looked at the PR logic, but I just want to highlight something about the type hinting:
1) I love it :D 
2) We should be careful about using it for hook callbacks because if a 3rd party passes a wrong type along, the fatal will happen in OUR code. We can then blame others, but still we will be blamed first by our customers.

What do you think?

includes/class-llms-media-protector.php
```diff
@@ -0,0 +1,754 @@
+<?php
+/**
+ * LLMS_Media_Protector class
+ *
+ * @package LifterLMS/Classes
+ *
+ * @since   [version]
+ * @version [version]
```
eri-trabiccolo:
```suggestion
 * @since [version]
 * @version [version]
```

includes/class-llms-media-protector.php
```diff
@@ -0,0 +1,754 @@
+<?php
+/**
+ * LLMS_Media_Protector class
```
eri-trabiccolo, line 3:
```suggestion
 * LLMS_Media_Protector class.
```

includes/class-llms-media-protector.php
```diff
@@ -0,0 +1,754 @@
...
+ *
+ * @package LifterLMS/Classes
+ *
+ * @since   [version]
+ * @version [version]
+ */
+
+/**
```
eri-trabiccolo, line 13:
```suggestion

defined( 'ABSPATH' ) || exit;

/**
```

pondermatic:
@eri-trabiccolo wrote: 
> We should be careful about using it (parameter type declaration) for hook callbacks because if a 3rd party passes a wrong type along, the fatal will happen in OUR code.

[Scalar](https://www.php.net/manual/en/language.types.intro.php) values (bool, int, float, string) passed to a parameter declared with a different scalar type will be [coerced](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict) into the declared type.

If the script that calls a function in our script has [strict mode](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict) enabled, or the passed type or declared type are not scalar (array, object, etc.), a [TypeError](https://www.php.net/manual/en/class.typeerror.php) is thrown that will mention our code **and** the caller's code, plus a stack trace.
```
PHP Fatal error:  Uncaught TypeError: LLMS_Media_Protector::authorize_media_image_src(): Argument #1 ($image) must be of type int, array given,
called in wp-includes/class-wp-hook.php on line 303 and defined in wp-content/plugins/lifterlms/includes/class-llms-media-protector.php:211
Stack trace:
#0 wp-includes/class-wp-hook.php(303): LLMS_Media_Protector->authorize_media_image_src(Array, 1136, 'thumbnail', false)
#1 wp-includes/plugin.php(189): WP_Hook->apply_filters(Array, Array)
#2 wp-includes/media.php(991): apply_filters('wp_get_attachme...', Array, 1136, 'thumbnail', false)
#3 wp-content/themes/astra/inc/blog/blog-config.php(360): wp_get_attachment_image_src(1136)
...
``` 

I put the parameter [type declarations](https://www.php.net/manual/en/language.types.declarations.php) in there to explore their usefulness but am prepared to take them out if we think it'll cause problems.

Because of the lack of the union types, e.g. `int|WP_Error` or `mixed`, in PHP 7, I'm thinking we:
1. Wait until PHP 8 is our minimum requirement and then we start adding parameter and return type declarations.
2. We don't declare types for a value that uses multiple types with at least one type being non-scalar.

I like that parameter type declarations remove the need for a method to manually check the type or coerce it into another type.

If a 3rd party **always** calls one of our public methods with the wrong parameter types, they'll find out quickly during their development. I'm concerned that if a caller sometimes passes something like `'1'` instead of `1` because they are relying on PHP's [type juggling](https://www.php.net/manual/en/language.types.type-juggling.php), then the problem won't be noticed until it is released into a production environment and someone will blame us for being too strict.

@thomasplevy, what do you think about type declarations? Is this the right time to start using them?

includes/class-llms-media-protector.php
pondermatic: The [WordPress](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#6-file-headers) and [LifterLMS](https://github.com/gocodebox/lifterlms/blob/trunk/docs/documentation-standards.md#file-headers-1) coding standard document say that the summary line in **file** level DocBlocks should not end with a period.
I don't understand why, because it is inconsistent.

eri-trabiccolo:
> @eri-trabiccolo wrote:
> 
> > We should be careful about using it (parameter type declaration) for hook callbacks because if a 3rd party passes a wrong type along, the fatal will happen in OUR code.
> 
> [Scalar](https://www.php.net/manual/en/language.types.intro.php) values (bool, int, float, string) passed to a parameter declared with a different scalar type will be [coerced](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict) into the declared type.

Well, that's partially true:
http://sandbox.onlinephpfunctions.com/code/ccbe0c1f3a60a6179cc37f8d32f35f44da1a795b

Although I think that you're right, in most of the cases we'll get a `string` representing a number where we'd want an `int` and PHP will just cast it.

> If the script that calls a function in our script has [strict mode](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict) enabled, or the passed type or declared type are not scalar (array, object, etc.), a [TypeError](https://www.php.net/manual/en/class.typeerror.php) is thrown that will mention our code **and** the caller's code, plus a stack trace.
> 
> ```
> PHP Fatal error:  Uncaught TypeError: LLMS_Media_Protector::authorize_media_image_src(): Argument #1 ($image) must be of type int, array given,
> called in wp-includes/class-wp-hook.php on line 303 and defined in wp-content/plugins/lifterlms/includes/class-llms-media-protector.php:211
> Stack trace:
> #0 wp-includes/class-wp-hook.php(303): LLMS_Media_Protector->authorize_media_image_src(Array, 1136, 'thumbnail', false)
> #1 wp-includes/plugin.php(189): WP_Hook->apply_filters(Array, Array)
> #2 wp-includes/media.php(991): apply_filters('wp_get_attachme...', Array, 1136, 'thumbnail', false)
> #3 wp-content/themes/astra/inc/blog/blog-config.php(360): wp_get_attachment_image_src(1136)
> ...
> ```

Yeah, and I also assume that this 3rd party script, since it's developed by someone who cares about types, will take care of respecting them :D 

> I put the parameter [type declarations](https://www.php.net/manual/en/language.types.declarations.php) in there to explore their usefulness but am prepared to take them out if we think it'll cause problems.

I love them, I think there's a misunderstanding here :D I said - or I meant to say that - we should be careful ONLY for which regards to hooks callbacks. I don't care about all our public methods - who uses them should pass the expected type - I'm talking about wp core filters mostly.

> Because of the lack of the union types, e.g. `int|WP_Error` or `mixed`, in PHP 7, I'm thinking we:
> 
> 1. Wait until PHP 8 is our minimum requirement and then we start adding parameter and return type declarations.
> 2. We don't declare types for a value that uses multiple types with at least one type being non-scalar.

I'm for 2.

> I like that parameter type declarations remove the need for a method to manually check the type or coerce it into another type.

I agree, but I still think this is just something we cannot took for granted when we talk about WP hook callbacks.

> If a 3rd party **always** calls one of our public methods with the wrong parameter types, they'll find out quickly during their development. I'm concerned that if a caller sometimes passes something like `'1'` instead of `1` because they are relying on PHP's [type juggling](https://www.php.net/manual/en/language.types.type-juggling.php), then the problem won't be noticed until it is released into a production environment and someone will blame us for being too strict.

I agree, I don't care about 3rd parties calling our public methods with the wrong parameter type. See above.

> @thomasplevy, what do you think about type declarations? Is this the right time to start using them?

includes/class-llms-media-protector.php
eri-trabiccolo: Jeez you're right, ok then, sorry.

pondermatic:
@pondermatic wrote:
> > [Scalar](https://www.php.net/manual/en/language.types.intro.php) values (bool, int, float, string) passed to a parameter declared with a different scalar type will be [coerced](https://www.php.net/manual/en/language.types.declarations.php#language.types.declarations.strict) into the declared type.

@eri-trabiccolo wrote:
> Well, that's partially true:
> http://sandbox.onlinephpfunctions.com/code/ccbe0c1f3a60a6179cc37f8d32f35f44da1a795b

OMG that's right! Not all types are coerced. In fact, it seems strange that some are and some are not. I wrote a script to test all the combinations and am planning on making a blog post about it (off-the-clock).

You have good points @eri-trabiccolo.
1. It is OK that anyone calling our methods must obey the type declaration.
2. We should not have declared types on hook callbacks because we can't trust that other callbacks using the same hook will pass the correct type. We can't even be sure that a WordPress hook will always pass the correct type.

pondermatic:
@gocodebox/engineering 
Blog post about parameter type declarations and what happens when the wrong type is passed in.
https://www.pondermatic.com/2021/12/06/php-parameter-type-declarations/

thomasplevy:
I'm in favor of type declarations but I don't like adding them with a bunch of exceptions (mainly there are no union types)... I don't like the idea of only declaring some types, it just feels sloppy... But I guess in reality it's *less sloppy* than not declaring any.

I *know* that if we start declaring them on callbacks for hooks *we will* run into issues. Let's be honest that most people using hooks don't actually know how to write PHP... We already have to explain code snippets we share (to developers) in far more detail that seems reasonable... if we also have to add explaining type declarations... well... you know...?

2. We don't declare types for a value that uses multiple types with at least one type being non-scalar.

And not declaring types on functions/methods we internally use as a hook callback.

PR #1871, opened 2021-12-01
Enable the block editor for certificates

thomasplevy:
Adds blocks editor functionality for certificate creation / editing

Requires #1872 which I moved to another branch to reduce PR size


eri-trabiccolo:
well you merged #1872 into this :D,
<3 u anyways, but a bit less.
(I didn't review it)

templates/certificates/dynamic-styles.php
```diff
@@ -0,0 +1,42 @@
...
+
+defined( 'ABSPATH' ) || exit;
+?>
+<style type="text/css">
+	html, body {
+		background-color: <?php echo $background_color; ?>;
+	}
+	.llms-certificate-container {
+		background-image: <?php echo "url( {$background_img} )"; ?>;
+		height: <?php echo $height; ?>;
```
eri-trabiccolo, line 28:
```suggestion
		background-image: <?php echo "url( {$background_img} )"; ?>;
		background-size: cover;
		height: <?php echo $height; ?>;
```

thomasplevy: I messed up and checked out the SCSS changes I made with this PR... I've updated. `cover` doesn't seem appropriate to me in this scenario. We are defining explicit widths for the users and they can choose to upload something that is expected to be that size. Just a bit of testing with `cover` results in default certificates being cut and not displayed well so I think using `100% 100%` (and stretching it a bit) is preferable in this case.

PR #1873, opened 2021-12-02
Prevent concurrent logins

eri-trabiccolo:
## Description
Prevent concurrent logins

## How has this been tested?
manually
unit tests wip

## Types of changes

## Checklist:

### DIFF
--- a/includes/admin/settings/class.llms.settings.accounts.php
+++ b/includes/admin/settings/class.llms.settings.accounts.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Admin/Settings/Classes
  *
  * @since 1.0.0
- * @version 5.0.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -43,13 +43,13 @@ class LLMS_Settings_Accounts extends LLMS_Settings_Page {
 	 * @since 3.30.3 Fixed spelling errors.
 	 * @since 3.37.3 Renamed duplicate field id for section close (`user_info_field_options` to `user_info_field_options_end`)
 	 * @since 5.0.0 Removed field display settings.
-	 *               Reorganized open registration setting.
-	 *               Renamed "User Information Options" to "User Privacy Options".
+	 *              Reorganized open registration setting.
+	 *              Renamed "User Information Options" to "User Privacy Options".
+	 * @since [version] Added options to disable concurrent logins.
 	 *
 	 * @return array
 	 */
 	public function get_settings() {
-
 		return apply_filters(
 			'lifterlms_' . $this->id . '_settings',
 			array(
@@ -101,6 +101,29 @@ class LLMS_Settings_Accounts extends LLMS_Settings_Page {
 					'id'      => 'lifterlms_enable_myaccount_registration',
 					'title'   => __( 'Open Registration', 'lifterlms' ),
 					'type'    => 'checkbox',
+				),
+				array(
+					'default'           => 'no',
+					'desc'              => __( 'This will ensure that two people cannot be logged in to the same account at once.', 'lifterlms' ),
+					'id'                => 'lifterlms_prevent_concurrent_logins',
+					'title'             => __( 'Prevent concurrent logins', 'lifterlms' ),
+					'type'              => 'checkbox',
+					'custom_attributes' => array(
+						'class'         => 'llms-conditional-controller',
+						'data-controls' => '#lifterlms_prevent_concurrent_logins_roles',
+					),
+				),
+				array(
+					'class'             => 'llms-select2',
+					'default'           => array( 'student' ),
+					'desc'              => __( 'Users with the selected roles won\'t be able to be logged in to the same account from different places.', 'lifterlms' ),
+					'id'                => 'lifterlms_prevent_concurrent_logins_roles',
+					'options'           => LLMS_Roles::get_all_role_names(),
+					'title'             => __( 'Prevent concurrent logins to', 'lifterlms' ),
+					'type'              => 'multiselect',
+					'custom_attributes' => array(
+						'data-placeholder' => __( 'Select user roles', 'lifterlms' ),
+					),
 				),
 				array(
 					'id'   => 'course_account_options_end',
--- a/includes/admin/settings/class.llms.settings.accounts.php
+++ b/includes/admin/settings/class.llms.settings.accounts.php
@@ -104,7 +104,7 @@ class LLMS_Settings_Accounts extends LLMS_Settings_Page {
 				),
 				array(
 					'default'           => 'no',
-					'desc'              => __( 'This will ensure that two people cannot be logged in to the same account at once.', 'lifterlms' ),
+					'desc'              => __( 'Discourages account sharing by only allowing the most recent session for each user account.', 'lifterlms' ),
 					'id'                => 'lifterlms_prevent_concurrent_logins',
 					'title'             => __( 'Prevent concurrent logins', 'lifterlms' ),
 					'type'              => 'checkbox',
@@ -116,10 +116,10 @@ class LLMS_Settings_Accounts extends LLMS_Settings_Page {
 				array(
 					'class'             => 'llms-select2',
 					'default'           => array( 'student' ),
-					'desc'              => __( 'Users with the selected roles won\'t be able to be logged in to the same account from different places.', 'lifterlms' ),
+					'desc'              => __( "Users with the selected roles won't be able to be logged in to the same account from different places.", 'lifterlms' ),
 					'id'                => 'lifterlms_prevent_concurrent_logins_roles',
 					'options'           => LLMS_Roles::get_all_role_names(),
-					'title'             => __( 'Prevent concurrent logins to', 'lifterlms' ),
+					'title'             => __( 'Prevent concurrent logins for these roles', 'lifterlms' ),
 					'type'              => 'multiselect',
 					'custom_attributes' => array(
 						'data-placeholder' => __( 'Select user roles', 'lifterlms' ),
--- a/includes/admin/settings/class.llms.settings.accounts.php
+++ b/includes/admin/settings/class.llms.settings.accounts.php
@@ -104,7 +104,7 @@ class LLMS_Settings_Accounts extends LLMS_Settings_Page {
 				),
 				array(
 					'default'           => 'no',
-					'desc'              => __( 'Discourages account sharing by only allowing the most recent session for each user account.', 'lifterlms' ),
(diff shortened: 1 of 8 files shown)


includes/admin/settings/class.llms.settings.accounts.php
```diff
@@ -102,6 +102,29 @@ public function get_settings() {
 					'title'   => __( 'Open Registration', 'lifterlms' ),
 					'type'    => 'checkbox',
 				),
+				array(
+					'default'           => 'no',
+					'desc'              => __( 'This will ensure that two people cannot be logged in to the same account at once.', 'lifterlms' ),
```
thomasplevy:
```suggestion
					'desc'              => __( 'Discourages account sharing by only allowing the most recent session for each user account.', 'lifterlms' ),
```

includes/admin/settings/class.llms.settings.accounts.php
```diff
@@ -102,6 +102,29 @@ public function get_settings() {
...
+					'type'              => 'checkbox',
+					'custom_attributes' => array(
+						'class'         => 'llms-conditional-controller',
+						'data-controls' => '#lifterlms_prevent_concurrent_logins_roles',
+					),
+				),
+				array(
+					'class'             => 'llms-select2',
+					'default'           => array( 'student' ),
+					'desc'              => __( 'Users with the selected roles won\'t be able to be logged in to the same account from different places.', 'lifterlms' ),
```
thomasplevy:
```suggestion
					'desc'              => __( "Users with the selected roles won't be able to be logged in to the same account from different places.", 'lifterlms' ),
```

includes/admin/settings/class.llms.settings.accounts.php
```diff
@@ -102,6 +102,29 @@ public function get_settings() {
...
+					'id'                => 'lifterlms_prevent_concurrent_logins_roles',
+					'options'           => LLMS_Roles::get_all_role_names(),
+					'title'             => __( 'Prevent concurrent logins to', 'lifterlms' ),
```
thomasplevy:
```suggestion
					'title'             => __( 'Prevent concurrent logins for these roles', 'lifterlms' ),
```

includes/class-llms-prevent-concurrent-logins.php
```diff
@@ -0,0 +1,165 @@
...
+	/**
+	 * Private Constructor.
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	private function __construct() {
+
+		if ( get_option( 'lifterlms_prevent_concurrent_logins', false ) &&
```
thomasplevy: The option stores yes/no values which are handled by `llms_parse_bool()` to true/false.
```suggestion
		if ( llms_parse_bool( get_option( 'lifterlms_prevent_concurrent_logins', 'no' ) ) &&
```

includes/class-llms-prevent-concurrent-logins.php
```diff
@@ -0,0 +1,165 @@
...
+
+		// No logged in user or current user has only one active session: nothing to do.
+		if ( empty( $this->user_sessions ) || count( $this->user_sessions ) < 2 ) {
+			return false;
+		}
+
+		// Current user doesn't have any restricted role: nothing to do.
+		if ( empty( array_intersect( get_userdata( $this->user_id )->roles, get_option( 'lifterlms_prevent_concurrent_logins_roles', array() ) ) ) ) {
+			return false;
+		}
```
thomasplevy:
```suggestion
```

Duplicate code from below?

includes/class-llms-prevent-concurrent-logins.php
```diff
@@ -0,0 +1,165 @@
...
+		 *
+		 * @param bool $allow   Whether or not the user should be allowed to have concurrent sessions.
+		 * @param int  $user_id WP_User ID of the current use.
+		 */
+		if ( (bool) apply_filters( 'llms_allow_user_concurrent_logins', false, $this->user_id ) ) {
+			return false;
+		}
+
+		// Current user doesn't have any restricted role: nothing to do.
+		if ( empty( array_intersect( get_userdata( $this->user_id )->roles, get_option( 'lifterlms_prevent_concurrent_logins_roles', array() ) ) ) ) {
```
thomasplevy: Just in case... I've seen people empty options and create fatal errors with code like this (that we then have to deal with)... Something I remember with admin notifications a while back, for example...

```suggestion
		if ( empty( array_intersect( get_userdata( $this->user_id )->roles, (array) get_option( 'lifterlms_prevent_concurrent_logins_roles', array() ) ) ) ) {
```

includes/class-llms-prevent-concurrent-logins.php
```diff
@@ -0,0 +1,165 @@
...
+		return $this->destroy_all_sessions_but_newest();
+
+	}
+
+	/**
+	 * Prevent login by destroying all the user's sessions but the newest.
+	 *
+	 * @since [version]
+	 *
+	 * @return int 1 if the kept session is the current one, 0 otherwise.
```
thomasplevy, line 115: Why return an int here instead of a bool? The function calling this signature is wrong if this isn't changed to a bool.

includes/class-llms-prevent-concurrent-logins.php
```diff
@@ -0,0 +1,165 @@
...
+	/**
+	 * Retrieve newest session login time for the current user.
+	 *
+	 * The bigger the login time is the newest the session is.
+	 *
+	 * @since [version]
+	 *
+	 * @return int
+	 */
+	private function current_user_newest_session_loin_time() {
```
thomasplevy:
```suggestion
	private function current_user_newest_session_login_time() {
```

includes/class-llms-prevent-concurrent-logins.php
```diff
@@ -0,0 +1,165 @@
...
+	/**
+	 * Prevent login by destroying all the user's sessions but the newest.
+	 *
+	 * @since [version]
+	 *
+	 * @return int 1 if the kept session is the current one, 0 otherwise.
+	 */
+	private function destroy_all_sessions_but_newest() {
+
+		$is_current_session_newest_session = $this->current_user_current_session_login_time() === $this->current_user_current_session_login_time();
```
thomasplevy:
```suggestion
		$is_current_session_newest_session = $this->current_user_newest_session_login_time() === $this->current_user_current_session_login_time();
```

thomasplevy (changes requested):

includes/class-llms-prevent-concurrent-logins.php
eri-trabiccolo: Yeah stupid me

includes/class-llms-prevent-concurrent-logins.php
eri-trabiccolo: Yeah..

includes/class-llms-prevent-concurrent-logins.php
eri-trabiccolo: I want this function to return an int because 'semantically' a false or true to me means something went fine or not.
But in this case we know that the "keep only one session" action cannot fail in this method.
The return int type is mostly for tests.
The method who's calling this actually should return just true.

thomasplevy: :+1:

tests/phpunit/unit-tests/class-llms-test-prevent-concurrent-logins.php
```diff
@@ -0,0 +1,322 @@
...
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function tear_down() {
+		parent::tear_down();
+		wp_set_current_user(0);
+	}
+
```
thomasplevy: This is redundant: https://github.com/WordPress/wordpress-develop/blob/281142bfb4c913e1945d0d014b29e334cf137e55/tests/phpunit/includes/abstract-testcase.php#L189

```suggestion
```

includes/admin/settings/class.llms.settings.accounts.php
```diff
@@ -102,6 +102,29 @@ public function get_settings() {
 					'title'   => __( 'Open Registration', 'lifterlms' ),
 					'type'    => 'checkbox',
 				),
+				array(
+					'default'           => 'no',
+					'desc'              => __( 'Discourages account sharing by only allowing the most recent session for each user account.', 'lifterlms' ),
```
thomasplevy:
```suggestion
					'desc'              => __( 'Only allow the most recent login for each user account.', 'lifterlms' ),
```

includes/admin/settings/class.llms.settings.accounts.php
```diff
@@ -102,6 +102,29 @@ public function get_settings() {
...
+						'data-controls' => '#lifterlms_prevent_concurrent_logins_roles',
+					),
+				),
+				array(
+					'class'             => 'llms-select2',
+					'default'           => array( 'student' ),
+					'desc'              => __( "Users with the selected roles won't be able to be logged in to the same account from different places.", 'lifterlms' ),
+					'id'                => 'lifterlms_prevent_concurrent_logins_roles',
+					'options'           => LLMS_Roles::get_all_role_names(),
+					'title'             => __( 'Prevent concurrent logins for these roles', 'lifterlms' ),
```
thomasplevy:
```suggestion
					'title'             => '',
```
Since this will display below the login option as a second option for the above option I think it looks better in context with the title removed so it's kind of obviously part of the above option:

Screenshot_2021-12-03_09-12-10

What do you think?

includes/admin/settings/class.llms.settings.accounts.php
```diff
@@ -102,6 +102,29 @@ public function get_settings() {
...
+					'type'              => 'checkbox',
+					'custom_attributes' => array(
+						'class'         => 'llms-conditional-controller',
+						'data-controls' => '#lifterlms_prevent_concurrent_logins_roles',
+					),
+				),
+				array(
+					'class'             => 'llms-select2',
+					'default'           => array( 'student' ),
+					'desc'              => __( "Users with the selected roles won't be able to be logged in to the same account from different places.", 'lifterlms' ),
```
thomasplevy:
```suggestion
					'desc'              => __( 'Prevent concurrent logins for users with the selected user roles.', 'lifterlms' ),
```

thomasplevy (commented):
Just ran it to do some manual checks before approving and decided I wanted to adjust the language on the settings a little bit again... what do you think of these changes?

includes/admin/settings/class.llms.settings.accounts.php
eri-trabiccolo: I think you're right.

PR #1883, opened 2021-12-13
Post tables for my_cert and my_achievments

thomasplevy:


includes/admin/post-types/class.llms.post.tables.php
```diff
@@ -159,13 +160,25 @@ public function handle_link_actions() {
...
 		<span class="llms-post-table-post-filter">
 			<label for="<?php printf( 'filter-by-llms-post-%s', $post_type ); ?>" class="screen-reader-text">
 				<?php printf( esc_html__( 'Filter by %s', 'lifterlms' ), $label ); ?>
 			</label>
-			<select class="llms-select2-post" data-allow_clear="true" data-placeholder="<?php printf( esc_attr__( 'Filter by %s', 'lifterlms' ), $label ); ?>" data-post-type="<?php echo $post_type; ?>" name="<?php echo $name; ?>" id="<?php printf( 'filter-by-llms-post-%s', $post_type ); ?>">
+			<select
+				class="llms-select2-post"
+				data-allow_clear="true"
+				data-no-view-button="true"
+				data-placeholder="<?php printf( esc_attr__( 'Filter by %s', 'lifterlms' ), $label ); ?>"
```
eri-trabiccolo: missing translators suggestion

eri-trabiccolo (changes requested):
I think we can remove this too:
https://github.com/gocodebox/lifterlms/blob/dev-600/includes/admin/post-types/meta-boxes/class.llms.meta.box.award.engagement.submit.php#L72
https://github.com/gocodebox/lifterlms/blob/dev-600/includes/admin/post-types/meta-boxes/class.llms.meta.box.award.engagement.submit.php#L223-L242
?
Also:
1) the Date column (in the post list table) should read Awarded in place of Published for the awarded certificates & achievements.
2) the certificate template migration doesn't seem to work? From the awarded certificates post list table I click on the Migrate Template action link and I'm taken to the certificate template block editor, I save, but when I edit again the template (or the awarded certificate) it still shows me the classic editor (and the label next to the awarded certificate still reads Legacy Template).

thomasplevy:
> 2\. the certificate template migration doesn't seem to work? From the awarded certificates post list table I click on the Migrate Template action link and I'm taken to the certificate template block editor, I save, but when I edit again the template (or the awarded certificate) it still shows me the classic editor (and the label next to the awarded certificate still reads Legacy Template).

This is *expected* based on how the code is written, but seeing your confusion I'm wondering if we should adjust the behavior... 

See the logic for determining if template v1 (legacy) or v2 (block ed) is being used: https://github.com/gocodebox/lifterlms/blob/499699c4cfbd3a9efbf634a2762017b90e149778/includes/models/model.llms.user.certificate.php#L533-L549

When "migrating" to v2, it forces the block editor to be opened. The assumption I'm making is that the user would convert the classic block to blocks before saving. But this is probably a dumb assumption. Based on the way the classic block works *no blocks are actually saved* when the classic block is used. So unless at least one block is added (or the classic is converted to blocks) it *will* look as if the migration failed. 

I wrote it this way as to not require saving additional metadata to determine the template version. I also thought that by writing it this way it would make it possible to "revert" back to the legacy template if a user decided they didn't like the new system and they want to stick on the legacy template. My thinking here was that we could continue supporting classic until classic completely goes away from the core and we could have a "secret" reversion method. Just delete your blocks and use the classic block or create a new certificate and insert only the classic block then save and exit and reopen the editor...

But anyway... It does seem likely that someone will do what you did and then say it didn't work. So...

A) Do we keep it as is and have documentation explaining this is how it works
B) Save metadata and switch the version check function to read metadata
C) Add some kind of notice in the block editor that alerts the user that they'll be reverted to the legacy template unless they remove the classic block?
D) When loading up the editor during a migration we could recreate the onClick event of the classic block and "force" the classic editor into blocks on load (see https://github.com/WordPress/gutenberg/blob/f3c9ed0e0d7d71469dca42f766fd3073d8fc54f1/packages/block-library/src/freeform/convert-to-blocks-button.js#L21-L27)

I feel like D is probably a bit of work but will probably be the most successful solution.

If we go with A users are confused and probably frustrtaed

If we go with C it's probably as much work to code as D and is more foolproof for users

B impacts performance negatively and we'll lose the ability to revert without also building an interface for updating the metadata

Thoughts @eri-trabiccolo ?

eri-trabiccolo:
@thomasplevy 
Yeah C) + doc makes sense.

PR #1884, opened 2021-12-13
Version 6.0.0 Working Branch

### LINKED ISSUE #290: delayed engagements may still fire after engagement has been deleted

### LINKED ISSUE #463: Certificate Enhancements
- [x] Add associated post's merge codes to certificates (and emails, achievements?) (see #567)
- [x] More merge codes
- [x] #1907
- [x] Export directly to PDF
- [x] Update template to exclude theme header / footer content
- [x] Live Merge Code Preview & Testing (see #561)
- [x] Certificate sizing options: Allow users to specify width/height of certificate background (overriding global settings) as well as defining the margin and padding.

### LINKED ISSUE #535: Blank account endpoint for achievements needs to remove dashboard main page summary section in addition to nav items

### LINKED ISSUE #669: "My Achievements" student dashboard endpoint should be paginated
This page currently displays all earned achievements which could be problematic on large sites with potentially hundreds of earned achievements.

This page should implement pagination and should display a reasonable number of achievements per page. Since the default loop is 4 achievements the default number of achievements should be a multiple of 4. 16 or 20 seem like reasonable defaults (4 or 5 rows of 4 achievements each)

Pagination and other aspects of this page should also be filterable to allow 3rd party / dev customization via hooks

### LINKED ISSUE #859: Fix spelling error in LLMS_Database_Query abstract method `preprare_query()`

### LINKED ISSUE #1027: Add utility function for escaping and quoting a string.

### LINKED ISSUE #1078: Add "Sync" options when updating an engagement (cert, achievement)
When an admin edits the template for a cert or achievement (collectively "engagement") there should be an option to "sync" the changes to all pre-existing engagements which were generated from that template.

- [ ] Add a "Sync" button to the admin edit post UI
- [ ] When the button is clicked it should perform an "update" on earned engagements generated from the template.
- [ ] We should spawn a background process to handle the regeneration of the engagements instead of doing this synchronously

## Update process

We do not currently have an "update" method for earned engagements. One should be created. This method should function largely like the creation method except:

it should *not* trigger notifications
it should update the post content, meta, and title instead of creating a new one (the ID should stay the same)

### LINKED ISSUE #1081: Update default images

### LINKED ISSUE #1082: Reduce code redundancy of code shared between certificates and achievements

### LINKED ISSUE #1640: Certificate {first_name} merge code fallback to user nickname is confusing

### LINKED ISSUE #1886: [6.x] Remove (most) previously deprecated code and record breaking changes
In preparation for the 6.0 release we might as well remove previously deprecated code.

@pondermatic can you start working on removing deprecated code and logging the removed items as changelogs on a branch to merge into dev-600.

thomasplevy:
+ Fixes #290
+ Fixes #535
+ Fixes #669 
+ Fixes #859
+ Closes #463 
+ Closes #1027
+ Closes #1078
+ Closes #1081
+ Closes #1082
+ Fixes #1640 
+ Closes #1886


thomasplevy:
@gocodebox/engineering when doing code review in favor of leaving suggestions please update the code on a new branch and open a PR -- especially with documentation fixes. This will expedite development!

PR #1887, opened 2021-12-16
Fix award engagement submit mb appearing in llms_my_certificates with…

eri-trabiccolo:
… block editor


thomasplevy:
this is fancy

PR #1899, opened 2021-12-21
Cert title block

### LINKED ISSUE #1888: Dev-600: Certificate(s) title(s)
Can we clarify the role of the titles in the certificate template and awarded certificate?
Certificate template:
Schermata 2021-12-16 alle 19 07 31
What's the role of the Title under Settings?

Awarded certificate:
Schermata 2021-12-16 alle 19 05 04
Shouldn't "Add Title" be hidden there? As the real title would be the Certificate Title?

thomasplevy:
## Description

Adds the llms/certificate-title block

This block aims to reduce confusion around the certificate title across both cert post types (Fixes #1888)

The block is a wrapper around the core/heading block and is conditionally tied to the `certificate_title` meta field when editing an `llms_certificate` and tied to the `title` post field when editing an `llms_my_certificate`.

The meta document control is conditionally visible on the `llms_certificate` post type depending on the presence of the block. Similarly, the main post title field is conditionally visible on the `llms_my_certificate` post type.

This additionally introduces the ability to customize the font of the title, including 4 google web fonts (2 old-english style fonts and 2 script fonts).

We will have to update the certificate export functionality to allow google web fonts. Alternatively we could include the font kits into the plugin directly for these fonts to ensure exported certs can be fully offline. I'll need to investigate how webfonts will wor via PDFLayer. I'm not addressing these issues in this PR. Just noting that I am aware of them.

## How has this been tested?
+ New & existing tests
+ Manually

## Types of changes
+ New feature

## Checklist:


includes/admin/class.llms.admin.assets.php
```diff
@@ -47,6 +48,17 @@ public function admin_print_footer_scripts() {
 		llms()->assets->output_inline( 'footer' );
 	}
 
+	/**
+	 * Output inline styles in the header
```
eri-trabiccolo:
```suggestion
	 * Output inline styles in the header.
```

includes/class-llms-block-library.php
```diff
@@ -0,0 +1,79 @@
...
+ * @since [version]
+ * @version [version]
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Load the LifterLMS block library.
+ *
+ * @sinc [version]
```
eri-trabiccolo:
```suggestion
 * @since [version]
```

includes/class-llms-block-library.php
```diff
@@ -0,0 +1,79 @@
...
+	/**
+	 * Retrieves a list of blocks to register.
+	 *
+	 * @since [version]
+	 *
+	 * @return string[] A list of directory paths that can individually be passed to `register_block_type()`.
+	 */
+	private function get_blocks() {
+
+		$blocks = array(
```
eri-trabiccolo, line 42: maybe add a filter here?

includes/schemas/llms-block-templates.php
```diff
@@ -0,0 +1,115 @@
+<?php
+/**
+ * Post type block templates.
+ *
+ * Returns an array of post type block types for use in post type registration.
+ *
+ * @package LifterLMS/Schemas
+ *
+ * @since 5.2.0
+ * @version 5.2.0
```
eri-trabiccolo:
```suggestion
 * @since [version]
 * @version [version]
```

tests/phpunit/unit-tests/class-llms-test-block-library.php
```diff
@@ -0,0 +1,58 @@
...
+	 */
+	public function set_up() {
+
+		parent::set_up();
+		$this->main = new LLMS_Block_Library();
+
+	}
+
+	/**
+	 * Test register()
```
eri-trabiccolo:
```suggestion
	 * Test register().
```

tests/phpunit/unit-tests/class-llms-test-post-types.php
```diff
@@ -119,6 +119,28 @@ public function test_deregister_sitemap_post_types() {
 
 	}
 
+	/**
+	 * Test get_template()
```
eri-trabiccolo:
```suggestion
	 * Test get_template().
```

tests/phpunit/unit-tests/functions/class-llms-test-functions-certificates.php
```diff
@@ -84,6 +84,19 @@ public function test_llms_get_certificate_content() {
 
 	}
 
+	/**
+	 * Test llms_get_certificate_fonts()
```
eri-trabiccolo:
```suggestion
	 * Test llms_get_certificate_fonts().
```

eri-trabiccolo (changes requested):
Looks good overall (pretty proud of you <3), although:
1) gfonts are not enqueued on front at the moment:
2) when you create an awarded certificate the default template including merge codes is loaded, although merge codes are not "parsed" - they are not even available in the toolbar when using the classic editor, - so I'm not sure what should be the expected behavior here...

includes/class-llms-block-library.php
thomasplevy: @eri-trabiccolo I originally had a filter on the return but then I thought better of it because, simply, it's basically list of things that can be easily passed into `register_block_type()`. I suppose that there might be some quality of life for us in our add-ons to save a few lines of code though and maybe some developer might want to use it...But... also... who cares? I'll add it back in.

thomasplevy: No... I'm not filtering it. This registers our core blocks, not external blocks. External blocks can be registered on their own and if someone wants to remove a core LLMS block they can do so by deregistering the block using the WP_Block_Registry class.

My reasoning here is that we'd need to create an exception handler for users to define a different namespace for the dupcheck in the `register()` method.

All in all if someone wants to register their own block... the API already exists in WP and they don't need to use our API to do so.

eri-trabiccolo: Mmm I did a mess I think. I meant to add that comment to the list of fonts...

thomasplevy: Here: https://github.com/gocodebox/lifterlms/pull/1899/files#diff-572a859272cff978c2699df82291b4a0605f707085cb2d3bca0a058f30589ccaR162

?

eri-trabiccolo: Yeah 😂

PR #1919, opened 2022-01-04
Print styles

thomasplevy:
Fixes a handful of visual issues found on certificate templates during testing

+ Removes top & bottom margins on blocks found within columns (aligns the front-end with the backend)
+ Did you know percent-based padding uses the width of the parent element (instead of the element itself)? This wraps main certificate <div> in a wrapper and moves the certificate size (width/height) to the wrapper so that the main element (with the padding) will always be the same percentage regardless of the width / height of the screen.
+ Adds "compatibility" code that automatically dequeues all enqueued stylesheets marked as "print" media. This aims to reduce theme conflicts arising when the theme specifies alternate font sizes / etc... in print. We aim to build a direct representation of the certificate on screen and print that exactly.
  + I've tested this compatibility on Astra, OceanWP, Kadence, 2019-2022. 2021 and 2022 have a *few* issues remaining in print view after this change but the others all look nearly identical when printing.
  + The 2021 issues have a different font sizes specified as "only screen" in the main CSS file. This can be addressed via theme compat CSS.
  + 2022 had a bunch of spacing issues (margins and paddings). Probably others. We can address this when we add 2022 support to the core.


thomasplevy:
@pondermatic can I distract you sometime over the next 48-72 hours to have a review of the strategy of dequeuing print styles to improve compatibility. It seems like a good idea to me but I want signoff that I'm not missing something obvious so that at the very least if this was a bad idea we can share the load <3

Thanks!

pondermatic:
@thomasplevy, the dequeuing print styles strategy looks good to me.

PR #1929, opened 2022-01-11
Certs wp version support

thomasplevy:
## Description

This PR implements a version check against the WordPress core version and enables usage of the block editor for certificates for WordPress core 5.8 and later.

The reasoning for this restriction is that there are `@wordpress/*` pacakge apis/functions/etc used to build the block editor functionality that were introduced in version 5.8 of WordPress. I looked into what would be involved into updating the JS for earlier version support and decided that it would not be worth the effort. Additionally, we implemented a check like this on the introduction of forms block editor building (version 5.0) and no complaints about wanting to use the new functionality on an old version came to my attention. That doesn't mean no one complained, but the "Upgrade WP" response to "why can't I use the editor?" will likely be sufficient. I'm hoping.

On WP 5.7 and earlier, users will still be using the classic editor. They won't see the "Legacy" post status flag, nor will they have post action link to upgrade the template. Once upgraded to 5.8 all this functionality will be present and the block editor will become the default.

Additionally, this will add a simple flag for disabling the block editor on 5.8 or later: `add_filter( 'llms_block_editor_supported_for_certificates`, '__return_false' );`

There could be someone unhappy with the block editor that wants to stick on classic... and with this there's a simple way to allow that choice to be made.

## How has this been tested?

+ Manually tested WP 5.7 - 5.9.
+ Modified some unit tests to check for version support (some tests on 5.7 and lower were failing as a result of PHP function signatures that worked on 5.8 or later. These tests are now skipped on 5.7 and earlier)

### DIFF
--- a/includes/admin/post-types/post-tables/class-llms-admin-post-table-certificates.php
+++ b/includes/admin/post-types/post-tables/class-llms-admin-post-table-certificates.php
@@ -38,7 +38,7 @@ class LLMS_Admin_Post_Table_Certificates {
 		}
 
 		$post_types = array( 'llms_certificate', 'llms_my_certificate' );
-		if ( in_array( llms_filter_input( INPUT_GET, 'post_type', FILTER_SANITIZE_FULL_SPECIAL_CHARS ), $post_types, true ) ) {
+		if ( in_array( llms_filter_input( INPUT_GET, 'post_type' ), $post_types, true ) ) {
 			add_filter( 'display_post_states', array( $this, 'add_states' ), 20, 2 );
 			add_filter( 'post_row_actions', array( $this, 'add_actions' ), 20, 2 );
 		}
--- a/includes/admin/post-types/post-tables/class-llms-admin-post-table-certificates.php
+++ b/includes/admin/post-types/post-tables/class-llms-admin-post-table-certificates.php
@@ -33,8 +33,12 @@ class LLMS_Admin_Post_Table_Certificates {
 	 */
 	public function __construct() {
 
+		if ( ! llms_is_block_editor_supported_for_certificates() ) {
+			return;
+		}
+
 		$post_types = array( 'llms_certificate', 'llms_my_certificate' );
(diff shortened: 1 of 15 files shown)


tests/phpunit/unit-tests/functions/class-llms-test-functions-certificates.php
```diff
@@ -173,4 +173,46 @@ public function test_llms_get_certificate_title() {
...
+			array( '5.9-alpha.1', true ),
+			array( '5.9-RC.1', true ),
+			// Future versions?
+			array( '6.0', true ),
+			array( '6.0.1', true ),
+		);
+
+		foreach ( $tests as $test ) {
+			list( $wp_version, $expect ) = $test;
+			$this->assertEquals( $expect, llms_is_block_editor_supported_for_certificates() );
```
eri-trabiccolo:
```suggestion
			$this->assertEquals( $expect, llms_is_block_editor_supported_for_certificates(), $wp_version );
```
so that we know which test fails at a glance.

includes/admin/post-types/post-tables/class-llms-admin-post-table-certificates.php
```diff
@@ -33,8 +33,12 @@ class LLMS_Admin_Post_Table_Certificates {
 	 */
 	public function __construct() {
 
+		if ( ! llms_is_block_editor_supported_for_certificates() ) {
+			return;
+		}
+
 		$post_types = array( 'llms_certificate', 'llms_my_certificate' );
-		if ( in_array( llms_filter_input( INPUT_GET, 'post_type', FILTER_SANITIZE_STRING ), $post_types, true ) ) {
+		if ( in_array( llms_filter_input( INPUT_GET, 'post_type', FILTER_SANITIZE_FULL_SPECIAL_CHARS ), $post_types, true ) ) {
```
eri-trabiccolo:
```suggestion
		if ( in_array( llms_filter_input( INPUT_GET, 'post_type' ), $post_types, true ) ) {
```
I think we're fine with the default filter here, right? (do not filter anything) we're not going to use that value if not for this comparison.

eri-trabiccolo (changes requested):

Issue #1850, opened 2021-11-08
WP 5.9 & PHP 8.1


thomasplevy:
## Timeline

11/23/21 - WP 5.9 Beta.2
11/25/21 - PHP 8.1 General Release
11/30/21 - WP 5.9 RC.1
12/14/21 - WP 5.9 General Release

## Tasks

+ [x] PHP 8.1 fixes for core
+ [x] WP 5.9 fixes for core
+ [ ] PHP 8.1 fixes for add-ons
+ [ ] WP 5.9 fixes for add-ons
+ [x] Twenty Twenty-Two Theme Support, #1824

---

Errors found in CI resulting from WP Core "issues"

https://core.trac.wordpress.org/ticket/54504

```
PHP Deprecated:  Return type of Requests_Cookie_Jar::offsetExists($key) should either be compatible with ArrayAccess::offsetExists(mixed $offset): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /home/thoams/srv/www/llms/htdocs/wp-content/plugins/lifterlms/vendor/rmccue/requests/library/Requests/Cookie/Jar.php on line 63

PHP Deprecated:  Return type of Requests_Cookie_Jar::getIterator() should either be compatible with IteratorAggregate::getIterator(): Traversable, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /home/thoams/srv/www/llms/htdocs/wp-content/plugins/lifterlms/vendor/rmccue/requests/library/Requests/Cookie/Jar.php on line 111

PHP Deprecated:  Return type of Requests_Utility_CaseInsensitiveDictionary::offsetGet($key) should either be compatible with ArrayAccess::offsetGet(mixed $offset): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /home/thoams/srv/www/llms/htdocs/wp-content/plugins/lifterlms/vendor/rmccue/requests/library/Requests/Utility/CaseInsensitiveDictionary.php on line 51

[...]

PR #1948, opened 2022-01-20
Resolve PHP 8.1 Deprecations

thomasplevy:
## Description

Per #1850

The remaining errors (as reported by the CI) are coming from WP Core functions and the ActionScheduler library. THe issues from Action Scheduler are resolved on their main branch but they haven't released an update yet.

## How has this been tested?

+ Manually (need to do more thorough manual testing)
+ New and existing PHPUnit and E2E tests

## Types of changes

+ PHP 8.1 & WP 5.9 compatibility

## Checklist:


thomasplevy:
@gocodebox/engineering this PR is not finished but a lot of the changes we need to make are specific to the usage of `FILTER_SANITIZE_STRING` alongside `llms_filter_input()` (our wrapper for `filter_input()`.

The filter is being deprecated because developers "misunderstood" the filter, but I (I guess we, collectively) have not misunderstood it.

Switching to `FILTER_SANITIZE_FULL_SPECIAL_CHARS` is *not* a direct replacement and could lead to stored XSS injections if we aren't careful when replacing.

So I've proposed a new function `llms_filter_input_sanitize_string()` which will be used to replace `llms_filter_input()` when we desire to strip tags and encode quotes. In this way can continue to use it where we desire. In many instances (where we're not using the result of `llms_filter_input()` for storage and only comparing it against a value and discarding the actual input value we can drop `FILTER_SANITIZE_INPUT` and rely on `FILTER_UNSAFE_RAW`/`FILTER_DEFAULT`.

In some instances I think we will want to switch to `FILTER_SANITIZE_FULL_SPECIAL_CHARS`

The options I've added to the polyfill-ish function are arrived at after reviewing how we use the filter. In most cases we simply use it to encode and strip tags on a single string. Sometimes we act against an array (requiring the usage of `FILTER_REQUIRE_ARRAY` and sometimes we want to skip encoding quotes.

In instances where we *do not wish* to strip tags, `llms_filter_input()` can be used by itself with or without other filters / flags.

Can I get your immediate feedback on this before I move too much further along?

includes/llms.functions.core.php
```diff
@@ -603,6 +603,52 @@ function llms_get_instructors_defaults() {
...
+ * @return string|string[]|null|boolean Value of the requested variable on success, `false` if the filter fails, or `null` if the `$variable_name` variable is not set.
+ */
+function llms_filter_input_sanitize_string( $type, $variable_name, $flags = array() ) {
+
+	$require_array = in_array( FILTER_REQUIRE_ARRAY, $flags, true );
+
+	$string = llms_filter_input( $type, $variable_name, FILTER_UNSAFE_RAW, $require_array ? FILTER_REQUIRE_ARRAY : array() );
+
+	// If we have an empty string or the input var isn't found we can return early.
+	if ( '' === $string || null === $string || false === $string ) {
```
eri-trabiccolo: I think we can use `empty($string)` here.
https://www.php.net/manual/en/function.empty.php
https://www.php.net/manual/en/language.types.boolean.php#language.types.boolean.casting
So also with `"0"` we will return earlier, I think it's fine, no?

includes/models/model.llms.order.php
```diff
@@ -647,14 +647,23 @@ public function get_customer_full_address() {
...
 	 * @since 3.10.0 Unknown.
 	 * @since 5.3.1 Set the `post_date` property using `llms_current_time()`.
+	 * @since [version] Remove usage of deprecated `strftime()`.
 	 *
 	 * @param string $title Title to create the post with.
 	 * @return array
 	 */
 	protected function get_creation_args( $title = '' ) {
 
+		$ts = llms_current_time( 'timestamp' );
```
eri-trabiccolo: I'm not totally sure of the implications of this change here.
This impacts on the `post_date`.
Can we keep passing `llms_current-time( 'mysql' )` as the `post_date` below?

includes/models/model.llms.transaction.php
```diff
@@ -89,21 +107,20 @@ public function get_refundable_amount() {
...
 	 * @since 3.37.6 Add a default date information using `llms_current_time()`.
 	 *               Remove ordering placeholders from strftime().
+	 * @since [version] Remove usage of deprecated `strftime()`.
 	 *
 	 * @param int $order_id LLMS_Order ID of the related order.
 	 * @return array
 	 */
 	protected function get_creation_args( $order_id = 0 ) {
 
+		$ts = llms_current_time( 'timestamp' );
```
eri-trabiccolo: same as said for the order model.

eri-trabiccolo (changes requested):
I'm overall very happy you took care of this massive PR :P 
Looks good to me, just a couple of thoughts left I would like your opinion on.

includes/models/model.llms.order.php
thomasplevy: This is the same as before as `mysql` = `Y-m-d H:i:s: https://github.com/WordPress/wordpress-develop/blob/5.8.1/src/wp-includes/functions.php#L77-L79

Since I needed a timestamp for `date_create()` I figured pass that same timestamp to create the post date without having to call `llms_current_time()` again. Perhaps it's not much of an "improvement" but it seemed better to me this way without actually changing any existing behavior on the post creation date.

includes/llms.functions.core.php
thomasplevy: yea i guess that's okay. I was thinking that we shouldn't return early on `"0"` but after rereading it seems unnecessary to pass it further anyway since it won't be affected by any of the later logic.

includes/models/model.llms.order.php
eri-trabiccolo: Not sure man, I don't feel it's the same.
Because first you call `current_time( 'timestamp' )` with `$gmt=false` which is discouraged:
https://github.com/WordPress/wordpress-develop/blob/5.8.1/src/wp-includes/functions.php#L72
Then you get that `$ts` and turn it into a `mysql` format, which doesn't seem to be the same as calling `current_time( 'mysql' );`

includes/llms.functions.core.php
eri-trabiccolo: yeah I agree

includes/models/model.llms.order.php
thomasplevy: @eri-trabiccolo how bout this way?

thomasplevy: https://github.com/gocodebox/lifterlms/pull/1948/commits/5e9d33d9e408130ab600faa086eb3d257d424adf

eri-trabiccolo: Seems good, thanks <3

PR #1952, opened 2022-01-24
Full Site Editing / 2022 Compatibility

### LINKED ISSUE #1824: Twenty Twenty-Two theme support
Twenty Twenty-Two is the new WP default theme to be included in WordPress 5.9 

Upon release we should include new theme support for the theme

The first beta for 5.9 will be available 11/16 and the first RC on 11/30

I don't think we should spend any time on this until the beta is ready so we reduce the amount of API changes that may affect us.

Theme: https://github.com/WordPress/twentytwentytwo
5.9 dev schedule: https://make.wordpress.org/core/5-9/

eri-trabiccolo:
## Description

Fixes #1824 and adds FSE compatibilty.
run `npm run build:styles` prior to test 2022 compatiblity (for the forms).

Requires: https://github.com/gocodebox/lifterlms-blocks/pull/146

Additional tests need to be added.
There are at least two knowns "issues" at the moment that might be improved:
1) When with a FSE theme, any notice print at `llms_loop` on a non-singular template results in the notice be printed *before* the `html` opening tag. I was thinking about handling the notices printing in js, at least with FSE, trying to guess the first available element to which append the notice.
2) When a plugin block template is modified in the FSE, the "blue dot", which better shows the template is customized doesn't appear over it. This seems to be a WP bug to me, so I haven't investigated further.

## How has this been tested?
manually

## Types of changes

## Checklist:


includes/class-llms-block-templates.php
```diff
@@ -0,0 +1,564 @@
+<?php
+/**
+ * Handles the block templates
```
thomasplevy:
```suggestion
 * LLMS_Block_Templates class file.
```

Over time I've been switching the way the header matter is defined because the developer docs *do not parse* the file header. Adding a better description to the class header is then output on the parsed site at developer.lifterlms.com

includes/class-llms-block-templates.php
```diff
@@ -0,0 +1,564 @@
...
+ * @package LifterLMS/Classes
+ *
+ * @since [version]
+ * @version [version]
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * LLMS_Block_Templates class.
```
thomasplevy:
```suggestion
 * Handles the block templates.
```

includes/class-llms-block-templates.php
```diff
@@ -0,0 +1,564 @@
...
+	const LLMS_BLOCK_TEMPLATES_DIRECTORY_NAME = 'block-templates';
+
+	/**
+	 * Block Template namespace.
+	 *
+	 * This is used to save templates to the DB which are stored against this value in the wp_terms table.
+	 *
+	 * @var string
+	 */
+	const LLMS_BLOCK_TEMPLATES_NAMESPACE = 'lifterlms/lifterlms';
```
thomasplevy, line 36:
```suggestion
	const LLMS_BLOCK_TEMPLATES_NAMESPACE = 'llms/lifterlms';
```

Add-ons might be `llms/lifterlms-groups` for example?

I think using a shorter version where possible makes sense but maybe this isn't worth the change?

includes/class-llms-block-templates.php
```diff
@@ -0,0 +1,564 @@
...
+		);
+
+	}
+
+}
+
+$block_templates = new LLMS_Block_Templates();
+$block_templates->init();
+
+return $block_templates;
```
thomasplevy: What's the purpose of this new pattern instead of following preexisting conventions?

Most other classes return `new ClassName()` and have their initialization stuff in the `__construct()`, or, in the case of static classes, `return ClassName::init()`

If there's no good reason for this new convention I'd rather change `init()` to `__construct()` and switch the return to follow our existing patterns.

includes/class-llms-block-templates.php
```diff
@@ -0,0 +1,564 @@
...
+		 * Filter the block templates directories.
+		 *
+		 * @since [version]
+		 *
+		 * @param string[] Array of directory paths.
+		 */
+		$block_templates_base_paths = apply_filters(
+			'lifterlms_block_templates_directories',
+			array(
+				LLMS()->plugin_path() . '/templates/' . self::LLMS_BLOCK_TEMPLATES_DIRECTORY_NAME,
```
thomasplevy:
```suggestion
				llms()->plugin_path() . '/templates/' . self::LLMS_BLOCK_TEMPLATES_DIRECTORY_NAME,
```

includes/class-llms-block-templates.php
```diff
@@ -0,0 +1,564 @@
...
+
+	/**
+	 * Block Templates admin js strings.
+	 *
+	 * @since [version]
+	 *
+	 * @param string[] $strings Localization admin strings.
+	 * @return string[]
+	 */
+	public function add_block_templates_admin_strings_js( $strings ) {
```
thomasplevy: I'd rather code all this stuff into the block directly and rely on `@wordpress/i18n` in favor of using `LLMS.l10n` -- I've been working (silently) to phase out usage of `LLMS.l10n` and I don't see any advantage to doing this this way.

includes/class.llms.template.loader.php
```diff
@@ -28,9 +28,18 @@ class LLMS_Template_Loader {
 	 * @since 1.0.0
 	 * @since 3.20.0 Unknown.
 	 * @since 3.41.1 Predispose posts content restriction in REST requests.
+	 * @since [version] Handle block templates loading.
 	 */
 	public function __construct() {
 
+		// Template loading for FSE themes.
+		add_action(
```
thomasplevy: I don't want us to start using anonymous functions for hooks.I don't have any really good reason for this but it *feels* like it's a really bad pattern to start implementing, unless there's a really good reason for it. This seems like the reason is "I don't want to add a callback that only adds a filter" and that doesn't feel like a good reason for it in my opinion.

includes/class.llms.template.loader.php
```diff
@@ -462,52 +516,62 @@ public function template_loader( $template ) {
...
-			// Blog should bypass checks, except when sitewide restrictions are enabled.
 			if ( is_home() && 'sitewide_membership' === $page_restricted['reason'] ) {
-
 				// Prints notices on the blog page when there's not redirects setup.
 				add_action( 'loop_start', 'llms_print_notices', 5 );
+			}
+		}
 
-				return $template;
+		$forced_template = apply_filters( 'llms_force_llms_template_loading', true ) ? $this->get_maybe_forced_llms_template() : false;
```
thomasplevy:
```suggestion
		$forced_template = apply_filters( 'llms_force_template_loading', true ) ? $this->get_maybe_forced_llms_template() : false;
```

Also needs docblock

includes/class.llms.template.loader.php
```diff
@@ -462,52 +516,62 @@ public function template_loader( $template ) {
...
-			if ( in_array( get_post_type(), array( 'course', 'llms_membership' ), true ) ) {
-				return $template;
+	/**
+	 * Retrieve the hierarchical template to be loaded.
+	 *
+	 * @since [version]
+	 *
+	 * @return null|string
+	 */
+	private function get_maybe_forced_llms_template() {
```
thomasplevy:
```suggestion
	private function get_maybe_forced_template() {
```

Is `llms` necessary in our own internal private function?

includes/functions/llms.functions.template.php
```diff
@@ -254,3 +255,22 @@ function llms_get_template_override_directories() {
...
+ * @param string $template           Template file name, without extension.
+ * @param string $extension          Template file extension.
+ * @param string $template_directory Template directory relative to the plugin base directory.
+ * @return string
+ */
+function llms_template_file_path( $template, $extension = 'php', $template_directory = 'templates' ) {
+
+	// We have reason to use a LifterLMS template, check if there's an override we should use from a theme / etc...
+	$override      = llms_get_template_override( $template, $extension );
+	$template_path = $override ? $override : LLMS()->plugin_path() . "/{$template_directory}/";
```
thomasplevy:
```suggestion
	$template_path = $override ? $override : llms()->plugin_path() . "/{$template_directory}/";
```

includes/functions/llms.functions.template.php
```diff
@@ -193,23 +193,24 @@ function llms_locate_template( $template_name, $template_path = '', $default_pat
 }
 
 /**
- * Get template override
+ * Get template override.
  *
  * @since Unknown
  * @since 4.8.0 Move template override directories logic into llms_get_template_override_directories.
+ * @since [version] Added `$extension`, and `$is_block_template` parameter.
```
thomasplevy:
```suggestion
 * @since [version] Added `$extension` parameter.
```

includes/functions/llms.functions.template.php
```diff
@@ -193,23 +193,24 @@ function llms_locate_template( $template_name, $template_path = '', $default_pat
...
  *
- * @param string $template Template file.
+ * @param string $template  Template file.
+ * @param string $extension The file extension.
  * @return mixed Template file or false if none exists.
  */
-function llms_get_template_override( $template = '' ) {
+function llms_get_template_override( $template = '', $extension = '' ) {
```
thomasplevy: This means that block templates and php templates can both be stored in the same override directory, right? 

A new block template (in LifterLMS) is stored in `templates/block-templates/archive-course.html` and a php template could be at `templates/archive-course.php`

So if a theme wanted to override both of these, the directory would have them at `lifterlms/archive-course.html` and `lifterlms/archive-course.php` (both in the same folder)

Right?

I think this is fine but I just want to make sure I'm reading this right.

templates/block-templates/single-certificate.html
```diff
@@ -0,0 +1,2 @@
+
```
thomasplevy: Is the leading space needed for some reason?

templates/block-templates/single-no-access.html
```diff
@@ -0,0 +1,5 @@
+<!-- wp:template-part {"slug":"header"} /-->
+<!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group"><!-- wp:llms/php-template {"template":"single-no-access", "title":"LifterLMS Single Template requiring Membership"} /--></div>
```
thomasplevy:
```suggestion
<div class="wp-block-group"><!-- wp:llms/php-template {"template":"single-no-access", "title":"LifterLMS Template Access Restricted"} /--></div>
```

This is not often displayed but I don't think it's displayed exclusively when membership is required so I think the title should reflect that to prevent confusion or bug reports or whatever.

thomasplevy (changes requested):

thomasplevy:
@eri-trabiccolo overall I think this is really solid. I've tested it on 5.9-rc4 and haven't spotted any show-stopping issues

However, when loading the membership catalog and single certificate templates (in the site editor) the php-template block renders no title:

Screenshot_2022-01-24_16-27-07
Screenshot_2022-01-24_16-27-54

includes/functions/llms.functions.template.php
eri-trabiccolo: Nope, themes have to override the block templates in
`lifterlms/block-templates/`

The reason for `$extension` is that... I think we just don't need it anymore. I added it when I was doing something else, but it's not needed anymore (gonna check). I'll remove it. :D

templates/block-templates/single-no-access.html
eri-trabiccolo: mmm I think you're right :D

templates/block-templates/single-certificate.html
eri-trabiccolo: no, typo, thanks.

includes/class-llms-block-templates.php
eri-trabiccolo: The purpose of this is that I want to be able to call this object methods from another file without adding the hooks.
See: 
https://github.com/gocodebox/lifterlms/pull/1952/files#diff-7947d18155c2e03348f57a9ef19a46d9e47a3a999f09f521726fb3c2e82a7be7R466

that `init()` should be actually be renamed `hooks()` or something of the sort.
Hooks addition are not `initialization` to me.

If you prefer having them in the constructor I'd say that this must become a singleton, and I add a `block_templates()` method here:
https://github.com/gocodebox/lifterlms/blob/dev/class-lifterlms.php

what do you think?

includes/class-llms-block-templates.php
eri-trabiccolo: I used that because (looking at wc) I found that WP automatically turns `lifterlms/lifterlms` (that we use as the `theme` block template property) into LifterLMS (you know, plugin name etc.).
But I think we can, as you propose, go with your suggestion and use this that I called `namespace` to build the id. Then introduce a new constant like:
`const LLMS_BLOCK_TEMPLATES_THEME = 'lifterlms/lifterlms;`
so everybody's happy and makes more sense. Let me check if this doesn't introduce any issue.

includes/class-llms-block-templates.php
eri-trabiccolo: Well my idea is that this was more easy to extend, in an add-on.
E.g. I'm going to add a block template for the groups plugin, and I wanted to only add the template html + a php file where I add all the filters needed to load the template, and its js title, correctly.
But that's fine I'll add some inline script in groups to do the same (and a filter hook in https://github.com/gocodebox/lifterlms-blocks/pull/147/files#diff-c12382294566f22f66d320592d72c6293bae6b752c5b4eb0296d7990685d81c5R4). As I don't feel to add a js file only for that in groups.

includes/class.llms.template.loader.php
eri-trabiccolo: Well yeah, I thought that isn't a bad pattern and couldn't think why not using it in this specific case. :D 
Why is it a bad pattern? I think it is bad when you might want to unhook a callback, but:
1) in this specific case you wouldn't be able to unhook the `template_redirect` callback from outside this class in any case as the object is instantiated on file loading
2) if it wasn't for 1) you could always unhook the `pre_get_block_templates` callback hooking your "unhooking logic" at `template_redirect|11`.
That was my reasoning: no reason to not do that, and gold.

That said, I'm going to follow your feeling <3

includes/class.llms.template.loader.php
eri-trabiccolo: wasn't a prefixing choice, it was more a naming choice, a name that tells we're talking about llms specific templates.

eri-trabiccolo:
> @eri-trabiccolo overall I think this is really solid. I've tested it on 5.9-rc4 and haven't spotted any show-stopping issues
> 
> However, when loading the membership catalog and single certificate templates (in the site editor) the php-template block renders no title:
> 
> Screenshot_2022-01-24_16-27-07 Screenshot_2022-01-24_16-27-54

Oh thanks for reporting! It's because of the double closing brace on this line:
https://github.com/gocodebox/lifterlms/pull/1952/files#diff-8fc257ecd0c4769c75c10a3a7fe9b40a8b519caa0f9ab5b69a5d182055dfe6caR3

:D

includes/class-llms-block-templates.php
eri-trabiccolo: going to revert this because the first part of the id and the "theme name" are somewhat tied, also, what will be saved in the DB (which might occupy more space) is the theme name, and that - for the reason above - I would like to keep it as `lifterlms/lifterlms`.

includes/class-llms-block-templates.php
thomasplevy: Mostly what I don't like here is having functionality / logic buried at the bottom of a class file, especially since we *don't do this anywhere else* it makes it hard to locate in the future since we're not primed to look for any functional code at the bottom of a class file.

I think, based on existing patterns, the singleton makes the most sense then.

includes/class-llms-block-templates.php
thomasplevy: @eri-trabiccolo I'm find with localizing with PHP instead of JS for lighter-weight extensibility in this case. However, I don't see any reason to use `LLMS.l10n` in order to do so.

Instead, keep this but turn it into a JS localization object and then in JS we can read the map to output the object instead of loading them into `LLMS.l10n` and reading them that way. I think that will work, you?

eri-trabiccolo: so I'm going to localize our blocks script in a more formal way?!

thomasplevy: @eri-trabiccolo I don't think it needs to be anything more formal than outputting an object of translation data, mapping the template ID/name to the translated title, and outputting it as a JS variable using [wp_localize_script()](https://developer.wordpress.org/reference/functions/wp_localize_script/). Add a filter on the object before it's passed into the function. Then add-ons can add their own data to it via PHP.

In the legacy block JS, read the map I hard-coded into the `getLabel()` method from the localized JS object instead of it being hard coded.

eri-trabiccolo: oh, with formal I meant use `wp_localize_script()` yeah :D, rather than build the object as we do in other parts of the codebase. <3

includes/class-llms-block-templates.php
thomasplevy: Okay! <3

includes/class-llms-block-templates.php
eri-trabiccolo: @thomasplevy singleton thing done

includes/class-llms-block-templates.php
eri-trabiccolo: php-block templates localized

eri-trabiccolo:
@thomasplevy wanna have another look at this?

includes/class-llms-block-templates.php
```diff
@@ -0,0 +1,663 @@
...
+	 * Localize block templates.
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function localize_blocks() {
+		wp_localize_script(
+			'llms-blocks-editor',
+			'LLMS_Block_Templates_l10n',
```
thomasplevy: I know that in a lot of the codebase we use phpcs standards for JS and this kind of stuff but as since this will ultimately be a JS variable we should make it a JS-style variable.

```suggestion
			'llmsBlockTemplatesL10n',
```

thomasplevy (changes requested):

PR #1961, opened 2022-01-28
Cert back btn

### LINKED ISSUE #1959: [6.0] Return to (student) Dashboard (from certificate screen)
Students no longer see the header or menu on certificates and will want a way to go back to their dashboard or other content that isn't the back button 

image

thomasplevy:
## Description

Closes #1959

Screenshot_2022-01-28_11-12-19

Added a text-link (not a button) on the bottom left. This returns users to the "My Certificates" endpoint on the student dashboard if it's enabled and returns to the student dashboard (main page) if the cert endpoint is not enabled.

Language says "Dashboard" instead of "All Certificates" if the endpoint is disabled.

## How has this been tested?

+ Manually

## Types of changes
+Update

## Checklist:

### DIFF
--- a/assets/scss/certificates.scss
+++ b/assets/scss/certificates.scss
@@ -69,6 +69,7 @@ body {
  */
 .llms-print-certificate {
 	margin-top: 40px;
+	margin-bottom: 40px;
 	text-align: center;
 
 	form {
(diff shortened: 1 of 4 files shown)


eri-trabiccolo (changes requested):
I think this link should only be visible:
1) on awarded certificates
2) if the current user is the certificate's "owner"

thomasplevy:
> I think this link should only be visible:
> 
> 1. on awarded certificates
> 2. if the current user is the certificate's "owner"

Okay, I agree with point 1, implemented on the latest commit.

I left point 2 off because I don't think the additional logic is necessary. I think, while not strictly necessary, if an admin is viewing a student's earned cert and it looks identical to what the student will actually see, that will be beneficial. Even though the admin might get confused by the link it seems like it would be good for them to see what the student will see. Maybe I'm wrong about this but I don't know....

eri-trabiccolo (approved):

PR #1967, opened 2022-02-03
Fixed how protected properties are accessed.

pondermatic:
## Description
Replaced reading of protected properties with method calls.

## How has this been tested?
`composer run tests` and `composer check-cs-errors`.

## Types of changes

## Checklist:

### DIFF
--- a/includes/abstracts/llms.abstract.notification.controller.php
+++ b/includes/abstracts/llms.abstract.notification.controller.php
@@ -458,7 +458,7 @@ abstract class LLMS_Abstract_Notification_Controller extends LLMS_Abstract_Optio
 			)
 		);
 
-		return (bool) $query->get_found_results();
+		return $query->has_results();
 
 	}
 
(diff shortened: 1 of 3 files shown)


includes/abstracts/llms.abstract.notification.controller.php
```diff
@@ -457,7 +458,7 @@ public function has_subscriber_received( $type, $subscriber ) {
 			)
 		);
 
-		return $query->found_results ? true : false;
+		return (bool) $query->get_found_results();
```
thomasplevy:
```suggestion
		return $query->has_results();
```

I think that this maybe semantically doesn't matter that much but this function already exists

thomasplevy (changes requested):

includes/abstracts/llms.abstract.notification.controller.php
pondermatic: The result is the same but it is semantically better. Thanks!

pondermatic: I wish I had given that commit a better message than the default on that GitHub provided.

PR #1975, opened 2022-02-05
Reset certificate

### LINKED ISSUE #1960: [6.0] Reset to Default in Certificate Builder
A reset to default button in the certificate builder like what we have for forms, so if something breaks or the user does too many weird changes they don't like we can get it back to the default easily to fix the issue!

thomasplevy:
Closes #1960

Adds a "Reset template" button in the post status header area (near "move to trash" allowing reseting the certificate to the default template.

The reset template button displays on certificate templates and on awarded templates which are NOT connected to a template. The idea behind this last piece of logic is that if it's connected with a template it should be synced with that template, not with the default post type template.

In order to expose the template post (stored as wp_post.post_parent) I have added a rest field, `certificate_template`. I initially switched the `llms_my_certificate` post type to be hierarchical but the WP admin UI for heirachical post types doesn't look nice when the parent post is of a different post type. I started futzing with fixing the UI and ran into too many issues. This seemed to be a much better way to continue using the post_parent field without having to also disable all the default admin chrome.

Tested manually and with new E2E tests


eri-trabiccolo (commented):
Looks good although:
1) I think the reset button should only appear for the certificate template post types, and not for the awarded certificates, or at least not for awarded certificates which have a parent template (in that case I think it should only possible to sync them).
2) When resetting to default the following alert appears:
> Are you sure you wish to replace the certificate content with the original default layout? This action cannot be undone!

This doesn't seem to be completely true, as if you don't also click on Update the "reset" doesn't take place. Am I wrong? Is this expected?

thomasplevy:
> 1. I think the reset button should only appear for the certificate template post types, and not for the awarded certificates, or at least not for awarded certificates which have a parent template (in that case I think it should only possible to sync them).

Yea good call. I'll update.

> 2. When resetting to default the following alert appears:
> 
> > Are you sure you wish to replace the certificate content with the original default layout? This action cannot be undone!
> 
> This doesn't seem to be completely true, as if you don't also click on Update the "reset" doesn't take place. Am I wrong? Is this expected?


It is not at all true, however, I didn't want to write a message that explained it *can only be undone if you don't save or if you do save and later use a revision or hit undo after confirming. I figured it's a potentially destructive action that should be considered "final". Adding the confirm dialog at all might not be necessary but I don't want users to be resetting and then yelling at us and expecting us to document the way they can undo it. We get feedback about "desctructive" actions not having a confirmation dialog so I've put one in. Yes, it's not technically true that it can be undone but I don't think anyone is going to be upset about the lie. Though they will be upset if we don't tell them it can't be undone and then they can't figure out how to undo it. Better give them a stern warning. Right? Maybe I'm wrong here. What do you suggest instead?

eri-trabiccolo:
I'm not sure how to better describe it to the users. I'm "worried" about people clicking on "Reset", "okaying" he dialog, not updating the certificate and saying "Hey the reset didn't work, it worked in the editor but then when I viewed it in front the old content was shown, additionally when back in the editor the old conte was still there, how so!".
So I guess that the first thing I thought was that the action was really destructive, and it should, like the sync does.
So that's what I suggest, I think.

thomasplevy:
@eri-trabiccolo How about performing the update and then saving it automatically?

Or would you suggest leaving it as is and not saying "This cannot be undone"

eri-trabiccolo:
> @eri-trabiccolo How about performing the update and then saving it automatically?

This. IMHO

thomasplevy:
@eri-trabiccolo I agree with you, working on it now <3

thomasplevy:
@eri-trabiccolo in addition to the changes above, I've also added a condition to switch publish posts to draft during a reset. My thinking is that, especially with awarded certificates, the user should have time to work with the reset certificate before physically applying those changes on the frontend.

includes/class-llms-rest-fields.php
```diff
@@ -96,12 +96,50 @@ public function register() {
...
+						},
+					),
+				),
+				'get_callback'    => function( $object ) {
+					return wp_get_post_parent_id( $object['id'] );
+				},
+				'update_callback' => function( $value, $post ) {
+					$update = array(
+						'ID'          => $post->ID,
+						'post_parent' => $value,
```
eri-trabiccolo, line 134: shall we sanitize this to be sure it's an int?

eri-trabiccolo (changes requested):
Looks very good to me. :D

includes/class-llms-rest-fields.php
thomasplevy: the schema type will take care of sanitizing the values before it gets to the update callback, see https://developer.wordpress.org/reference/functions/rest_sanitize_value_from_schema/

PR #1985, opened 2022-02-09
Add post table row for Cert & Achievement templates that links to the…

### LINKED ISSUE #1932: [6.x] add post table row for Cert & Achievement templates that links to the filtered list of awards for the given template.

eri-trabiccolo:
… filtered list of awards for the given template

## Description
Fixes #1932 
I did consider to make the link appear ONLY if the template actually had awards referring to it, but I considered that potentially too much heavy (a new query should run for each item, and depending on the screen settings they could be a lot), for a relatively not crucial feature.
Let me know if you think otherwise.

## How has this been tested?
Manually and with unit tests.

## Types of changes

## Checklist:


includes/admin/post-types/post-tables/class-llms-admin-post-table-achievements.php
```diff
@@ -0,0 +1,40 @@
...
+require_once LLMS_PLUGIN_DIR . '/includes/traits/llms-trait-certificate-achievement-template-post-list-table.php';
+
+/**
+ * Customize display of the achievement post table.
+ *
+ * @since [version]
+ */
+class LLMS_Admin_Post_Table_Achievements {
+
+	use LLMS_Trait_Certificate_Achievement_Templates_Post_List_Table;
```
thomasplevy: I don't like this classname, I think I'd prefer:

```suggestion
	use LLMS_Trait_Award_Templates_Post_List_Table;
```

includes/admin/post-types/post-tables/class-llms-admin-post-table-certificates.php
```diff
@@ -10,13 +10,18 @@
...
+require_once LLMS_PLUGIN_DIR . '/includes/traits/llms-trait-certificate-achievement-template-post-list-table.php';
+
 /**
  * Customize display of the certificate post tables
  *
  * @since [version]
  */
 class LLMS_Admin_Post_Table_Certificates {
 
+	use LLMS_Trait_Certificate_Achievement_Templates_Post_List_Table;
```
thomasplevy:
```suggestion
	use LLMS_Trait_Award_Templates_Post_List_Table;
```

includes/traits/llms-trait-certificate-achievement-template-post-list-table.php
```diff
@@ -0,0 +1,87 @@
...
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * LifterLMS Certificate/Achievement Templates Post List Table trait.
+ *
+ * @since [version]
+ */
+trait LLMS_Trait_Certificate_Achievement_Templates_Post_List_Table {
```
thomasplevy:
```suggestion
trait LLMS_Trait_Award_Templates_Post_List_Table {
```

includes/traits/llms-trait-certificate-achievement-template-post-list-table.php
```diff
@@ -0,0 +1,87 @@
...
+	);
+
+	/**
+	 * Add post row actions filter callback.
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function certificate_achievement_template_row_actions() {
```
thomasplevy:
```suggestion
	protected function award_template_row_actions() {
```
This doesn't need to be public, does it?

And using generic "award" rather than "certificate_achievement"

thomasplevy (changes requested):
I don't love the classname you've chosen -- using "certificate_achievement"  --- I think using "award_template" is satisfactory. It's shorter and allows extensibility later if we ever end up adding other award types.

Additionally, I envisioned this is a new column... I know I said row in the issue. I mixed up my language.

I figured this would show a column with a count & link or an `&ndash` in the event that there were no awards generated from the template.

You identified that running a query on each row to get that count would be inefficient and you're not wrong about that and I figured "This is probably a better way to handle it" but now that I'm seeing it in action I don't like it as a post row action (it's long, only displays on hover, etc...)

Maybe this is the best solution though... what are your thoughts here?

includes/traits/llms-trait-certificate-achievement-template-post-list-table.php
eri-trabiccolo: you're right, but then I want it to be private. :D

thomasplevy: Well, it can't be.

eri-trabiccolo: ahah yeah right

eri-trabiccolo:
> I don't love the classname you've chosen -- using "certificate_achievement" --- I think using "award_template" is satisfactory. It's shorter and allows extensibility later if we ever end up adding other award types.

I don't love them too. You're right

> Additionally, I envisioned this is a new column... I know I said row in the issue. I mixed up my language.

Ok.
 
> I figured this would show a column with a count & link or an `&ndash` in the event that there were no awards generated from the template.

I don't know man, count seems to heavy to me, we might end up saving this info in a meta, is it worthwhile?

> You identified that running a query on each row to get that count would be inefficient and you're not wrong about that and I figured "This is probably a better way to handle it" but now that I'm seeing it in action I don't like it as a post row action (it's long, only displays on hover, etc...)

Let's go with the column then, but without counts

> Maybe this is the best solution though... what are your thoughts here?

^

thomasplevy:
@eri-trabiccolo I don't want to mess with saving it in a meta... too much code to keep it in sync properly I think.

Anyway I think the column without meta is good, but what should the text of the column be if it's not a count? Just "view awarded certificate?" that doesn't seem right to me as the columns are usually meant to display information which sometimes link to a filtered view (I'm thinking like how categories are listed in a column and clicking the category name will filter the view down to posts in that category).

eri-trabiccolo:
> @eri-trabiccolo I don't want to mess with saving it in a meta... too much code to keep it in sync properly I think.
> 
> Anyway I think the column without meta is good, but what should the text of the column be if it's not a count? Just "view awarded certificate?" that doesn't seem right to me as the columns are usually meant to display information which sometimes link to a filtered view (I'm thinking like how categories are listed in a column and clicking the category name will filter the view down to posts in that category).

Well about categories, I think the good similitude is how categories look in their  `wp-admin/edit-tags.php?taxonomy=category`, where the posts count is shown, and that's a term_taxonomy db column :(

I'm not sure what to do here.

thomasplevy:
@eri-trabiccolo alright leave it as is. We can always (never) change it later. I don't like any other potential solution so this one is good!

PR #1994, opened 2022-02-14
Add a sync action to the list of templates.

### LINKED ISSUE #1933: [6.x] Add template syncing to the cert and awarded cert post tables
As a post row action

I think that a bulk action *should* also be possible (spawning multiple processors) but if the bulk action isn't possible then that's fine, I think.

This will be dependent on #1931 as currently achievments don't have any sync functionality

pondermatic:
## Description

Added a sync action to the list of templates in the admin posts table.

Fixes #1933.

## How has this been tested?
Manually and automatically.

## Types of changes

## Checklist:

### DIFF
--- a/includes/abstracts/llms-abstract-processor-user-engagement-sync.php
+++ b/includes/abstracts/llms-abstract-processor-user-engagement-sync.php
@@ -38,13 +38,22 @@ abstract class LLMS_Abstract_Processor_User_Engagement_Sync extends LLMS_Abstrac
 	protected const TEXT_SYNC_NOTICE_AWARDED_ENGAGEMENTS_COMPLETE = 1;
 
 	/**
+	 * A text type for an admin notice that there are no awarded engagements to sync the template with.
+	 *
+	 * @since [version]
+	 *
+	 * @var int
+	 */
+	protected const TEXT_SYNC_NOTICE_NO_AWARDED_ENGAGEMENTS = 2;
+
+	/**
 	 * A text type for an admin notice that the sync of awarded engagements to an engagement template is scheduled.
 	 *
 	 * @since [version]
 	 *
 	 * @var int
 	 */
-	protected const TEXT_SYNC_NOTICE_SCHEDULED = 2;
+	protected const TEXT_SYNC_NOTICE_SCHEDULED = 3;
 
 	/**
 	 * Clear notices.
@@ -226,8 +235,16 @@ abstract class LLMS_Abstract_Processor_User_Engagement_Sync extends LLMS_Abstrac
 
 		$this->clear_notices( $engagement_template_id );
 
-		$args = array( $engagement_template_id );
-		if ( wp_next_scheduled( $this->schedule_hook, $args ) ) {
+		$args           = array( $engagement_template_id );
+		$awarded_count = $this->count_awarded_engagements( $engagement_template_id );
+
+		if ( 0 === $awarded_count ) {
+
+			$log_message    = 'no awarded %1$ss to bulk sync with the %1$s template %2$s (#%3$d)';
+			$notice_message = $this->get_text( self::TEXT_SYNC_NOTICE_NO_AWARDED_ENGAGEMENTS, compact( 'engagement_template_id' ) );
+			$notice_id      = 'awarded-%1$ss-sync-%2$d-no-awarded';
+
+		} elseif ( wp_next_scheduled( $this->schedule_hook, $args ) ) {
 
 			$log_message    = 'awarded %1$ss bulk sync already scheduled for the %1$s template %2$s (#%3$d)';
 			$notice_message = $this->get_text( self::TEXT_SYNC_NOTICE_ALREADY_SCHEDULED, compact( 'engagement_template_id' ) );
--- a/includes/abstracts/llms-abstract-processor-user-engagement-sync.php
+++ b/includes/abstracts/llms-abstract-processor-user-engagement-sync.php
@@ -208,6 +208,7 @@ abstract class LLMS_Abstract_Processor_User_Engagement_Sync extends LLMS_Abstrac
 			array(
 				'dismissible'      => true,
 				'dismiss_for_days' => 0,
+				'type'             => 'success',
 			)
 		);
 	}
@@ -235,7 +236,7 @@ abstract class LLMS_Abstract_Processor_User_Engagement_Sync extends LLMS_Abstrac
 
 		$this->clear_notices( $engagement_template_id );
 
-		$args           = array( $engagement_template_id );
+		$args          = array( $engagement_template_id );
 		$awarded_count = $this->count_awarded_engagements( $engagement_template_id );
 
 		if ( 0 === $awarded_count ) {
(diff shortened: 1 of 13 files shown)


eri-trabiccolo (changes requested):
I think that in this case we should inform the user, better sooner than later, when there are no awards to sync, I mean:

If you test this with a template which has no awards, a sync scheduled notice appears and then nothing else.

includes/traits/llms-trait-award-templates-post-list-table.php
```diff
@@ -80,23 +80,37 @@ public function add_post_actions( $actions, $post ) {
...
-		$engagement_type = str_replace( 'llms_', '', $post->post_type );
-		$sync_action     = "sync_awarded_{$engagement_type}s";
-		$sync_url        = add_query_arg(
-			array(
-				'action'                                      => $sync_action,
-				"_llms_{$engagement_type}_sync_actions_nonce" => wp_create_nonce( "llms-{$engagement_type}-sync-actions" ),
-			),
-			get_edit_post_link( $post, 'raw' )
-		);
+		$awarded_number = $this->count_awarded_engagements( $post->ID );
```
eri-trabiccolo: don't you think that running this query for each and every template in the list might be a bit too much heavy?
I did consider doing the same when adding the "View Awarded *" link, see: https://github.com/gocodebox/lifterlms/pull/1985#issue-1128842305

I think would be better to run the check when you click on the link. I mean your solution is cleaner, but I think it might be pretty heavy for no real reason. What do you think?

pondermatic: You have a good point @eri-trabiccolo.

Every template should have a "Sync Awarded Xs" post action.
When there are no awarded engagements to sync, display an admin notice. 'There are no awarded Xs to sync with the template "NAME" (#ID).`

Is this what you had in mind?

eri-trabiccolo: Yeah. That's what I had in mind. <3

includes/admin/post-types/post-tables/class-llms-admin-post-table-achievements.php
```diff
@@ -20,7 +20,7 @@
  */
 class LLMS_Admin_Post_Table_Achievements {
 
-	use LLMS_Trait_Award_Templates_Post_List_Table;
+	use LLMS_Trait_Award_Templates_Post_List_Table, LLMS_Trait_User_Engagement_Type;
```
eri-trabiccolo:
```suggestion
	use LLMS_Trait_Award_Templates_Post_List_Table;
	use LLMS_Trait_User_Engagement_Type;
```

I think I would prefer this notation for two _minor_ reasons:
1) that list might become long and, for readability, at a certain point will be forced to break on a new line, so let's do it by default
2) say I want to search for for all those classes which use ` LLMS_Trait_User_Engagement_Type`, the current notation makes it a bit harder, I should use reg exp at least.
What do you think?

includes/traits/llms-trait-award-templates-post-list-table.php
```diff
@@ -80,6 +80,24 @@ public function add_post_actions( $actions, $post ) {
 			)
 		);
 
+		$engagement_type = str_replace( 'llms_', '', $post->post_type );
```
eri-trabiccolo: I think we can refactor the whole method a little bit.
I realize two things here:
1) `$engagement_type` === (basically) `$this->engagement_type`.
2) this method runs twice on each table, that's why we can't use 1)
I think the best will be to turn the trait's private `post_types` array into a protected `post_type` that each class that use this trait will define in their constructor (as you do with the `engagement_type`), and early bail if the `$post->post_type !== $this->post_type`.
What do you think?

includes/traits/llms-trait-user-engagement-type.php
```diff
@@ -26,6 +26,32 @@ trait LLMS_Trait_User_Engagement_Type {
...
+
+		$awarded_engagements_query = new LLMS_Awards_Query(
+			array(
+				'fields'    => 'ids',
+				'templates' => $template_id,
+				'per_page'  => 1,
+				'status'    => array(
+					'publish',
+					'future',
+				),
```
eri-trabiccolo:
```suggestion
					'future',
					'draft',
				),
```
I think we want this too now.
At the beginning we just didn't allow for `draft` awards.

eri-trabiccolo (changes requested):
Some changes requested, but overall works fine.

includes/abstracts/llms-abstract-processor-user-engagement-sync.php
```diff
@@ -226,8 +235,16 @@ public function schedule_sync( $engagement_template_id ) {
 
 		$this->clear_notices( $engagement_template_id );
 
-		$args = array( $engagement_template_id );
-		if ( wp_next_scheduled( $this->schedule_hook, $args ) ) {
+		$args           = array( $engagement_template_id );
```
eri-trabiccolo:
```suggestion
		$args          = array( $engagement_template_id );
```

includes/abstracts/llms-abstract-processor-user-engagement-sync.php
```diff
@@ -226,8 +235,16 @@ public function schedule_sync( $engagement_template_id ) {
...
+		$awarded_count = $this->count_awarded_engagements( $engagement_template_id );
+
+		if ( 0 === $awarded_count ) {
+
+			$log_message    = 'no awarded %1$ss to bulk sync with the %1$s template %2$s (#%3$d)';
+			$notice_message = $this->get_text( self::TEXT_SYNC_NOTICE_NO_AWARDED_ENGAGEMENTS, compact( 'engagement_template_id' ) );
+			$notice_id      = 'awarded-%1$ss-sync-%2$d-no-awarded';
+
```
eri-trabiccolo, line 248: also, what do you think if we use the notice [`type`](https://github.com/gocodebox/lifterlms/blob/dev/includes/admin/class.llms.admin.notices.php#L108) too?
I'd say:
- no awards => error
- already scheduled => warning
- scheduled => info
- completed => success

includes/traits/llms-trait-award-templates-post-list-table.php
pondermatic: `LLMS_Trait_Award_Templates_Post_List_Table::add_post_actions()` is called twice for each row in the post table because `LLMS_Trait_Award_Templates_Post_List_Table::award_template_row_actions()` is called by `LLMS_Admin_Post_Table_Achievements::__construct()` and `LLMS_Admin_Post_Table_Certificates::__construct()`.

In `LLMS_Trait_Award_Templates_Post_List_Table`, if we replace 
```php
	protected function award_template_row_actions() {

		add_filter( 'post_row_actions', array( $this, 'add_post_actions' ), 20, 2 );

	}
```
with
```php
	protected function award_template_row_actions() {

		if ( "llms_{$this->engagement_type}" === llms_filter_input( INPUT_GET, 'post_type', FILTER_SANITIZE_STRING ) ) {
			add_filter( 'post_row_actions', array( $this, 'add_post_actions' ), 20, 2 );
		}

	}
```
we won't need to check the post type in `LLMS_Trait_Award_Templates_Post_List_Table::add_post_actions()` (although we will for a sanity check and for `LLMS_Test_Trait_Award_Templates_Post_List_Table::test_add_post_actions()`).

eri-trabiccolo: I think this makes sense too.
<3

includes/abstracts/llms-abstract-processor-user-engagement-sync.php
pondermatic: That's a great suggestion!

eri-trabiccolo (approved):
Looks good to me <3

Issue #1997, opened 2022-02-15
Checkout broken on Twenty Twenty Two Theme


thomasplevy:
### Reproduction Steps

+ Load the checkout screen using the 2022 theme
+ Observe the JS errors in the console

```
llms-form-checkout.js?ver=5.9.0:65 Uncaught TypeError: Cannot read properties of undefined (reading 'checkout')
    at llms_checkout.init (llms-form-checkout.js?ver=5.9.0:65:38)
    at new llms_checkout (llms-form-checkout.js?ver=5.9.0:568:8)
    at llms-form-checkout.js?ver=5.9.0:575:25
    at llms-form-checkout.js?ver=5.9.0:577:4
init @ llms-form-checkout.js?ver=5.9.0:65
llms_checkout @ llms-form-checkout.js?ver=5.9.0:568
(anonymous) @ llms-form-checkout.js?ver=5.9.0:575
(anonymous) @ llms-form-checkout.js?ver=5.9.0:577
llms-stripe.js?ver=5.4.0:388 Uncaught TypeError: Cannot read properties of undefined (reading 'add_gateway')
    at llms-stripe.js?ver=5.4.0:388:23
    at llms-stripe.js?ver=5.4.0:392:2
```

### Notes

This, I think, is a change in how FSE themes handle asset enqueue... On this line we conditionally enqueue the PW strength meter JS: https://github.com/gocodebox/lifterlms/blob/trunk/includes/forms/class-llms-form-field.php#L788

But this line never runs because on the theme scripts haven't been enqueued when the HTML is parsed? I guess maybe there's a pre-parse that happens with FSE and these php templates or something? I'm not sure.. I haven't investigate thorougly.

A simple fix may be to remove this conditional and just define the pw strength meter script as a dependency of the checkout script and always localize with the PW strength meter stuff. This is a small performance hit for those not using the PW strength meter but it's a less convoluted way of defining the dependencies. And my guess is we wouldn't encounter this issue if we did it this way regardless of what's going on in the theme (or maybe FSE themes in general??)

---

Edit: The stripe issue is irrelevant because the checkout script never fully initializes because of the initial error. I included it for thoroughness though.

eri-trabiccolo:
Sorry @thomasplevy this is due to the fact that I tend to disable the password strength meter :/
Anyways we can fix it in a simple way.

PR #2003, opened 2022-02-17
Add missing changelog yml for #1997

eri-trabiccolo:
missing changelog for #1997

### DIFF
--- a/.changelogs/2022-checkout.yml
+++ b/.changelogs/2022-checkout.yml
@@ -1,0 +1,5 @@
+significance: patch
+type: fixed
+links: 
+- "#1997"
+entry: Fixed broken checkout on Twenty Twenty Two Theme when using password strength meter.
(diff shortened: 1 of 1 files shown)


.changelogs/2022-checkout.yml
```diff
@@ -0,0 +1,5 @@
+significance: patch
+type: fixed
+links: 
+- "#1997"
+entry: Fixed broken checkout on Twenty Twenty Two Theme when using password strength meter.
```
thomasplevy:
```suggestion
entry: Fixed broken checkout on Twenty Twenty-Two Theme when using password strength meter.
```

Source: https://github.com/WordPress/twentytwentytwo#twenty-twenty-two

thomasplevy (changes requested):

.changelogs/2022-checkout.yml
eri-trabiccolo: booo

PR #2010, opened 2022-02-22
Handle specific form title for free access plans

### LINKED ISSUE #1774: “Billing Information” is not replaced with “Student Information”
### Reproduction Steps

Create an access plan for a course and check the "free" option.
Then visit the purchase link

### Expected Behavior

According to: https://lifterlms.com/docs/checkout-free-access-plans/ "Billing Information" should be replaced with "Student Information"

### Actual Behavior

Stays as "Billing Information"

### Error Messages / Logs

```
N/A

```

<details>

```
N/A

```

</details>

This issue has be recreated:
+ [x] Locally
+ [x] On a staging site
+ [x] On a production website

eri-trabiccolo:
## Description

Fixes #1774 
requires https://github.com/gocodebox/lifterlms-blocks/pull/153

## How has this been tested?
manually

## Types of changes
Bug fix (non-breaking change which fixes an issue/New feature (non-breaking change which adds functionality)

## Checklist:


includes/functions/llms-functions-forms.php
```diff
@@ -83,11 +84,24 @@ function llms_get_form_html( $location, $args = array() ) {
...
+		?
+		apply_filters( 'the_title', get_post_meta( $post->ID, '_llms_form_title_free_access_plans', true ) )
+		:
+		(
+			$show_title
+			?
+			get_the_title( $post->ID )
+			:
+			''
+		);
```
thomasplevy: I find nested ternary's extremely difficult to read and I'd just prefer we avoid them... I think in this case readability is preferable to increased conditions or longer code... Maybe I'm the only one but every time there's a nested ternary I find myself having to read it a dozen times to understand what it's doing

includes/functions/llms-functions-forms.php
```diff
@@ -83,11 +84,24 @@ function llms_get_form_html( $location, $args = array() ) {
...
 		return '';
 	}
 
-	return get_the_title( $post->ID );
+	$is_free_ap = 'checkout' === $location && isset( $args['plan'] ) && $args['plan']->is_free();
+	$show_title = $is_free_ap || llms_parse_bool( get_post_meta( $post->ID, '_llms_form_show_title', true ) );
+
+	return $is_free_ap
+		?
+		apply_filters( 'the_title', get_post_meta( $post->ID, '_llms_form_title_free_access_plans', true ) )
```
thomasplevy, line 93: We're going to run into some potential theme conflicts with this: https://github.com/WordPress/twentytwentyone/blob/ba9f20cad89163761185c0467b346ba42541ae22/inc/template-functions.php#L163-L177

Currently, they can disable the form title entirely if they want it to be blank but by passing the meta through the filter and using a "Leave blank to disable" style field then we're going to be unable to disable the title on certain themes for free access plans only. I don't know how prevalent functionality like this is in themes but it's something to consider...

I don't love having the "Leave blank to disable" but it does make it so it's independent of the form title for non-free access plans...

I think alternatively we could have two options, which feels a bit cleaner to me anyway.

A "Show Form Title" option (for default) and then a "Show Free Access Plan Form Title"

The new bool option can then toggle the presence of this new text field and in order to show a free access plan with no title the toggle should be turned off.


Or we can use the initial "Show Form Title" toggle for both options, which I actually think will be okay long term. I think most people are going to want to either disable or have different titles, not disable one but not the other... I think at least....

thomasplevy (changes requested):

includes/functions/llms-functions-forms.php
eri-trabiccolo: ok

includes/functions/llms-functions-forms.php
eri-trabiccolo: Ok I tie it to the `Show Form Title` then.

thomasplevy (changes requested):
Needs changelog :+1:

eri-trabiccolo:
I'm not sure what kind of changelog entry is this, it's either a patch/fixed (because fixes a regression introduced in 5.0) and an enhancemen/feature because it allows a customization that wasn't really possible (at least not that easy) in pre 5.0 as well.

PR #2017, opened 2022-02-22
Set core/columns to not stack on mobile by default for certificate posts

### LINKED ISSUE #1972: [6.0] Certificate's inner elements are responsive and probably they should be not
### Reproduction Steps

+ Create a certificate template with the default content
+ View it
+ emulate a mobile device

### Expected Behavior

+ I would expect it to be not responsive, e.g. the columns not stacked, or the font-size kept as in desktop viewport

### Actual Behavior

+ The certificate's inner elements are responsive

### Error Messages / Logs

n/a

+ [x] Locally
+ [ ] On a staging site
+ [ ] On a production website

+ Browser name and version

thomasplevy:
## Description

Closes #1972 

This PR modifies all `core/columns` block variations during block registration (client-side) to have `isStackedOnMobile` set to `false`.

Additionally passes this value in the default certificate block template.

This, effectively, makes it so that all columns added to a certificate will not stack by default until a user opts into stacking. They have the power to do so but they'll only do so if they choose.

I have tried ever filter found on [this page](https://developer.wordpress.org/block-editor/reference-guides/filters/block-filters/) to simply change the default value (via either PHP or JS) so that we don't need any fancy hacks but no combination I can find works without through a block validation error in the block editor. The validation error seems to be in error as the stored HTML is correct with most of the filters I've attempted. But... I can't figure out the chain of events that ultimately cause that validation error.

I cannot locate a way to disable the stacking toggle in the block's editor inspector. The only way I can figure out how to do it would be to redefine the `Edit()` element for the core block. This is possible but would make maintenance difficult long term. I'd rather avoid this.

I also tried creating a custom block (llms/certificate-columns) that leverages code from the core block (as we did for the lllms/certificate-title block, using the core/heading block as the base) but this block is quite a bit more complicated and I found myself having to copy too much logic again, making maintenance difficult.

### DIFF
--- a/includes/schemas/llms-block-templates.php
+++ b/includes/schemas/llms-block-templates.php
@@ -60,7 +60,9 @@ $certificates = array(
 	),
 	array(
 		'core/columns',
-		array(),
+		array(
+			'isStackedOnMobile' => false,
+		),
 		array(
 			array(
 				'core/column',
--- a/src/js/admin-certificate-editor/index.js
+++ b/src/js/admin-certificate-editor/index.js
@@ -6,6 +6,7 @@ import './editor';
 import './i18n';
 import './merge-codes';
 import './migrate';
+import './modify-blocks';
 import CertificateDocumentSettings from './document-settings';
 import CertificateResetTemplate from './reset-template';
 import CertificateUserSettings from './user-settings';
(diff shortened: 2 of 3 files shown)


eri-trabiccolo (changes requested):
you nailed it!

Although my boss would have added:
> fix lint

I also propose to add three new issue labels:
Has a fuckin' PR
Has a badass PR
Has a fuckin' badass PR

This PR is a fuckin' one, although not a badass one, do you agree?

Issue #1918, opened 2022-01-04
PHP & WP Support Drop


thomasplevy:
In conjunction with the release of (and added support for) PHP 8.1, LifterLMS will drop support for PHP 7.3. [LifterLMS maintains support for all PHP versions which are receiving active security support](https://lifterlms.com/docs/minimum-system-requirements-lifterlms/#supported-versions) and [PHP 7.3s EOL was reached in Decemeber 2021](https://www.php.net/eol.php).

The supported PHP versions will changed to 7.4, 8.0, and 8.1 with our recommended version adjusting from 7.4 to 8.0.

With support for WordPress 5.9, we will drop Support for WP 5.4. We support the latest 5 versions of WordPress, which will change from 5.4 to 5.8 to 5.5 to 5.9

---

+ [ ] Update docs: https://lifterlms.com/docs/minimum-system-requirements-lifterlms/#supported-versions
+ [ ] Update build matrix: https://github.com/gocodebox/.github/blob/trunk/.github/actions/setup-matrix/src/defaults.json
+ [ ] Update versions in plugin headers
+ [ ] Add version support warning in the next LifterLMS core release

PR #2020, opened 2022-02-23
Update minimum supported PHP version to 7.4

pondermatic:
## Description
Updated the minimum supported PHP version from 7.3 to 7.4.

See #1918.

## Checklist:


thomasplevy:
At this moment I'm not confident enough people are off 7.3

Let's wait another month or two

We bumped assignments to 7.4 and immediately received complaints from users whou couldn't upgrade. With the 6.0 changes imminent I don't want to create any additional havoc

eri-trabiccolo:
@pondermatic 
from 7.3 to 7.4
:P 

also
@gocodebox/engineering we dropped 7.3 automatic testing https://github.com/gocodebox/.github/blob/a7d6ad273077163c90872c5e8f73e121eef2196d/.github/actions/setup-matrix/src/defaults.json
Shall we re-introduce it until we update the min support version to 7.4? I think so.

thomasplevy:
@eri-trabiccolo I don't want to *reintroduce* official support. What I am suggesting is that we don't bump the official version in the plugin for another little while. My reasoning is that A) We don't currently use any features that aren't available in 7.3 B) As soon as we bump min support, the support team will get lots of emails from unhappy users who can't upgrade and don't understand why and will think that *we've made a mistake*. I want to reduce the number of emails that they have to field about this kind of stuff for a bit longer.

eri-trabiccolo:
@thomasplevy 
I understand and I agree on not bumping the minimum PHP version.
What I'm saying is that if we keep the plugin's PHP supported version set to 7.3 would be better to reintroduce the automatic tests against it to avoid shipping something with any features that aren't available in 7.3. That's exactly what tests agains php version are for, am I wrong?

thomasplevy:
@eri-trabiccolo I am ignoring what is the correct answer but I just don't have the energy or time to add 7.3 back. This wasn't on anyone's radar and now it's more work for me that I don't have time to do. I want us focusing on ensuring 6.0 is working well... Not reintroducing more tested versions because I preemptively removed them and then said "pause" on publicly bumping the version.

PR #2022, opened 2022-02-24
Fix certificate {sequential_id} only incremented when programmatically inserting an awarded certificate

### LINKED ISSUE #2021: [6.x] certificate {sequential_id} only increments when programmatically inserting an awarded certificate
Currently not easily reproducible via the UI but there's a design flaw in:

https://github.com/gocodebox/lifterlms/blob/c3e2b51ad322bb6bb60ab95a642dd9ed88ec545f/includes/models/model.llms.user.certificate.php#L108-L119

If we create a new awarded certificate via the REST API and set it to have a parent, this code won't run and therefore the sequential ID won't be set properly.

Changes I'm working on for #1965 make this apparent.

I propose that this code be moved out of the model and into a controller class, probably to run here:

https://github.com/gocodebox/lifterlms/blob/c3e2b51ad322bb6bb60ab95a642dd9ed88ec545f/includes/controllers/class-llms-controller-awards.php#L143-L153

eri-trabiccolo:
## Description
Fixes #2021 

## How has this been tested?
manually and with unit tests

## Types of changes

## Checklist:

### DIFF
--- a/includes/models/model.llms.user.certificate.php
+++ b/includes/models/model.llms.user.certificate.php
@@ -106,18 +106,24 @@ class LLMS_User_Certificate extends LLMS_Abstract_User_Engagement {
 	}
 
 	/**
-	 * Maybe set this awarded certificate sequential id based on the parent's meta.
+	 * Set this awarded certificate sequential id based on the parent's meta.
 	 *
 	 * @since [version]
 	 *
-	 * @return void
+	 * @return int|false Returns the awarded certificate sequenatial id.
+	 *                   Returns false if the awarded certificate has no parent template.
 	 */
-	public function maybe_set_sequential_id() {
+	public function update_sequential_id() {
 
 		$parent = $this->get( 'parent' );
-		if ( $parent ) {
-			$this->set( 'sequential_id', llms_get_certificate_sequential_id( $parent, true ) );
+		if ( ! $parent ) {
+			return false;
 		}
+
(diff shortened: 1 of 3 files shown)


includes/models/model.llms.user.certificate.php
```diff
@@ -106,16 +106,19 @@ public function __construct( $model, $args = array() ) {
...
 	/**
-	 * Called immediately after creating / inserting a new post into the database
+	 * Maybe set this awarded certificate sequential id based on the parent's meta.
 	 *
 	 * @since [version]
 	 *
 	 * @return void
 	 */
-	protected function after_create() {
+	public function maybe_set_sequential_id() {
```
thomasplevy: I'm being nitpicky but I don't like the idea of a "maybe" setter... I know we use the "maybe_do_x" pattern quite a bit, but usually in a callback function for a hook/filter, not in a function that is part of the truly public API.

My suggestion here would be to change this to `update_sequential_id()` and add a return to the function so that if there's no parent, it returns `false` 

If there *is* a parent, it then get's the next sequential ID, sets it, and then returns it.

Does this seem stupid and nitpicky?

thomasplevy (changes requested):
This looks really good other than my nitpicky request to change the method name

includes/models/model.llms.user.certificate.php
eri-trabiccolo: No it's fine. ❤️

PR #2027, opened 2022-02-24
Cert award from template

thomasplevy:
## Description

Requires #2026 (submitted as a PR against taht branch for now to reduce codesize for review)
Requires https://github.com/gocodebox/lifterlms-blocks/pull/158

Closes #1965

Adds a button to:

1) Certificate Template block editor sidebar
2) Awarded certificate post table (replaces the default "Add New Button" provided via PHP by the WP core
3) LLMS Student Certificates reporting screen

When pressed, the button pops a modal allowing student and / or template selection based on the context

In 1): student can be selected, template cannot (template is the current template being edited) and the "start from scratch" button is disabled

In 2): both the template and the student can be selected. Start from scratch exists and does not pre-select a student

In 3) template can be selected, not student (student is the current student being veiwed). Start from scratch prefills the current student using query string vars

The only thing I'm uncertain about is the way we change the UI on the post table screen. The button looks *similar* but not identical. Had I thought about this before I started I would have separated this into multiple components so the modal could be popped independently based on the pressing of an existing dom button... It's not too late to do this but I wanted feedback before reworking too much.

## How has this been tested?

+ Manually
+ New tests (php and js unit)
+ I will be writing additional e2e tests tomorrow

## Types of changes
New feature

## Checklist:
- [x] My code has been tested.
- [x] My code passes all existing automated tests. 
- [x] My code follows the LifterLMS Coding & Documentation Standards.

PR #2028, opened 2022-02-25
Components: fix post search with terms containing spaces

eri-trabiccolo:
fixes https://github.com/gocodebox/lifterlms/pull/2027#pullrequestreview-893561896 bullet 2.

Looks like `addQueryArgs` already encodes the arguments, so that an argument like `Foo Bar` becomes `Foo%25%20Bar` instead of the expected `Foo%20Bar`.

This PR fixes that.


thomasplevy:
Looks good thanks <3

PR #2035, opened 2022-03-02
Added set of options to define custom default certificate size

### LINKED ISSUE #2029: [6.0] Default Custom Certificate Size
Create UI for a new selectable size in default certificates (LifterLMS->Engagements) that allows you to still create a custom size per certificate that way if a user wants to globally change their custom size they can do it once under their default settings.

eri-trabiccolo:
## Description

Fixes #2029

## How has this been tested?
manually and with new unit tests

## Types of changes

## Checklist:

### DIFF
--- a/includes/models/model.llms.user.certificate.php
+++ b/includes/models/model.llms.user.certificate.php
@@ -752,13 +752,6 @@ class LLMS_User_Certificate extends LLMS_Abstract_User_Engagement {
 		$default_size                    = get_option( 'lifterlms_certificate_default_size', 'LETTER' );
 		$this->property_defaults['size'] = ! $default_size ? 'LETTER' : $default_size;
 
-		// Custom size.
-		if ( llms_parse_bool( get_option( 'lifterlms_certificate_default_custom_size', false ) ) ) {
-			$this->property_defaults['size']   = 'CUSTOM';
-			$this->property_defaults['width']  = get_option( 'lifterlms_certificate_default_custom_width', 8.5 );
-			$this->property_defaults['height'] = get_option( 'lifterlms_certificate_default_custom_height', 11 );
-			$this->property_defaults['unit']   = get_option( 'lifterlms_certificate_default_custom_unit', 'in' );
-		}
 	}
 
 	/**
--- a/includes/models/model.llms.user.certificate.php
+++ b/includes/models/model.llms.user.certificate.php
@@ -319,7 +319,7 @@ class LLMS_User_Certificate extends LLMS_Abstract_User_Engagement {
 	 * @return array[] Array of fonts by the certificate. Each array is a font definition with the font's
 	 *                 id added to the array.
 	 */
-	public function get_custom_fonts( $blocks = null ) {
+	public function get_default_custom_fonts( $blocks = null ) {
 
 		$fonts = array();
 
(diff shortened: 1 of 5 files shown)


eri-trabiccolo:
about https://github.com/gocodebox/lifterlms/issues/2029#issuecomment-1055624778
I actually decided to use a checkbox rather than a new select option so to control the visibility of the height/width/unit settings, using the already existing _api_

includes/models/model.llms.user.certificate.php
```diff
@@ -319,7 +319,7 @@ public function get_background_image() {
 	 * @return array[] Array of fonts by the certificate. Each array is a font definition with the font's
 	 *                 id added to the array.
 	 */
-	public function get_custom_fonts( $blocks = null ) {
+	public function get_default_custom_fonts( $blocks = null ) {
```
thomasplevy: What's up with this name change here?

If this is somehow an essential name change (which I don't think it is) the associated tests need to be updated as well.

Additionally, this is used here and should be updated as well: https://github.com/gocodebox/lifterlms/blob/8e9c2293977a14950275b3b9ca0d2284d051b804/includes/functions/llms.functions.templates.certificates.php#L56

includes/models/model.llms.user.certificate.php
```diff
@@ -752,6 +752,13 @@ private function set_property_defaults() {
 		$default_size                    = get_option( 'lifterlms_certificate_default_size', 'LETTER' );
 		$this->property_defaults['size'] = ! $default_size ? 'LETTER' : $default_size;
 
+		// Custom size.
+		if ( llms_parse_bool( get_option( 'lifterlms_certificate_default_custom_size', false ) ) ) {
+			$this->property_defaults['size']   = 'CUSTOM';
+			$this->property_defaults['width']  = get_option( 'lifterlms_certificate_default_custom_width', 8.5 );
+			$this->property_defaults['height'] = get_option( 'lifterlms_certificate_default_custom_height', 11 );
+			$this->property_defaults['unit']   = get_option( 'lifterlms_certificate_default_custom_unit', 'in' );
+		}
```
thomasplevy: My vision for this was that instead of replacing the `CUSTOM` definition with these default dimensions, it would register a new size definition using this filter: https://github.com/gocodebox/lifterlms/blob/8e9c2293977a14950275b3b9ca0d2284d051b804/includes/functions/llms.functions.certificate.php#L312-L364

The size would be `USER_DEFINED` which would mean the `CUSTOM` (that gets added outside this filter because it's not a defined / registered size) would still exist.

Your implementation does allow the custom size to be redefined per-certificate but in doing so it gets "detached" from the custom definition. If a user wishes to change their custom definition later, then they can't.

You've created a *default* value for the `CUSTOM` definition whereas I was planning for a UI that allows registration of a single custom size.

I also don't like the "leave checked to disable the default above" which is just kind of confusing to the user. I think if the `USER_DEFINED` is added (with pre-defined default values) and then users can use the settings below to modify the size settings it would be less confusing.

What do you think?

thomasplevy (changes requested):

includes/models/model.llms.user.certificate.php
eri-trabiccolo: sorry this is due to a S/R
good catch :P

includes/models/model.llms.user.certificate.php
eri-trabiccolo: Oh ok I get what you mean.
you're really smart!

includes/models/model.llms.user.certificate.php
thomasplevy: mmm hmmm :+1:  <3

eri-trabiccolo:
what do you think now?

thomasplevy:
I'm happy with this, nice work. I'm adding inline JS and updating some language directly on dev-600 after merging.

PR #2084, opened 2022-03-28
Change llmsStudentsSelect2() to use the LifterLMS REST API

pondermatic:
## Description
Changed the `llmsStudentsSelect2()` JavaScript function to use the [LifterLMS REST API "list students" endpoint](https://developer.lifterlms.com/rest-api/#tag/Students/paths/~1students/get) instead of the [`LLMS_AJAX_Handler::query_students()`](https://github.com/gocodebox/lifterlms/blob/6.1.0/includes/class.llms.ajax.handler.php#L398) PHP function.

Fixes LifterLMS Private Areas [60](https://github.com/gocodebox/lifterlms-private-areas/issues/60).

## How has this been tested?
Manually:
1. Add new private post. Select a student. Optionally select a course.
2. LifterLMS -> Reporting -> Sales -> Toggle Filters. Select students.

Checked JavaScript console and PHP error log.

## Types of changes

## Checklist:

### DIFF
--- a/.changelogs/deprecate-ajax-query-students.yml
+++ b/.changelogs/deprecate-ajax-query-students.yml
@@ -1,3 +1,3 @@
 significance: minor
 type: deprecated
-entry: Deprecated `LLMS_AJAX_Handler::query_students()`. Use LifterLMS REST API 'GET https://example.tld/wp-json/llms/v1/students' instead.
+entry: Deprecated `LLMS_AJAX_Handler::query_students()`. Use the [REST API list students](https://developer.lifterlms.com/rest-api/#tag/Students/paths/~1students/get) endpoint instead.
--- a/.changelogs/deprecate-ajax-query-students.yml
+++ b/.changelogs/deprecate-ajax-query-students.yml
@@ -1,0 +1,3 @@
+significance: minor
+type: deprecated
+entry: Deprecated `LLMS_AJAX_Handler::query_students()`. Use LifterLMS REST API 'GET https://example.tld/wp-json/llms/v1/students' instead.
--- a/assets/js/llms-admin.js
+++ b/assets/js/llms-admin.js
@@ -2,7 +2,7 @@
  * LifterLMS Admin Panel Javascript
  *
  * @since Unknown
- * @version 4.5.1
+ * @version [version]
  *
  * @param obj $ Traditional jQuery reference.
  * @return void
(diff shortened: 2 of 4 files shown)


assets/js/llms-admin.js
```diff
@@ -183,34 +185,37 @@
...
 						enrolled_in: params.enrolled_in || options.enrolled_in,
 						roles: params.roles || options.roles,
-						term: params.term,
+						search: params.term,
+						search_columns: 'email,name,username',
 					};
 				},
 				processResults: function( data, params ) {
+					var page       = params.page || 1;
+					var totalPages = this._request.getResponseHeader( 'X-WP-TotalPages' );
```
pondermatic, line 205: I did not find a "public" way to get the jQuery ajax object, so I used `_request`.

thomasplevy (changes requested):
Looks good to me, tested it in a few different scenarios and seems to work as expected. I think it might be a bit faster than the old method.

Only change I'd like to see made here is that we also deprecated the AJAX query method that this is replacing: https://github.com/gocodebox/lifterlms/blob/2982a141376d129a8d63f00214e00163196e3842/includes/class.llms.ajax.handler.php#L382-L597 

It is a public method so we shouldn't just remove it, but marking it as deprecated and adding a `_deprecated_function` call would be ideal so that we don't forget to do it later.

.changelogs/deprecate-ajax-query-students.yml
```diff
@@ -0,0 +1,3 @@
+significance: minor
+type: deprecated
+entry: Deprecated `LLMS_AJAX_Handler::query_students()`. Use LifterLMS REST API 'GET https://example.tld/wp-json/llms/v1/students' instead.
```
thomasplevy:
```suggestion
entry: Deprecated `LLMS_AJAX_Handler::query_students()`. Use the [REST API list students](https://developer.lifterlms.com/rest-api/#tag/Students/paths/~1students/get) endpoint instead.
```

includes/class.llms.ajax.handler.php
```diff
@@ -392,11 +392,15 @@ public static function membership_remove_auto_enroll_course( $request ) {
 	 * @since 3.14.2 Unknown.
 	 * @since 5.5.0 Do not encode quotes when sanitizing search term.
 	 * @since 5.9.0 Stop using deprecated `FILTER_SANITIZE_STRING`.
+	 * @deprecated [version] `LLMS_AJAX_Handler::query_students()` is deprecated. Use LifterLMS REST API
```
thomasplevy:
```suggestion
	 * @deprecated [version] `LLMS_AJAX_Handler::query_students()` is deprecated in favor of the REST API list students endpoint.
```

includes/class.llms.ajax.handler.php
```diff
@@ -392,11 +392,15 @@ public static function membership_remove_auto_enroll_course( $request ) {
...
+	 *                   'GET https://example.tld/wp-json/llms/v1/students' instead.
 	 *
 	 * @return void
 	 */
 	public static function query_students() {
 
+		_deprecated_function( __METHOD__, '[version]', 'GET https://example.tld/wp-json/llms/v1/students' );
```
thomasplevy:
```suggestion
		_deprecated_function( __METHOD__, '[version]', 'the REST API list students endpoint' );
```

thomasplevy (changes requested):
I'm nitpicking the language but I don't like the idea of having example URLs as the "replacement" in our docs / changelogs. Even having those examples in the REST API docs occassionally result in people not realizing that "example.tld" isn't meant to be replaced with their actual domain.

PR #2102, opened 2022-04-07
AJAX order creation

### LINKED ISSUE #2173: Implement new gateway feature: `modify_recurring_payments`
See related: https://github.com/gocodebox/lifterlms-gateway-paypal/issues/52

## Goal

Allow payment gateways to determine if recurring payment details (such as trial end date, next payment date, etc...) can be modified via the admin panel order UI.

The new gateway feature can be added to the [supports](https://github.com/gocodebox/lifterlms/blob/81b1fff19d7b8a97e6a5d2780f800212050c6e1f/includes/abstracts/abstract.llms.payment.gateway.php#L101-L109) interface for a new gateway feature, probably called "modify_recurring_payments".

I think that this feature should be enabled by default *when* `recurring_payments` is `true`. We don't currently have a concept of a conditional feature so we'll probably need to reengineer the existing supports/feature functionality to accommodate a conditional feature default.

Ideally this should be implemented in a way that *does not* mess with existing gateways (which can edit by default)

thomasplevy:
## Description

+ Closes #2173

This PR ~attempts to~ constructs an alternate checkout strategy which uses AJAX to setup a pending order *without* creating a user.

~Gateways will have to opt into utilizing this method. If they don't specify any support for this method they'll continue to use the previous strategy.~

Gateways will continue to work using the synchronous form submission. If a gateway chooses to utilize this new set of AJAX endpoints they will do so by utilizing JS to submit the checkout form (in favor of defaulting to PHP form submission). I've tested this PR against the latest version of LifterLMS Stripe to ensure nothing has broken as a result (and I don't see any problems, though I need to setup tests to run the automated test suite as a set of second checks before we commit to shipping this).

This will be utilized by the new PayPal Commerce Platform gateway (actively in development).

~At this moment, this PR is wholly incomplete but opened for early (mostly theoretical) review on the strategy and the code that's written thus far (and pretty well tested)~

## How has this been tested?

+ Manually 

## Types of changes
New feature (backwards compatible)

## Checklist:


includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,597 @@
+<?php
```
thomasplevy, line 1: The purpose of this class is to handle all of the logic currently in [llms_setup_pending_order()](https://developer.lifterlms.com/reference/functions/llms_setup_pending_order/)

This class is more atomic and testable(tested) than the massive setup function.

Additionally, it provides an option to either "commit" (save user information) or just validate it.

includes/forms/class-llms-form-handler.php
```diff
@@ -331,6 +335,10 @@ public function submit( $posted_data, $location, $args = array() ) {
 		// Make sure the user id cannot be forced by user submission.
 		unset( $posted_data['user_id'] );
 
+		if ( ! empty( $args['validate_only'] ) ) {
+			return $this->validate_fields( $posted_data, $location, $fields, $action );
+		}
+
```
thomasplevy, line 341: I've explored a few different ways to enable separating validation from the form submission (so that we can validate all the form data without actually processing it) and this seems to be a pretty good way to do it. It doesn't require much new code to be written or modified. I do worry that adding it as an argument option buries it a little but it's clear enough (and I added the `llms_validate_user()` function to expose it less densely as well.

includes/models/model.llms.order.php
```diff
@@ -1318,36 +1318,25 @@ public function has_trial_ended() {
...
 	 * @since 3.8.0
 	 * @since 3.10.0 Unknown.
 	 * @since 5.3.0 Don't set unused legacy property `date_billing_end`.
+	 * @since [version] Use `LLMS_Order::set_user_data()` to update user data.
 	 *
-	 * @param LLMS_Student         $person  The LLMS_Student placing the order.
-	 * @param LLMS_Access_Plan     $plan    The purchase LLMS_Access_Plan.
-	 * @param LLMS_Payment_Gateway $gateway The LLMS_Payment_Gateway used.
-	 * @param LLMS_Coupon          $coupon  LLMS_Coupon if a coupon was used or false.
+	 * @param array|LLMS_Student|WP_User|integer $user_data User info for the person placing the order. See
```
thomasplevy, line 1426: Passing in an *array* of information in favor of a fully registered user object is the crux of all of this... This allows us to save everything to the order so we can still create a pending order, but doesn't create the user (and possibly more importantly trigger user registration hooks which do things like send emails)

There are interface changes that will need to be made to accommodate an order that doesn't have a user ID.

Additionally, I have opted to *not* save custom field data or passwords to the order.

I considered doing this as a way to "resume" checkout later but the idea here is to complete everything in the same session. Based on how I'm building things (that I haven't pushed yet) it won't be necessary to reference this stuff from the order for user creation later as it'll still be available for future AJAX requests to confirm or complete the order.

thomasplevy:
@gocodebox/engineering requesting an early review on strategy, mostly.

includes/forms/class-llms-form-handler.php
eri-trabiccolo: I wonder if the `$action` above (used in the hook names) should be changed to adding `_validate` when validating only.
Like 
```php
		if ( ! empty( $args['validate_only'] ) ) {
			$action .= '_validate';
			return $this->validate_fields( $posted_data, $location, $fields, $action );
		}
```
?

includes/models/model.llms.order.php
eri-trabiccolo: Ok for the reasoning "everything will be completed in the same session" but still, if there's a way to "resume", then custom fields should be saved to me. Why cannot they?

includes/models/model.llms.order.php
```diff
@@ -1436,7 +1425,27 @@ public function init( $person, $plan, $gateway, $coupon = false ) {
...
+		 * @param LLMS_Order   $order                           The order object.
+		 * @param LLMS_Student $student                         The student object. If an array of data is passed
+		 *                                                      to `LLMS_Order::init()` then an empty student object
+		 *                                                      will be passed.
+		 * @param array|LLMS_Student|WP_User|integer $user_data User data.
+		 */
+		do_action(
+			'lifterlms_new_pending_order',
+			$this,
+			is_array( $user_data ) ? new LLMS_Student( -1 ) : llms_get_student( $user_data ),
```
eri-trabiccolo: perplexed about this `-1` considering that it will populate the `LLMS_Student::$id`, and `LLMS_Abstract_User_Data::get_user_id()` returns and `absint($id)`, so it'll return `1` => usually the first admin mmm.

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,597 @@
...
+	/**
+	 * User Action: validate and then commit (register or update) the user.
+	 *
+	 * @var string
+	 */
+	const UA_COMMIT = 'commit';
+
+	/**
+	 * User Action: perform user validation only.
+	 * @var string
```
eri-trabiccolo:
```suggestion
	 * User Action: perform user validation only.
	 *
	 * @var string
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,597 @@
...
+	 * Returns an error object.
+	 *
+	 * This method accepts an error code and message and passes them directly to `WP_Error` and
+	 * adds all class variables to the error objects `$data` parameter.
+	 *
+	 * @since [version]
+	 *
+	 * @param string $code      Error code.
+	 * @param string $message   Error message.
+	 * @param array $extra_data Additional data to pass to WP_Error's 3rd parameter.
```
eri-trabiccolo:
```suggestion
	 * @param string $code       Error code.
	 * @param string $message    Error message.
	 * @param array  $extra_data Additional data to pass to WP_Error's 3rd parameter.
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,597 @@
...
+		$validations = array(
+			'validate_plan',
+			'validate_coupon',
+			'validate_gateway',
+			'validate_terms',
+			'validate_user',
+		);
+
+		foreach ( $validations as $func ) {
+			$res = $this->{$func}();
```
eri-trabiccolo, line 528: oh my lord, someone finally stopped using `call_user_func(_array)`.

eri-trabiccolo (changes requested):
ok it looks promising :P

includes/models/model.llms.order.php
thomasplevy: good catch... `-1` ensures the `WP_User` object isn't found. There's no way to construct an empty object otherwise as the current user ID would be used by default.

I'll have to do some refactoring on the student class to make that work the way I wanted it to...

Really this is for backwards compatibility on the hook so maybe a better option would be to just deprecate the filter and add a replacement that functions differently

includes/class-llms-order-generator.php
thomasplevy: There's still time to change this, tread carefully.

includes/models/model.llms.order.php
thomasplevy: We *can* store all the custom fields but in doing so it will require us to create a method for doing so which won't be strictly necessary given the way this is being built.

Essentially it feels like we'd be storing data for the sake of doing so instead of because it's required / necessary.

The real problem I have here too is with passwords... I know it's possible but we'll need to figure out how to properly hash/store the password into the order so then later we can pull it off the order and store it again (but without hashing it this time because it's already hashed). I'm sure this is possible but I haven't worked through how to do it. And I don't think this is essential for any of the existing payment gateways we're working with... Maybe I'm being short-sighted here but it feels like a lot of code that we're going to have to write in order to solve a problem that we don't currently have. I'm looking primarily at PayPal where there is a 2-step process.

The real concern would be a network interruption scenario. In this case the user is going to need to start on step one and re-enter their password.

includes/models/model.llms.order.php
thomasplevy: Addressed in 1666a47d08928dbecdc06db7e679308435ba355a

I think this is a more sane way of accomplishing this. Thoughts?

thomasplevy:
@gocodebox/engineering this is no longer a WIP and is ready for a full review and testing.

thomasplevy:
@gocodebox/engineering we're going to need to ship this within the next week or two at the latest. I know it's a large changeset so I anticipate more working needing to be done. I don't want this PR to turn into a blocker on the first internal PayPal alpha so let's get to reviewing. Thank you!

includes/abstracts/abstract.llms.payment.gateway.php
```diff
@@ -178,6 +197,37 @@ public function complete_transaction( $order, $deprecated = null ) {
...
+		$data = wp_parse_args(
+			$data,
+			array(
+				'redirect' => $this->get_complete_transaction_redirect_url( $order ),
+				'status'   => 'SUCCESS',
+			)
+		);
+
+		// Ensure notification processors get dispatched since shutdown wont be called.
+		do_action( 'llms_dispatch_notification_processors' );
```
eri-trabiccolo, line 226: This action hook should be documented in `complete_transaction()` and add a reference here like:
`This action is documented in complete_transaction()` or something.

includes/controllers/class-llms-controller-checkout.php
```diff
@@ -0,0 +1,404 @@
+<?php
+/**
+ * LLMS_Controller_Checkout
+ *
+ * @package LifterLMS/Controllers/Classes
+ *
+ * @since 3.0.0
+ * @version 5.9.0
```
eri-trabiccolo:
```suggestion
 * @since [version]
 * @version [version]
```

includes/controllers/class-llms-controller-checkout.php
```diff
@@ -0,0 +1,404 @@
...
+			$user_data['user_id'] = $uid;
+		}
+
+		foreach ( LLMS_Forms::instance()->get_form_fields( 'checkout', compact( 'plan' ) ) as $field ) {
+			if ( isset( $posted_data[ $field['name'] ] ) ) {
+				$user_data[ $field['name'] ] = $posted_data[ $field['name'] ];
+			}
+		}
+
+		return $user_data;
```
eri-trabiccolo, line 354:
```suggestion
		$plan      = $plan_id ? llms_get_post( $plan_id ) : false;
		$user_data = array_intersect_key( 
			$posted_data, 
			array_flip( array_column( LLMS_Forms::instance()->get_form_fields( 'checkout', compact( 'plan' ) ), 'name' ) )
		);

		$uid = get_current_user_id();
		if ( $uid ) {
			$user_data['user_id'] = $uid;
		}

		return $user_data;
```
what do you think about this golf?

includes/controllers/class-llms-controller-checkout.php
```diff
@@ -0,0 +1,404 @@
...
+
+	}
+
+	/**
+	 * Retrieves the AJAX url for the requested action
```
eri-trabiccolo:
```suggestion
	 * Retrieves the AJAX url for the requested action.
```

includes/controllers/class-llms-controller-checkout.php
```diff
@@ -0,0 +1,404 @@
...
+	 *
+	 * @since [version]
+	 *
+	 * @param array|WP_Error $data Response data.
+	 * @return void
+	 */
+	private function send_json( $data ) {
+
+		// Tell WP we're doing AJAX.
+		add_filter( 'wp_doing_ajax', '__return_true' );
```
eri-trabiccolo, line 404: should we set this earlier? like in `create_pending_order_ajax()`? Why do we set this here?

includes/controllers/class-llms-controller-checkout.php
```diff
@@ -0,0 +1,404 @@
...
+	 */
+	private function verify_request( $field, $nonce ) {
+
+		if ( ! llms_verify_nonce( $field, $nonce, 'POST' ) ) {
+			return null;
+		}
+
+		if ( llms_filter_input( INPUT_POST, 'action' ) !== $nonce ) {
+			return false;
+		}
```
eri-trabiccolo, line 610: I'm a bit confused by this, semantically speaking.
I think that when this condition is true: `( llms_filter_input( INPUT_POST, 'action' ) !== $nonce )` should be run before verifying the nonce, am I wrong?
Also in this case we should return `null` because it's like "there's nothing to do, you didn't even wanted to call me :P".
While when the action is correct BUT `! llms_verify_nonce( $field, $nonce, 'POST' )` then the request is incorrect, so return `false`.
What do you think?

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,732 @@
...
+	 * @var string
+	 */
+	const E_ORDER_NOT_FOUND = 'llms-order-gen-order-not-found';
+
+	/**
+	 * Error code: order cannot be confirmed.
+	 *
+	 * @var string
+	 */
+	const E_ORDER_NOT_CONFIRMABLE = 'llms-order-gent-order-not-confirmable';
```
eri-trabiccolo:
```suggestion
	const E_ORDER_NOT_CONFIRMABLE = 'llms-order-gen-order-not-confirmable';
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,732 @@
...
+
+	/**
+	 * Error code: access plan not found.
+	 *
+	 * @var string
+	 */
+	const E_PLAN_NOT_FOUND = 'llms-order-gen-plan-not-found';
+
+	/**
+	 * Error code: invalid coupon code submitted.
```
eri-trabiccolo:
```suggestion
	 * Error code: site's terms not accepted.
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,732 @@
...
+	 *     An associative array of input data used to generate the order, usually from $_POST.
+	 *
+	 *     @type integer $llms_plan_id         An LLMS_Access_Plan ID.
+	 *     @type string  $llms_agree_to_terms  A yes/no value determining whether or not the user has agreed to the site's terms.
+	 *     @type string  $llms_payment_gateway The ID of the payment gateway used to process the order.
+	 *     @type string  $llms_coupon_code     Optional. The coupon code string being used.
+	 *     @type string  $llms_order_key       Optional. An `LLMS_Order` key used to modify an existing pending order rather than creating a new one.
+	 *     @type array   ...$user_data         All remaining data is passed to the user creation functions.
+	 * }
+	 */
```
eri-trabiccolo:
```suggestion
	 * }
	 * @return void
	 */
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,732 @@
...
+		$args = array(
+			'plan' => $this->plan,
+		);
+
+		$user = get_current_user_id() ?
+			llms_update_user( $this->data, 'checkout', $args ) :
+			llms_register_user( $this->data, 'checkout', true, $args );
+
+		if ( ! is_wp_error( $user ) ) {
+			$this->student = llms_get_student( $user );
```
eri-trabiccolo: we cache the student here, why? Can't we just cache the WP_User?
Where are we going to use the student object?

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,732 @@
...
+				self::E_ORDER_NOT_FOUND,
+				__( 'Could not locate an order to confirm.', 'lifterlms' )
+			);
+		}
+
+		$order = llms_get_post( $order_id );
+		if ( ! $order->can_be_confirmed() ) {
+			return $this->error(
+				self::E_ORDER_NOT_CONFIRMABLE,
+				__( 'Could not locate an order to confirm.', 'lifterlms' )
```
eri-trabiccolo, line 643: same message as above?

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,732 @@
...
+	/**
+	 * Validates the site's terms and conditions were submitted.
+	 *
+	 * @since [version]
+	 *
+	 * @return boolean|WP_Error Returns `true` on success or an error object.
+	 */
+	protected function validate_terms() {
+
+		if ( llms_are_terms_and_conditions_required() && ! llms_parse_bool( $this->data['llms_agree_to_terms'] ?? 'no' ) ) {
```
eri-trabiccolo, line 687: Instead of something like this:
`$this->data['llms_agree_to_terms'] ?? 'no'`
maybe we can use `wp_parse_args()` in the class constructor so to merge passed data with defaults?

includes/forms/class-llms-form-handler.php
eri-trabiccolo: ^

includes/traits/llms-trait-award-templates-post-list-table.php
```diff
@@ -60,7 +60,7 @@ public function add_post_actions( $actions, $post ) {
 			add_query_arg(
 				array(
 					LLMS_Admin_Post_Table_Awards::TEMPLATE_FILTER_QUERY_VAR => $post->ID,
-					'post_type'                                             => $award_post_type,
+					'post_type' => $award_post_type,
```
eri-trabiccolo, line 63: stupid CS rule... this is buggy.

eri-trabiccolo (changes requested):

includes/controllers/class-llms-controller-checkout.php
thomasplevy: `llms_verify_nonce()` not only verifies the nonce but also verifies that the form has been submitted. Frankly it doesn't really matter which one runs first. Historically with form submission controllers I've followed a pattern (yes, it's not documented) where `null` is returned if the form wasn't submitted (in my mind this has always meant if the request method doesn't match and the nonce isn't found in the request payload). Returning `null` because the form wasn't submitted and also if the nonce is invalid.

The semantic difference between `null` and `false` really doesn't matter. I have different returns so I can easily differentiate the different conditions in testing. We could make the returns `1`, `2,` and `3` if we wanted. Or we could return `null` (or `false` for both conditions using an `||` operator but I wanted to be able to test each case independently.

eri-trabiccolo: alright you win

includes/class-llms-order-generator.php
thomasplevy: Yes it is. I don't think it needs to be different for the end user. But the error code can be used to programmatically differentiate between the conditions.

includes/class-llms-order-generator.php
thomasplevy: Yea but why?

eri-trabiccolo: to reduce the checks

includes/controllers/class-llms-controller-checkout.php
thomasplevy: mmm I'm not sure if it has much impact earlier though semantically I see that makes sense.

I put it here because [wp_send_json()](https://developer.wordpress.org/reference/functions/wp_send_json/) does `wp_die()` rather than `die()` making the function testable (because `wp_die()` can be easily mocked to not actually die)

I had initially had it after nonce verification in both the ajax callbacks but I moved it here to reduce redundant code. You think I should move it back?

includes/controllers/class-llms-controller-checkout.php
thomasplevy: yessss!

includes/forms/class-llms-form-handler.php
thomasplevy: I think this could be problematic because any action we'd expect to validate will now need to be hooked twice (once for the real action and again for the validate only action). Ideally when only validating we should have identical behavior to when validating immediately prior to registration/update. If we change the hookname that might not happen with 3rd party validations

includes/controllers/class-llms-controller-checkout.php
eri-trabiccolo: Mmm, if that's our only purpose, maybe you can just extend that comment to explain the reason that's there.
On the other hand I think that forcing that filter earlier should be also accompanied by a check on whether we're actually serving an ajax request, and we don't.

includes/forms/class-llms-form-handler.php
eri-trabiccolo: What if then `$action` is mixed => `string|array( string $name, bool $validate_only )` and we pass this `validate_only` as additional parameter to the hooks?

includes/class-llms-order-generator.php
thomasplevy: even if the default value is set in the constructor we still need to run this check though.

Reducing the need for null coalesce by running `wp_parse_args()` seems unnecessary to me... It's probably more performant this way, don't you think?

includes/class-llms-order-generator.php
thomasplevy: Because it's LifterLMS and the `get_student()` method returns a student object.

includes/class-llms-order-generator.php
eri-trabiccolo: Yeah, it's more performant, but having defaults set in one place makes me feel better. nevermind. :)

includes/class-llms-order-generator.php
eri-trabiccolo: I know, but it's less performant :P

includes/forms/class-llms-form-handler.php
thomasplevy: What's the requirement you're pushing here? 

I'm not sure there's any world in which we want the hooks to be able to respond differently based on whether or not we're validating for validation purposes only or for actually registering a user... You know what I mean?

The hooks should *always* respond the same...

Sure I can rewrite it to include the information but using that information to respond differently seems like creating a way to create a confusing experience...

Am I just being stubborn here? I don't think providing this info to hooks is at all necessary.

eri-trabiccolo: Because this hook, for example, is fired just before an user is created/updated, but now it can also be fired while only validating:
https://github.com/gocodebox/lifterlms/blob/9286df048b0063b499bff9a81cdc7271cc216d66/includes/forms/class-llms-form-handler.php#L483
And this seems odd to me, I'd expect that a user would be created right after but it won't.

includes/traits/llms-trait-award-templates-post-list-table.php
thomasplevy: I'm not going to try to rewrite the PHPCS rules to fix this. Thanks for your feedback.

thomasplevy: I'm not rewriting the linter or beautifier either. It's all coming frmo "upstream" in either WPCS or PHPCS itself. This barely matters.

includes/forms/class-llms-form-handler.php
thomasplevy: Okay... but still nothing has changed here. Previously this hook would have the same behavior, except that we didn't have a separate condition for *only validating*.

By this I mean if you as a developer *expect* that this hook is immediately followed by a user registration/update you're expecting something incorrectly because validation could still fail after this runs. 

Semantically sure the action name maybe doesn't make sense after this change but do we really need to provide every shred of internal context to every single hook we create?

In order to provide this context we need a SECOND action parameter to be passed to the function and in other contexts that parameter doesn't really make sense...

Adding a 5th parameter in order to pass this to a single action because the context *maybe* doesn't make sense seems like it's wholly unnecessary to me. I wa

includes/abstracts/abstract.llms.payment.gateway.php
thomasplevy: I think this action may have actually be deprecated during the 6.0 release and these were missed. I need to look further into this before committing to adding the documentation since it might actually just need to be removed.

includes/traits/llms-trait-award-templates-post-list-table.php
eri-trabiccolo: yeah sure, it was just a comment :P

includes/forms/class-llms-form-handler.php
eri-trabiccolo: > By this I mean if you as a developer expect that this hook is immediately followed by a user registration/update you're expecting something incorrectly because validation could still fail after this runs.

If you're talking about the specific hook `lifterlms_before_user_{action}`, yeah you're right, but if  you consider this other hook
https://github.com/gocodebox/lifterlms/blob/9286df048b0063b499bff9a81cdc7271cc216d66/includes/forms/class-llms-form-handler.php#L548
then it's not outrageous to expect a user registration/update after that.

Anyways ok, I see your reasons <3

includes/models/model.llms.order.php
```diff
@@ -357,6 +357,37 @@ private function calculate_trial_end_date( $format = 'Y-m-d H:i:s' ) {
...
+	/**
+	 * Determines if an order can be confirmed.
+	 *
+	 * An order can be confirmed only when the order's status is pending.
+	 *
+	 * Additional requirements can be introduced via the filter `llms_order_can_be_confirmed`.
+	 *
+	 * @since [version]
+	 *
+	 * @return boolaen
```
pondermatic:
```suggestion
	 * @return boolean
```

includes/models/model.llms.order.php
```diff
@@ -1436,7 +1456,27 @@ public function init( $person, $plan, $gateway, $coupon = false ) {
 			$this->set( 'access_period', $plan->get( 'access_period' ) );
 		}
 
-		do_action( 'lifterlms_new_pending_order', $this, $person );
+		/**
+		 * Action triggered after a the order is initialized.
```
pondermatic:
```suggestion
		 * Action triggered after the order is initialized.
```

includes/models/model.llms.order.php
```diff
@@ -1436,7 +1456,27 @@ public function init( $person, $plan, $gateway, $coupon = false ) {
...
+		 *
+		 * @since Unknown.
+		 * @since [version] Added `$user_data` parameter.
+		 *                 The `$student` parameter returns an "empty" student object
+		 *                 if the method's input data is an array instead of an existing
+		 *                 user object.
+		 *
+		 * @param LLMS_Order   $order                           The order object.
+		 * @param LLMS_Student $student                         The student object. If an array of data is passed
```
pondermatic:
```suggestion
		 * @param LLMS_Order                         $order     The order object.
		 * @param LLMS_Student                       $student   The student object. If an array of data is passed
```

includes/models/model.llms.order.php
```diff
@@ -1436,7 +1456,27 @@ public function init( $person, $plan, $gateway, $coupon = false ) {
...
+		 *                                                      to `LLMS_Order::init()` then an empty student object
+		 *                                                      will be passed.
+		 * @param array|LLMS_Student|WP_User|integer $user_data User data.
```
pondermatic, line 1537:
```suggestion
		 * @param array|LLMS_Student|WP_User|integer $user_data User info for the person placing the order.
		 *                                                      See {@see LLMS_Order::set_user_data()} for more info.
```

includes/models/model.llms.order.php
```diff
@@ -1703,6 +1743,80 @@ public function set_status( $status ) {
...
+	 * @since [version]
+	 *
+	 * @param array|LLMS_Student|WP_User|integer $user_or_data Accepts a raw array user meta-data or
+	 *                                                         an input string accepted by `llms_get_student()`.
+	 *                                                         When passing an existing user the data will be pulled
+	 *                                                         from the user metadata and saved to the order.
+	 * @return array {
+	 *     Returns an associative array representing the user metadata that was stored on the order.
+	 *
+	 *     @type integer $user_id            User's WP_User id.
```
pondermatic, line 1823:
```suggestion
	 *     @type integer $user_id            User's WP_User ID.
```
The abbreviation for "identity" is ID. 
"id" means "the one of the three divisions of the psyche in psychoanalytic theory that is completely unconscious and is the source of psychic energy derived from instinctual needs and drives".
https://www.merriam-webster.com/dictionary/id

includes/models/model.llms.order.php
```diff
@@ -1703,6 +1743,80 @@ public function set_status( $status ) {
...
+	 *     @type string  $user_ip_address    User's ip address.
```
pondermatic, line 1824:
```suggestion
	 *     @type string  $user_ip_address    User's IP address.
```

includes/models/model.llms.order.php
```diff
@@ -1703,6 +1743,80 @@ public function set_status( $status ) {
...
+
+			$map = array(
+				'user_id'            => 'id',
+				'billing_email'      => 'user_email',
+				'billing_phone'      => 'phone',
+				'billing_first_name' => 'first_name',
+				'billing_last_name'  => 'last_name',
+			);
+
+			foreach ( $to_set as $order_key => &$val ) {
```
pondermatic:
```suggestion
			foreach ( $to_set as $order_key => $val ) {
```
`$val` isn't used.

pondermatic:
The [LLMS_Payment_Gateway](https://github.com/gocodebox/lifterlms/blob/async-orders/includes/abstracts/abstract.llms.payment.gateway.php#L8) version tag needs updated.

includes/abstracts/abstract.llms.payment.gateway.php
```diff
@@ -178,6 +197,37 @@ public function complete_transaction( $order, $deprecated = null ) {
 
 	}
 
+	/**
+	 * This should be called by AJAX-powered gateways after verify a transaction was completed successfully.
```
pondermatic:
```suggestion
	 * This should be called by AJAX-powered gateways after verifying that a transaction was completed successfully.
```

includes/abstracts/abstract.llms.payment.gateway.php
```diff
@@ -178,6 +197,37 @@ public function complete_transaction( $order, $deprecated = null ) {
...
+
+		$data = wp_parse_args(
+			$data,
+			array(
+				'redirect' => $this->get_complete_transaction_redirect_url( $order ),
+				'status'   => 'SUCCESS',
+			)
+		);
+
+		// Ensure notification processors get dispatched since shutdown wont be called.
```
pondermatic:
```suggestion
		// Ensure notification processors get dispatched since shutdown won't be called.
```

includes/abstracts/abstract.llms.payment.gateway.php
pondermatic: I'm not finding any history of this action hook being deprecated. It's also not used in the LifterLMS core (I only searched back to LifterLMS 5.0.0).

includes/abstracts/abstract.llms.payment.gateway.php
```diff
@@ -570,14 +620,12 @@ public function get_logging_enabled() {
 	}
 
 	/**
-	 * Retrieves a list of "secure" strings which should be anonymized if they're found within debug logs.
+	 * Adds the gateways registered secured strings to the default list of site-wide secure strings.
```
pondermatic:
```suggestion
	 * Adds the gateway's registered secured strings to the default list of site-wide secure strings.
```

includes/abstracts/abstract.llms.payment.gateway.php
```diff
@@ -937,6 +959,49 @@ protected function get_option_prefix() {
...
+
+		}
+
+		/**
+		 * Filters the list of the gateway's secure strings.
+		 *
+		 * @since 6.4.0
+		 *
+		 * @param strings[] $gateway_strings List of secure strings for the payment gateway.
+		 * @param string    $id              The gateway ID.
```
pondermatic:
```suggestion
		 * @param string[] $gateway_strings List of secure strings for the payment gateway.
		 * @param string   $id              The gateway ID.
```

includes/abstracts/abstract.llms.payment.gateway.php
```diff
@@ -937,6 +959,49 @@ protected function get_option_prefix() {
 	 */
 	public function process_refund( $transaction, $amount = 0, $note = '' ) {}
 
+	/**
+	 * Retrieves a list of "secure" strings which should be anonymized if they're found within debug logs.
+	 *
+	 * This method will load the value any gateway option with a `secure_option` declaration. Additional
```
pondermatic:
```suggestion
	 * This method will load the values of any gateway options with a `secure_option` declaration. Additional
```

includes/abstracts/llms.abstract.user.data.php
```diff
@@ -42,18 +42,22 @@ abstract class LLMS_Abstract_User_Data {
...
-	 * @return   void
-	 * @since    2.2.3
-	 * @version  3.9.0
+	 * @since 2.2.3
+	 * @since 3.9.0 Unknown.
+	 * @since [version]
+	 *
+	 * @param int|null|WP_User|LLMS_Abstract_User_Data $user     A `WP_User` ID, instance of a `WP_User`, or instance of any class extending this class.
+	 * @param boolean                                  $autoload If `true` and `$user` input is empty, the user will be loaded from `get_current_user_id()`.
+	 *                                                           If `$user` is not empty then this parameter has no impact.
```
pondermatic, line 55:
```suggestion
	 * @param int|null|WP_User|LLMS_Abstract_User_Data $user     A `WP_User` ID, instance of a `WP_User`,
	 *                                                           or instance of any class extending this class.
	 * @param boolean                                  $autoload If `true` and `$user` input is empty,
	 *                                                           the user will be loaded from `get_current_user_id()`.
	 *                                                           If `$user` is not empty then this parameter has no impact.
```

includes/abstracts/llms.abstract.user.data.php
```diff
@@ -42,18 +42,22 @@ abstract class LLMS_Abstract_User_Data {
...
+	 * @return void
 	 */
-	public function __construct( $user = null ) {
+	public function __construct( $user = null, $autoload = true ) {
 
-		$user = $this->get_user_id( $user );
+		$user = ( $user || ( ! $user && $autoload ) ) ? $this->get_user_id( $user ) : false;
```
pondermatic: The `! user` condition is always 'true' because '$user' is already 'false' at this point.
```suggestion
		$user = ( $user || $autoload ) ? $this->get_user_id( $user ) : false;
```

includes/controllers/class.llms.controller.orders.php
```diff
@@ -927,6 +720,57 @@ private function validate_selected_gateway( $gateway_id, $plan ) {
...
+	 * @since 3.34.4 Added filter `llms_order_can_be_confirmed`.
+	 * @since 3.34.5 Fixed logic error in `llms_order_can_be_confirmed` conditional.
+	 * @since 3.35.0 Return early if nonce doesn't pass verification and sanitize `$_POST` data.
+	 * @since 5.9.0 Stop using deprecated `FILTER_SANITIZE_STRING`.
+	 * @deprecated [version] Deprecated in favor of {@see LLMS_Controller_Checkout::confirm_pending_order}.
+	 *
+	 * @return void
+	 */
+	public function confirm_pending_order() {
+		_deprecated_function( __METHOD__, '[verison]', 'LLMS_Controller_Checkout::confirm_pending_order' );
```
pondermatic:
```suggestion
		_deprecated_function( __METHOD__, '[version]', 'LLMS_Controller_Checkout::confirm_pending_order' );
```

includes/controllers/class.llms.controller.orders.php
```diff
@@ -927,6 +720,57 @@ private function validate_selected_gateway( $gateway_id, $plan ) {
...
+	 * @since 3.35.0 Sanitize `$_POST` data.
+	 * @since 5.0.0 Build customer data using LLMS_Forms fields information.
+	 * @since 5.0.1 Delegate sanitization of user information fields of the `$_POST` to LLMS_Form_Handler::submit().
+	 * @since 5.9.0 Stop using deprecated `FILTER_SANITIZE_STRING`.
+	 * @deprecated [version] Deprecated in favor of {@see LLMS_Controller_Checkout::create_pending_order}.
+	 *
+	 * @return void
+	 */
+	public function create_pending_order() {
+		_deprecated_function( __METHOD__, '[verison]', 'LLMS_Controller_Checkout::create_pending_order' );
```
pondermatic:
```suggestion
		_deprecated_function( __METHOD__, '[version]', 'LLMS_Controller_Checkout::create_pending_order' );
```

includes/forms/class-llms-form-handler.php
```diff
@@ -304,11 +304,15 @@ protected function prepare_data_for_insert( $posted_data, $fields, $action ) {
 	 *
 	 * @since 5.0.0
 	 * @since 5.1.0 Remove invisible fields from when loading the checkout form.
+	 * @since [version] Allow submission validation only (without actually submitting the fields) using the
+	 *               `validate_only` flag in the `$args` array.
```
pondermatic:
```suggestion
	 * @since [version] Allow submission validation only (without actually submitting the fields) using the
	 *              `validate_only` flag in the `$args` array.
```

includes/forms/class-llms-form-handler.php
```diff
@@ -304,11 +304,15 @@ protected function prepare_data_for_insert( $posted_data, $fields, $action ) {
...
 	 *
 	 * @param array  $posted_data User-submitted form data.
 	 * @param string $location    Form location ID.
 	 * @param array  $args        Additional arguments passed to the short-circuit filter.
-	 * @return int|WP_Error WP_User ID on success, error object on failure.
+	 * @return integer|boolean|WP_Error On success returns the `WP_User` ID.
+	 *                                  If the `validate_only` argument is passed returns `true` on succes.
```
pondermatic:
```suggestion
	 *                                  If the `validate_only` argument is passed returns `true` on success.
```

includes/functions/llms.functions.order.php
```diff
@@ -164,6 +164,46 @@ function llms_get_possible_order_statuses( $order ) {
...
+ *
+ * Used during AJAX checkout order creation when users are not created until the gateway confirms success.
+ *
+ * Ensures that only a single pending order for a given plan and email address will exist at any given time.
+ *
+ * @since [version]
+ *
+ * @param string $email   An email address.
+ * @param int    $plan_id Access plan WP_Post ID.
+ * @return null|int Returns the post id if found, otherwise returns `null`.
```
pondermatic, line 218:
```suggestion
 * @return null|int Returns the post ID if found, otherwise returns `null`.
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
+<?php
+/**
+ * LLMS_Order_Generator class file.
```
pondermatic, line 3:
```suggestion
 * LLMS_Order_Generator class file
```
https://github.com/gocodebox/lifterlms/blob/trunk/docs/documentation-standards.md#file-headers-1
https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#6-file-headers

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+
+	/**
+	 * Error code: issue encountered during order post creation.
+	 *
+	 * @var string
+	 */
+	const E_CREATE_ORDER = 'llms-order-gen-create-order';
+
+	/**
+	 * Error code: payment gateway id not submitted.
```
pondermatic, line 42:
```suggestion
	 * Error code: payment gateway ID not submitted.
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+ * @since [version]
+ */
+class LLMS_Order_Generator {
+
+	/**
+	 * Error code: invalid coupon code submitted.
+	 *
+	 * @var string
+	 */
+	const E_COUPON_INVALID = 'llms-order-gen-coupon-invalid';
```
pondermatic, line 25: Yay constants!

Class constants can (and should IMHO) use [visibility modifiers](https://www.php.net/manual/en/language.oop5.constants.php#example-226).

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+	 * @since [version]
+	 *
+	 * @param array $data {
+	 *     An associative array of input data used to generate the order, usually from $_POST.
+	 *
+	 *     @type integer $llms_plan_id         An LLMS_Access_Plan ID.
+	 *     @type string  $llms_agree_to_terms  A yes/no value determining whether or not the user has agreed to the site's terms.
+	 *     @type string  $llms_payment_gateway The ID of the payment gateway used to process the order.
+	 *     @type string  $llms_coupon_code     Optional. The coupon code string being used.
+	 *     @type string  $llms_order_key       Optional. An `LLMS_Order` key used to modify an existing pending order rather than creating a new one.
```
pondermatic, line 180:
```suggestion
	 *     @type string  $llms_order_key       Optional. An `LLMS_Order` key used to modify an existing pending order
	 *                                         rather than creating a new one.
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+
+	}
+
+	/**
+	 * Performs all required data validations necessary to create the order.
+	 *
+	 * @since [version]
+	 *
+	 * @param boolean $validate_order Whether or not order data should be validated. This is `true` when running `confirm()` and `false` otherwise.
+	 * @return boolean|WP_Errors Returns `true` if all validations pass or an error object.
```
pondermatic:
```suggestion
	 * @return boolean|WP_Error Returns `true` if all validations pass or an error object.
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+	 * @return WP_Error|array Returns an array of data from the payment gateway's `confirm_pending_order()` method on success.
+	 */
+	public function confirm() {
+
+		$validate = $this->validate( true );
+		if ( is_wp_error( $validate ) ) {
+			return $validate;
+		}
+
+		$gateway_confirm = $this->gateway->confirm_pending_order( $this->order );
```
pondermatic, line 203: The abstract `LLMS_Payment_Gateway::confirm_pending_order()` signature says this method returns `void`. `LLMS_Payment_Gateway_PayPal::confirm_pending_order()` in the `dev` and `dev-200` branches returns `void`. `LLMS_Payment_Gateway_Stripe::confirm_pending_order()` in the `dev` branch returns `void`.
The new tests return `WP_Error|array|void`.
This line expects `WP_Error|array`.

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+	 *
+	 * @since [version]
+	 *
+	 * @return WP_Error|LLMS_Order
+	 */
+	protected function create() {
+
+		$order = new LLMS_Order( $this->get_order_id() );
+
+		// If there's no id we can't proceed, return an error.
```
pondermatic, line 237:
```suggestion
		// If there's no ID we can't proceed. Return an error.
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+		if ( ! $order->get( 'id' ) ) {
+			return $this->error(
+				self::E_CREATE_ORDER,
+				__( 'There was an error creating your order, please try again.', 'lifterlms' )
```
pondermatic, line 241:
```suggestion
				__( 'There was an error creating your order. Please try again.', 'lifterlms' )
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+
+		$user = get_current_user_id() ?
+			llms_update_user( $this->data, 'checkout', $args ) :
+			llms_register_user( $this->data, 'checkout', true, $args );
+
+		if ( ! is_wp_error( $user ) ) {
+			$this->student = llms_get_student( $user );
+		}
+
+		return $user;
```
pondermatic:
```suggestion
		$user_id = get_current_user_id() ?
			llms_update_user( $this->data, 'checkout', $args ) :
			llms_register_user( $this->data, 'checkout', true, $args );

		if ( ! is_wp_error( $user_id ) ) {
			$this->student = llms_get_student( $user_id );
		}

		return $user_id;
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+		}
+
+		return new WP_Error( $code, $message, array_merge( $data, $extra_data ) );
+
+	}
+
+	/**
+	 * Attempts to locate a user ID.
+	 *
+	 * Uses the logged in users information and falls back to a lookup by email address if available.
```
pondermatic:
```suggestion
	 * Uses the logged in user's information and falls back to a lookup by email address if available.
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+	 * @since [version]
+	 *
+	 * @return LLMS_Payment_Gateway|null
+	 */
+	public function get_gateway() {
+		return $this->gateway;
+	}
+
+	/**
+	 * Retrieves the order id to use for the order.
```
pondermatic, line 377:
```suggestion
	 * Retrieves the order ID to use for the order.
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+			$order_id = $user_id ? llms_locate_order_for_user_and_plan( $user_id, $plan_id ) : $order_id;
+
+		}
+
+		// Lookup by email address.
+		if ( ! $order_id && $email ) {
+			$order_id = llms_locate_order_for_email_and_plan( $email, $plan_id );
+		}
+
+		return $order_id ? $order_id : 'new';
```
pondermatic, line 411:
```suggestion
		return $order_id ?: 'new';
```
[Ternary Operator](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.ternary)

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+		return $data;
+
+	}
+
+	/**
+	 * Performs all required data validations necessary to create the order.
+	 *
+	 * @since [version]
+	 *
+	 * @param boolean $validate_order Whether or not order data should be validated. This is `true` when running `confirm()` and `false` otherwise.
```
pondermatic, line 496:
```suggestion
	 * @param boolean $validate_order Whether or not order data should be validated.
	 *                                This is `true` when running `confirm()` and `false` otherwise.
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+	 * @return boolean|WP_Errors Returns `true` if all validations pass or an error object.
+	 */
+	protected function validate( $validate_order = false ) {
+
+		/**
+		 * Allows 3rd party validation prior to generation of an order.
+		 *
+		 * This validation hooks runs prior to all default validation.
```
pondermatic:
```suggestion
		 * This validation hook runs prior to all default validation.
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+		 *
+		 * @since [version]
+		 *
+		 * @param null|WP_Error $validation_errors Halts checkout and returns the supplied error.
```
pondermatic:
```suggestion
		 * @param null|WP_Error $validation_error Halts checkout and returns the supplied error.
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+		}
+
+		/**
+		 * Allows 3rd party prior to generation of an order.
+		 *
+		 * This validation hooks runs after to all default validation passes.
+		 *
+		 * @since [version]
+		 *
+		 * @param boolean|WP_Error $validation_errors Halts checkout and returns the supplied error.
```
pondermatic:
```suggestion
		 * Allows 3rd party validation prior to generation of an order.
		 *
		 * This validation hook runs after all default validation.
		 *
		 * @since [version]
		 *
		 * @param boolean|WP_Error $validation_error Halts checkout and returns the supplied error.
```

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+					sprintf(
+						// Translators: %s = The user-submitted coupon code.
+						__( 'Coupon code "%s" not found.', 'lifterlms' ),
+						$code
+					)
+				);
+			}
+
+			// Validate the coupon for the current plan.
+			$coupon = llms_get_post( $coupon_id );
```
pondermatic, line 575: Might want to do a sanity check here in case `llms_get_post()` returns `null` or `false`.

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+	 *
+	 * @since [version]
+	 *
+	 * @return boolean|WP_Error Returns `true` on success or an error object.
+	 */
+	protected function validate_gateway() {
+
+		$coupon_id = $this->coupon ? $this->coupon->get( 'id' ) : null;
+
+		/**
```
pondermatic, line 599:
```suggestion
		/*
```
It's a [multi-line comment](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#5-2-multi-line-comments), not a DocBlock.

includes/class-llms-order-generator.php
```diff
@@ -0,0 +1,733 @@
...
+		$order_id = $this->get_order_id();
+
+		if ( 'new' === $order_id || 'llms_order' !== get_post_type( $order_id ) ) {
+			return $this->error(
+				self::E_ORDER_NOT_FOUND,
+				__( 'Could not locate an order to confirm.', 'lifterlms' )
+			);
+		}
+
+		$order = llms_get_post( $order_id );
```
pondermatic, line 639: Might want to do a sanity check in case `llms_get_post()` returns `null` or `false`.

pondermatic (changes requested):
I still have 2 more classes and all of the unit tests to review, but need to stop for tonight.

includes/class-llms-order-generator.php
thomasplevy: Yes.

Moving forward gateways will choose how to implement `confirm_pending_order()`.

A gateway that implements AJAX functionality will return `WP_Error|array` while the "legacy" gateways (PayPal, Stripe, Auth.net, any other custom gateways) would never call this method and therefore returning `void()` would work just fine.

The only gateway that will utilize this will be the new PayPal gateway (in development): https://github.com/gocodebox/lifterlms-gateway-paypal/blob/34ff6938cea0de61e41f598777a32a1502e75167/includes/class-llms-payment-gateway-paypal-cp.php#L125-L137

includes/class-llms-order-generator.php
thomasplevy: I'm sure we've talked about this already but if we haven't we're *not* using short ternary because WordPress doesn't: https://make.wordpress.org/core/2019/07/12/php-coding-standards-changes/

includes/class-llms-order-generator.php
thomasplevy: yea but `llms_find_coupon()` above already validates that we have a coupon ID. I can't conceive of any scenario where the function is going to return a non-coupon post ID.

includes/class-llms-order-generator.php
thomasplevy: `get_order_id()` runs that sanity check for us

includes/models/model.llms.order.php
thomasplevy: good call: https://github.com/gocodebox/lifterlms/pull/2102/commits/7d4288b4cc6ecf940c3a75e2184ad074d2857edc

includes/class-llms-order-generator.php
pondermatic: An unlikely scenario is that the coupon post is deleted by another HTTP request in between this method calling `llms_find_coupon()` and `llms_get_post()`.

includes/class-llms-order-generator.php
pondermatic: The return type of the abstract `LLMS_Payment_Gateway::confirm_pending_order()` should be changed from `void` to `mixed`.

includes/class-llms-order-generator.php
pondermatic: Sorry. I forgot.
https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#ternary-operator

`??` null coalescing operator - good
`?:` ternary operator - bad

includes/controllers/class-llms-controller-checkout.php
```diff
@@ -0,0 +1,404 @@
...
+	 * @since [version] Relocated from `LLMS_Controller_Orders`.
+	 *
+	 * @return void
+	 */
+	public function confirm_pending_order() {
+
+		// Verify form submission.
+		$verify = $this->verify_request( '_wpnonce', self::ACTION_CONFIRM_PENDING_ORDER );
+		if ( ! $verify ) {
+			return $verify;
```
pondermatic: In addition to `void`, this method can return `null` or `false`.

includes/controllers/class-llms-controller-checkout.php
```diff
@@ -0,0 +1,404 @@
...
+
+		// Ensure we have an order key we can locate the order with.
+		$key = llms_filter_input_sanitize_string( INPUT_POST, 'llms_order_key' );
+		if ( ! $key ) {
+			return llms_add_notice( __( 'Could not locate an order to confirm.', 'lifterlms' ), 'error' );
+		}
+
+		// Lookup the order & return error if not found.
+		$order = llms_get_order_by_key( $key );
+		if ( ! $order || ! $order instanceof LLMS_Order ) {
```
pondermatic, line 111: The `! order` condition is unnecessary because it is checked by '! $order instanceof LLMS_Order' .

includes/controllers/class-llms-controller-checkout.php
```diff
@@ -0,0 +1,404 @@
...
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function confirm_pending_order_ajax() {
+
+		$verify = $this->verify_request( self::AJAX_QS_VAR, self::ACTION_CONFIRM_PENDING_ORDER );
+		if ( ! $verify ) {
+			return $verify;
```
pondermatic, line 147: In addition to `void`, this method can return `null` or `false`.

includes/controllers/class-llms-controller-checkout.php
```diff
@@ -0,0 +1,404 @@
...
+		$setup = llms_setup_pending_order( $this->extract_setup_data( $_POST ) );  // phpcs:ignore WordPress.Security.NonceVerification.Missing -- Nonce verified via `verify_request()`.
+		if ( is_wp_error( $setup ) ) {
+			llms_add_notice( $setup->get_error_message(), 'error' );
+			return $setup;
+		}
+
+		/**
+		 * Allow gateways, extensions, etc to do their own validation.
+		 *
+		 * After all standard validations are successfully.
```
pondermatic, line 217:
```suggestion
		 * Allow gateways, extensions, etc. to do their own validation after all standard validations are successfully completed.
```

includes/controllers/class-llms-controller-checkout.php
```diff
@@ -0,0 +1,404 @@
...
+	 * of the method is halted early. Gateways should do the same if they encounter errors during processing.
+	 *
+	 * Upon success the gateway should redirect the user to the relevant next step. For multi-step checkout that
+	 * requires payment confirmation, the user should be redirected to the order confirmation page, for one-step
+	 * gateways assuming the order is moved to active or completed status and enrollment takes place, the user
+	 * should be redirected to the relevant course or membership URL.
+	 *
+	 * @since [version] Moved from `LLMS_Controller_Orders.
+	 *
+	 * @return null|boolean|WP_Error
```
pondermatic: This method also returns `void`. Perhaps the places it returns `void` should be changed to return `bool`.

includes/controllers/class-llms-controller-checkout.php
```diff
@@ -0,0 +1,404 @@
...
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function create_pending_order_ajax() {
+
+		$verify = $this->verify_request( self::AJAX_QS_VAR, self::ACTION_CREATE_PENDING_ORDER );
+		if ( ! $verify ) {
+			return $verify;
```
pondermatic, line 277: In addition to `void`, this method can return `null` or `false`.

includes/controllers/class-llms-controller-checkout.php
```diff
@@ -0,0 +1,404 @@
...
+
+		// Generate the order.
+		$generator = new LLMS_Order_Generator( $_POST ); // phpcs:ignore WordPress.Security.NonceVerification.Missing -- Nonce verified via `verify_request()`.
+		$order     = $generator->generate( $generator::UA_VALIDATE );
+		if ( is_wp_error( $order ) ) {
+			$this->send_json( $order );
+		}
+
+		// Pending order creation success, pass it over to the gateway.
+		$handle = $generator->get_gateway()->handle_pending_order(
```
pondermatic, line 290: The abstract `LLMS_Payment_Gateway::handle_pending_order()` return type should be changed from `void` to `mixed` or `void|array|WP_Error`.

includes/controllers/class-llms-controller-checkout.php
```diff
@@ -0,0 +1,404 @@
...
+
+		$user_data = array();
+		$plan      = $plan_id ? llms_get_post( $plan_id ) : false;
+
+		$uid = get_current_user_id();
+		if ( $uid ) {
+			$user_data['user_id'] = $uid;
+		}
+
+		foreach ( LLMS_Forms::instance()->get_form_fields( 'checkout', compact( 'plan' ) ) as $field ) {
```
pondermatic, line 348: Difficult to debug. Please consider:
```suggestion
		$fields = LLMS_Forms::instance()->get_form_fields( 'checkout', compact( 'plan' ) );
		foreach ( $fields as $field ) {
```

includes/models/model.llms.transaction.php
```diff
@@ -69,30 +69,97 @@ class LLMS_Transaction extends LLMS_Post_Model {
...
+	 * to the gateway for processing (via the gateway's API). For custom methods processing is handled
+	 * via the `llms_{$method}_refund_id` filter.
+	 *
+	 * @since [version]
+	 *
+	 * @param string $method Refund processing method ID.
+	 * @param float  $amount The amount to refund.
+	 * @param string $note   Refund notes.
+	 * @return string|boolean|WP_Error Returns the generated refund ID string or an error object. If a falsy value
+	 *                                 is returned the refund processing with fail with a generic error message.
```
pondermatic: I'm not sure what the 2nd sentence means. Should it be?
```suggestion
	 * @return string|boolean|WP_Error Returns the generated refund ID string or an error object. If a falsy value
	 *                                 is returned the refund processing will fail with a generic error message.
```

includes/models/model.llms.transaction.php
```diff
@@ -156,157 +223,338 @@ public function get_net_amount( $price_args = array(), $format = 'html' ) {
...
-	 *
-	 * @param    float  $amount   amount to refund
-	 * @param    string $note     optional note to record in the gateway (if possible) and as an order note
-	 * @param    string $method   method used to refund, either "manual" (available for all transactions) or "gateway" (where supported)
-	 * @return   string|WP_Error      a refund ID on success or a WP_Error object
-	 * @since    3.0.0
-	 * @version  3.0.0
+	 * Retrieves the title of the refund method using during refund processing.
+	 *
+	 * This method is record the method used to process a refund in the refund order note.
```
pondermatic:
```suggestion
	 * This method records the method used to process a refund in the refund order note.
```

includes/models/model.llms.transaction.php
```diff
@@ -156,157 +223,338 @@ public function get_net_amount( $price_args = array(), $format = 'html' ) {
...
+				 * Filters the refund method title for custom refund methods.
+				 *
+				 * The dynamic portion of this hook, `{$method}`, represents the ID of the custom refund method.
+				 *
+				 * @since 3.0.0
+				 * @deprecated [version] Replaced with `llms_{$method}_refund_title`.
+				 *
+				 * @param string $method The method ID.
+				 */
+				$method_title = apply_filters_deprecated( "llms_{$method}_title", array( $method ), '[version]', "llms_{$method}_refund_title" );
```
pondermatic, line 295: If this deprecated hook is actually used, the return value is never used.
Perhaps add:
```php
				if ( $method_title !== $method ) {
					return $method_title;
				}
```
or
```php
				if ( $method_title !== $method ) {
					$method = $method_title;
				}
```

tests/phpunit/unit-tests/forms/class-llms-test-form-handler.php
```diff
@@ -293,6 +292,42 @@ public function test_submit_registration_voucher_err_max() {
...
+	public function test_submit_validate_only_error() {
+
+		$ret = $this->handler->submit( array(), 'checkout', array( 'validate_only' => true ) );
+
+		$this->assertIsWPError( $ret );
+		$this->assertWPErrorCodeEquals( 'llms-form-missing-required', $ret );
+
+	}
+
+
```
pondermatic:
```suggestion
```

tests/phpunit/unit-tests/functions/class-llms-test-functions-person.php
```diff
@@ -374,4 +374,49 @@ public function test_llms_set_user_login_time() {
...
+
+		$data = $this->get_mock_user_data_array();
+		$user = $this->factory->user->create_and_get( array( 
+			'first_name' => 'Kyle',
+			'user_email' => $data['email_address'],
+		) );
+
+		wp_set_current_user( $user->ID );
+
+		$args['first_name'] = 'NotKyle';
```
pondermatic: Should this be?
```suggestion
		$data['first_name'] = 'NotKyle';
```

tests/phpunit/unit-tests/models/class-llms-test-model-llms-order.php
```diff
@@ -1034,6 +1065,135 @@ public function test_has_trial() {
...
+
+		$this->assertEquals( 'email@email.tld', $order->get( 'billing_email' ) );
+		$this->assertEquals( 0, $order->get( 'user_id' ) );
+
+		remove_action( 'lifterlms_new_pending_order', $handler, 10 );
+
+	}
+
+
+
```
pondermatic:
```suggestion
```

tests/phpunit/unit-tests/controllers/class-llms-test-controller-checkout.php
```diff
@@ -0,0 +1,636 @@
...
+ *
+ * @group orders
+ * @group controllers
+ * @group checkout
+ * @group controller_checkout
+ *
+ * @since [version]
+ */
+class LLMS_Test_Controller_Checkout extends LLMS_UnitTestCase {
+
```
pondermatic, line 15:
```suggestion

	/**
	 * The System Under Test (SUT).
	 *
	 * @since [version]
	 *
	 * @var LLMS_Controller_Checkout
	 */
	private $main;

```

tests/phpunit/unit-tests/functions/class-llms-test-functions-person.php
```diff
@@ -374,4 +374,49 @@ public function test_llms_set_user_login_time() {
...
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function test_llms_validate_user_success() {
+
+		LLMS_Forms::instance()->install( true );
+
+		$data = $this->get_mock_user_data_array();
+		$user = $this->factory->user->create_and_get( array(
```
pondermatic:
```suggestion
		$user = $this->factory->user->create_and_get( array(
```

tests/phpunit/unit-tests/controllers/class-llms-test-controller-checkout.php
```diff
@@ -0,0 +1,636 @@
...
+		foreach ( $actions as $hook => $priority ) {
+			remove_action( 'init', array( $this->main, $hook ), $priority );
+		}
+
+		LLMS_Unit_Test_Util::call_method( $this->main, '__construct' );
+
+		foreach ( $actions as $hook => $priority ) {
+			$this->assertEquals( $priority, has_action( 'init', array( $this->main, $hook ) ) );
+		}
```
pondermatic, line 49: There are extra spaces (and tabs) at the end of lines 47, 90, 284, 522, 550, 568, 598, 616 and 617.

tests/phpunit/unit-tests/controllers/class-llms-test-controller-checkout.php
```diff
@@ -0,0 +1,636 @@
...
+		);
+
+		$this->mockPostRequest( $post_data );
+
+		// Assert some things using an action hook from the manual payment gateway.
+		$handler = function( $order ) use ( $plan, $post_data ) {
+
+			$this->assertEquals( $plan->get( 'id' ), $order->get( 'plan_id' ) );
+
+			$student = llms_get_student( $order->get( 'user_id' ) );
```
pondermatic, line 430: `$student` isn't used.

tests/phpunit/unit-tests/class-llms-test-order-generator.php
```diff
@@ -0,0 +1,967 @@
...
+	 *
+	 * @return void
+	 */
+	public function test_create_error() {
+
+		// Forces an error to be encountered during order creation.
+		add_filter( 'wp_insert_post_empty_content', '__return_true' );
+
+		$gen = new LLMS_Order_Generator( array() );
+		$res = LLMS_Unit_Test_Util::call_method( $gen, 'create' );
```
pondermatic, line 128: `LLMS_Unit_Test_Util::call_method()` can throw an exception, so methods that use it should have a `@throws ReflectionException` tag in the DocBlock, unless they catch it.

tests/phpunit/unit-tests/class-llms-test-order-generator.php
```diff
@@ -0,0 +1,967 @@
...
+		$this->assertTrue( is_numeric( $res ) );
+		$student = llms_get_student( $res );
+		$this->assertEquals( $data['email_address'], $student->get( 'user_email' ) );
+		$this->assertEquals( $data['first_name'], $student->get( 'first_name' ) );
+
+		// Update the user.
+		wp_set_current_user( $res );
+		$data['first_name'] = 'Albert';
+		$gen  = new LLMS_Order_Generator( $data );
+		$res  = LLMS_Unit_Test_Util::call_method( $gen, 'commit_user' );
```
pondermatic:
```suggestion
		$gen = new LLMS_Order_Generator( $data );
		$res = LLMS_Unit_Test_Util::call_method( $gen, 'commit_user' );
```

tests/phpunit/unit-tests/class-llms-test-order-generator.php
```diff
@@ -0,0 +1,967 @@
...
+		$this->assertWPErrorMessageEquals( 'Mock Message', $res );
+		$this->assertWPErrorDataEquals(
+			array(
+				'coupon'      => 'coupon_value',
+				'data'        => $data,
+				'gateway'     => 'gateway_value',
+				'plan'        => 'plan_value',
+				'student'     => 'student_value',
+				'extra'       => 1,
+				'order'       => 'order_value',
```
pondermatic:
```suggestion
				'coupon'  => 'coupon_value',
				'data'    => $data,
				'gateway' => 'gateway_value',
				'plan'    => 'plan_value',
				'student' => 'student_value',
				'extra'   => 1,
				'order'   => 'order_value',
```

pondermatic (changes requested):
This should really increase our testability score!

includes/models/model.llms.transaction.php
thomasplevy: yes good catch: https://github.com/gocodebox/lifterlms/pull/2102/commits/8efc7d8c3eb94e259df22a8621a64870d8eedb50

includes/abstracts/llms.abstract.user.data.php
thomasplevy: yes good call: https://github.com/gocodebox/lifterlms/pull/2102/commits/4dc3c8be431af10a47e03ca54f754229007fa92d

thomasplevy:
@gocodebox/engineering I'd like to get some testing against this (both with existing gateways and the new PayPal gateway).

The new PayPal isn't entirely finished but checkout for both one-time and recurring payments is functioning as expected. See the (misnamed) dev-200 branch of the PP repo. PayPal sandbox credentials are in 1password.

Let me know if you have some feedback after doing real testing. Any questions or need help setting up environments let me know.

pondermatic:
> Any questions or need help setting up environments let me know.

I've tried both the "facilitator" and "seller" sandbox accounts (and even my own merchant sandbox account) and I always get "Unable to retrieve a signup link." After the LifterLMS "PayPal Payment Gateway Settings" page reloads.

thomasplevy:
You have to connect to the paypal sandbox, so enable test mode (the checkbox) and then try

pondermatic:
Yes, I was enabling the sandbox checkbox. I eventually got it to work by manually entering the credentials.

pondermatic:
After working through the settings issues, I was able to successfully test buying a course with 1 payment for lifetime access. I did not notice any errors in the PayPal log and there were zero errors in the PHP log.

Sadly I can not test any more until I get back next week. I was supposed to be packing an hour ago.

eri-trabiccolo (approved):
I've tested this with the PayPal 3.0 (and Stripe, ensuring it still works fine) and it looks pretty solid to me.

thomasplevy:
@gocodebox/engineering consensus:

Should we merge this into dev and ship with the next release or should we release a beta version of the core with this and hold on public release for a bit longer?

eri-trabiccolo:
While I think that this PR is good enough, since our main priority is releasing PayPal add-on 3.0 (and basically the code here at the moment is leveraged by pp 3.0 only), I think this should not be merged into dev, but should be included in a beta release (along with pp 3.0 beta). This is a conservative choice. My reasoning is that if we missed any issues in this PR, by releasing it publicly, if any issue pops out we're forced to change our focus on that and fix it in priority.
The downside of this choice it's, of course, that it'll be tested by a minor number of customers (the ones who test the demo).

includes/abstracts/llms.abstract.user.data.php
```diff
@@ -48,7 +48,7 @@ abstract class LLMS_Abstract_User_Data {
 	 *
 	 * @since 2.2.3
 	 * @since 3.9.0 Unknown.
-	 * @since [version]
+	 * @since [version] Added `$autoload` parameter.
```
pondermatic, line 51: The file-level `@version` tag needs to be updated as well.

includes/admin/views/metaboxes/view-order-details.php
```diff
@@ -8,7 +8,7 @@
  * @since 5.4.0 Inform about deleted products.
  * @since 6.1.0 Add validation to the remaining payments input.
  *              Allow the number of remaining payments to be `0` for already-completed payment plans.
- * @version 5.4.0
+ * @version [version]
```
pondermatic, line 11: Should there also be a new `@since` changelog?

includes/admin/views/metaboxes/view-order-submit.php
```diff
@@ -4,16 +4,17 @@
  *
  * @since 3.19.0
  * @since 5.4.0 The order status dropdown is now limited to a subset of possible status.
- * @version 5.4.0
+ * @version [version]
```
pondermatic, line 7: Add a `@since` changelog?

includes/controllers/class-llms-controller-checkout.php
```diff
@@ -166,7 +168,9 @@ public function confirm_pending_order_ajax() {
 	 *
 	 * @since [version] Moved from `LLMS_Controller_Orders.
 	 *
-	 * @return null|boolean|WP_Error
+	 * @return null|boolean|void Returns `null` when the form isn't submitted or there's a nonce verification issue.
+	 *                           Returns `false` when the the request is missing the action parameter or the action doesn't match
+	 *                           the expected action. Otherwise there is no/void return.
```
pondermatic, line 181: This method can also return WP_Error when setting up the order.

tests/phpunit/unit-tests/admin/post-types/meta-boxes/class-llms-test-meta-box-order-details.php
```diff
@@ -227,4 +269,140 @@ public function test_save_remaining_payments_success() {
...
 	}
 
+	/**
+	 * Test save_remaining_payments() when changes are made but the order doesn't support recurring payment modifications.
+	 *
+	 * @since [version]
+	 *
+	 * @return void
+	 */
+	public function test_save_remaining_payments_unsucces_when_order_does_not_support_recurring_payment_modifications() {
```
pondermatic:
```suggestion
	public function test_save_remaining_payments_unsuccess_when_order_does_not_support_recurring_payment_modifications() {
```
This method name reminds me of the South Park movie title "Bigger, Longer & Uncut".

pondermatic (changes requested):

includes/admin/views/metaboxes/view-order-details.php
thomasplevy: @pondermatic can you open a new PR to address the various documentation issues you've raised on your latest review. Thanks!

PR #2113, opened 2022-04-13
Initialize shortcodes early

### LINKED ISSUE #2070: LifterLMS Shortcodes from Dropdown don't render in emails
### Reproduction Steps

+ HS-189435
+ Create a custom email
+ from the mergecode menu copy a shortcode like [llms-user user-email] into the email
+ trigger the email to be sent

### Expected Behavior

+ the shortcode either shouldn't be in the merge code list or should render in the email

### Actual Behavior

+ the shortcode is an option in the merge codes and doesn't render in the email

### Error Messages / Logs

```

```

<details>

```
Wordpress

Home Url: https://natalie-test-site.myliftersite.com
Site Url: https://natalie-test-site.myliftersite.com
Login Url: https://natalie-test-site.myliftersite.com/wp-login.php
Version: 5.9.2
Debug Mode: No
Debug Log: No
Debug Display: Yes
Locale: en_US
Multisite: No
Page For Posts: Not Set
Page On Front: Homepage (#1398) [https://natalie-test-site.myliftersite.com/]
Show On Front: page
Wp Cron: Yes


Settings

Version: 6.0.0
Db Version: 6.0.0
Course Catalog: Course Catalog (#35) [https://natalie-test-site.myliftersite.com/courses/]
Membership Catalog: Membership Catalog (#36) [https://natalie-test-site.myliftersite.com/memberships/]
Student Dashboard: Dashboard (#38) [https://natalie-test-site.myliftersite.com/dashboard/]
Checkout Page: Checkout (#715) [https://natalie-test-site.myliftersite.com/checkout/]
Course Catalog Per Page: 9
Course Catalog Sorting: menu_order,ASC
Membership Catalog Per Page: 9
Membership Catalog Sorting: menu_order
Site Membership: Not Set
Courses Endpoint: my-courses
Edit Endpoint: edit-account
Lost Password Endpoint: lost-password
Vouchers Endpoint: redeem-voucher
Autogenerate Username: no
Password Strength Meter: no
Minimum Password Strength: 
Terms Required: yes
Terms Page: Activate (#443) [https://natalie-test-site.myliftersite.com/activate/]
Checkout Names: 
Checkout Address: 
Checkout Phone: 
Checkout Email Confirmation: no
Open Registration: yes
Registration Names: 
Registration Address: 
Registration Phone: 
Registration Voucher: 
Registration Email Confirmation: no
Account Names: 
Account Address: 
Account Phone: 
Account Email Confirmation: no
Confirmation Endpoint: confirm-payment
Force Ssl Checkout: no
Country: US
Currency: USD
Currency Position: left
Thousand Separator: ,
Decimal Separator: .
Decimals: 2
Trim Zero Decimals: no
Recurring Payments: yes
Email From Address: team+sandbox@lifterlms.com
Email From Name: A LifterLMS Sandbox
Email Footer Text: 
Email Header Image: 
Cert Bg Width: 1123
Cert Bg Height: 794
Cert Legacy Compat: no


Constants

LLMS_REMOVE_ALL_DATA: undefined
LLMS_REST_DISABLE: undefined
LLMS_SITE_FEATURE_RECURRING_PAYMENTS: undefined
LLMS_SITE_IS_CLONE: undefined


Gateways

Manual: Enabled
Manual Logging: no
Manual Order: 1


Server

Mysql Version: 5.7.37
Php Curl: Yes
Php Default Timezone: UTC
Php Fsockopen: Yes
Php Max Input Vars: 5000
Php Max Upload Size: 512 MB
Php Memory Limit: 256M
Php Post Max Size: 1024M
Php Soap: Yes
Php Suhosin: No
Php Time Limt: 30
Php Version: 8.0.16
Software: Apache/2.4.52 (Unix) OpenSSL/1.1.1
Wp Memory Limit: 256M


Browser

HTTP USER AGENT: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/99.0.4844.74 Safari/537.36


Theme

Name: Astra
Version: 3.7.9
Themeuri: https://wpastra.com/
Authoruri: https://wpastra.com/about/?utm_source=theme_preview&utm_medium=author_link&utm_campaign=astra_theme
Template: 
Child Theme: No
Llms Support: Yes


Plugins

LifterLMS: 6.0.0
LifterLMS WooCommerce: 2.2.2
Loco Translate: 2.6.1
User Switching: 1.5.8
WooCommerce: 6.3.1
WooCommerce Subscriptions: 4.0.2
WP Mail Logging: 1.10.4
WP Rollback: 1.7.1


Integrations

BbPress: No
BuddyPress: No
WooCommerce: Yes


Template Overrides




```

</details>

This issue has be recreated:
+ [x] Locally
+ [x] On a staging site
+ [ ] On a production website

+ Browser name and version

pondermatic:
## Description
Initialized shortcodes before the WP 'init' hook calls any other LifterLMS methods/functions.

Fixes #2070.
The `LLMS_Controller_Order` 'init' actions were being called before `LLMS_Shortcodes::init()` was called, which prevented shortcodes from being replaced in engagement emails.

Flow:
1. `lifterlms.php:49`
2. `LLMS_Loader::__construct():130`
3. `LLMS_Loader::includes():283`
4. `LLMS_Controller_Orders::__construct():41-43`
5. register order actions on 'init' hook priority 10
6. `llms(), lifterlms.php:65`
8. `LifterLMS::__construct():101`
9. register `LLMS_Shortcodes::init()` on 'init' hook priority 10

Options:
1. Register `LLMS_Controller_Orders` form processing actions with a priority higher than 10.
2. Register `LLMS_Shortcodes::init()` with a priority lower than 10.
3. Load `LLMS_Shortcodes` in `LLMS_Loader::includes()`, before loading `LLMS_Controller_Orders` AND add a constructor to `LLMS_Shortcodes` that registers `init()` with the 'init' hook priority 10

I was concerned that options 1 and 2 might cause a conflict with 3rd parties, so I implemented option 3.

## How has this been tested?
Manually and with a new unit test.

## Types of changes

## Checklist:

### DIFF
--- a/class-lifterlms.php
+++ b/class-lifterlms.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Main
  *
  * @since 1.0.0
- * @version 6.1.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -67,6 +67,7 @@ final class LifterLMS {
 	 * @since 4.13.0 Check site duplicate status on `admin_init`.
 	 * @since 5.3.0 Move the loading of the LifterLMS autoloader to the main `lifterlms.php` file.
 	 * @since 6.1.0 Automatically load payment gateways.
+	 * @since [version] Moved registration of `LLMS_Shortcodes::init()` with the 'init' hook to `LLMS_Shortcodes::__construct()`.
 	 *
 	 * @return void
 	 */
@@ -98,7 +99,6 @@ final class LifterLMS {
 		add_action( 'init', array( $this, 'init_session' ), 6 ); // After table installation which happens at init 5.
 		add_action( 'init', array( $this, 'payment_gateways' ) );
 		add_action( 'init', array( $this, 'include_template_functions' ) );
-		add_action( 'init', array( 'LLMS_Shortcodes', 'init' ) );
 
 		add_action( 'admin_init', array( 'LLMS_Site', 'check_status' ) );
 
--- a/includes/class-llms-loader.php
+++ b/includes/class-llms-loader.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Classes
  *
  * @since 4.0.0
- * @version 5.9.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -230,10 +230,14 @@ class LLMS_Loader {
 	 * @since 5.6.0 Include `LLMS_Prevent_Concurrent_Logins`.
 	 * @since 6.0.0 Included `LLMS_Block_Library`, `LLMS_Controller_Awards`, and `LLMS_Engagement_Handler`.
 	 *              Removed loading of class files that don't instantiate their class in favor of autoloading.
+	 * @since [version] Instantiated `LLMS_Shortcodes` before `LLMS_Controller_Orders`.
 	 *
 	 * @return void
 	 */
 	public function includes() {
+
+		// Instantiate LLMS_Shortcodes before LLMS_Controller_Orders.
+		new LLMS_Shortcodes();
 
 		// Functions.
 		require_once LLMS_PLUGIN_DIR . 'includes/llms.functions.core.php';
--- a/includes/class-llms-loader.php
+++ b/includes/class-llms-loader.php
@@ -230,14 +230,14 @@ class LLMS_Loader {
 	 * @since 5.6.0 Include `LLMS_Prevent_Concurrent_Logins`.
 	 * @since 6.0.0 Included `LLMS_Block_Library`, `LLMS_Controller_Awards`, and `LLMS_Engagement_Handler`.
 	 *              Removed loading of class files that don't instantiate their class in favor of autoloading.
-	 * @since [version] Instantiated `LLMS_Shortcodes` before `LLMS_Controller_Orders`.
+	 * @since [version] Included `LLMS_Shortcodes` before `LLMS_Controller_Orders`.
 	 *
 	 * @return void
 	 */
 	public function includes() {
 
 		// Instantiate LLMS_Shortcodes before LLMS_Controller_Orders.
-		new LLMS_Shortcodes();
+		require_once LLMS_PLUGIN_DIR . 'includes/shortcodes/class.llms.shortcodes.php';
 
 		// Functions.
 		require_once LLMS_PLUGIN_DIR . 'includes/llms.functions.core.php';
--- a/includes/shortcodes/class.llms.shortcodes.php
+++ b/includes/shortcodes/class.llms.shortcodes.php
@@ -619,4 +619,4 @@ class LLMS_Shortcodes {
 
 }
 
-new LLMS_Shortcodes();
+return new LLMS_Shortcodes();
--- a/includes/shortcodes/class.llms.shortcodes.php
+++ b/includes/shortcodes/class.llms.shortcodes.php
@@ -5,7 +5,6 @@
  * @package LifterLMS/Classes/Shortcodes
  *
  * @since 1.0.0
- * @since [version] Added self instantiation.
  * @version [version]
  */
 
--- a/includes/shortcodes/class.llms.shortcodes.php
+++ b/includes/shortcodes/class.llms.shortcodes.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Classes/Shortcodes
  *
  * @since 1.0.0
- * @version 6.0.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -17,6 +17,17 @@ defined( 'ABSPATH' ) || exit;
  * @since 4.0.0 Remove reliance on deprecated class `LLMS_Quiz_Legacy` & stop registering deprecated shortcode `[courses]` and `[lifterlms_user_statistics]`.
  */
 class LLMS_Shortcodes {
+
+	/**
+	 * Constructor.
+	 *
+	 * @since [version]
+	 * @return void
+	 */
+	public function __construct() {
+
+		add_action( 'init', array( 'LLMS_Shortcodes', 'init' ) );
+	}
 
 	/**
 	 * Initialize shortcodes array.
--- a/includes/shortcodes/class.llms.shortcodes.php
+++ b/includes/shortcodes/class.llms.shortcodes.php
@@ -5,6 +5,7 @@
  * @package LifterLMS/Classes/Shortcodes
  *
  * @since 1.0.0
+ * @since [version] Added self instantiation.
  * @version [version]
  */
 
@@ -617,3 +618,5 @@ class LLMS_Shortcodes {
 	}
 
 }
+
+new LLMS_Shortcodes();
--- a/.changelogs/init-shortcodes-early.yml
+++ b/.changelogs/init-shortcodes-early.yml
@@ -1,0 +1,5 @@
+significance: patch
+type: fixed
+links:
+  - "#2070"
+entry: Fixed an issue that caused shortcodes to not be replaced in some engagement emails.
--- a/includes/models/model.llms.user.certificate.php
+++ b/includes/models/model.llms.user.certificate.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Models/Classes
  *
  * @since 3.8.0
- * @version 6.0.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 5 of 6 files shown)


includes/class-llms-loader.php
```diff
@@ -230,11 +230,15 @@ public function autoload( $class ) {
...
 	 * @since 6.0.0 Included `LLMS_Block_Library`, `LLMS_Controller_Awards`, and `LLMS_Engagement_Handler`.
 	 *              Removed loading of class files that don't instantiate their class in favor of autoloading.
+	 * @since [version] Instantiated `LLMS_Shortcodes` before `LLMS_Controller_Orders`.
 	 *
 	 * @return void
 	 */
 	public function includes() {
 
+		// Instantiate LLMS_Shortcodes before LLMS_Controller_Orders.
+		new LLMS_Shortcodes();
```
thomasplevy: I know why this is done this way but instantiating a class in a method that only includes files seems like a break in an established pattern. I know the autoloader loads it and it performs the same function but I don't like it. I'd prefer to instantiate elsewhere instead of a method with an explicit purpose of "including files"

thomasplevy: I know why this is done this way but instantiating a class in a method that only includes files seems like a break in an established pattern. I know the autoloader loads it and it performs the same function but I don't like it. I'd prefer to instantiate elsewhere instead of a method with an explicit purpose of "including files"

class-lifterlms.php
```diff
@@ -98,7 +99,6 @@ private function __construct() {
 		add_action( 'init', array( $this, 'init_session' ), 6 ); // After table installation which happens at init 5.
 		add_action( 'init', array( $this, 'payment_gateways' ) );
 		add_action( 'init', array( $this, 'include_template_functions' ) );
-		add_action( 'init', array( 'LLMS_Shortcodes', 'init' ) );
```
thomasplevy, line 101: the core problem is that this is loaded at init 10 but too late? Couldn't we just change this to init 9 or 5 or something instead of the default 10 in that case?

includes/class-llms-loader.php
pondermatic: I don't like it either for the same reasons. However, the purpose of this method changed in 6.0.0 from "include a bunch of classes used by almost every page and instantiate some of those classes so they can hook into the system" to "include only the classes that instantiate themselves so they can hook into the system" (except `llms.functions.core.php` and `llms.template.hooks.php`). Loading the class file is not the goal. It's just a way to get to the goal of initializing the plug-in.

I'd like to rename this method, but it's public. Maybe we should deprecate this method and move its code to a private/protected "bootstrap" or "init" function/method? It belongs in the core `LifterLMS` class and should be called by the `init()` method.

I didn't want to go too deep on this because the timing of dependencies during startup can be tricky and time-consuming. So I "settled" for a method that, while doing essentially the same thing, isn't quite named accurately.

I would like to make it a rule that we never instantiate an object in a class definition file. That's a rule I've lived by forever, but I don't remember where I learned it. However, I need to think hard about how this change might affect 3rd parties using these classes.

class-lifterlms.php
pondermatic: You are correct. I was concerned that changing the priority might cause problems in 3rd party code (or even our own add-ons).

includes/class-llms-loader.php
pondermatic: From [PSR-1: Basic Coding Standard](https://www.php-fig.org/psr/psr-1/#23-side-effects):
> A file SHOULD declare new symbols (classes, functions, constants, etc.) and cause no other side effects, or it SHOULD execute logic with side effects, but SHOULD NOT do both.

class-lifterlms.php
thomasplevy: Yea you're right

includes/class-llms-loader.php
thomasplevy: > I would like to make it a rule that we never instantiate an object in a class definition file. That's a rule I've lived by forever, but I don't remember where I learned it. However, I need to think hard about how this change might affect 3rd parties using these classes.

This rule introduces a new standard that breaks a substantial (I'd say *most*) files in our codebases. So we're not going to introduce this as a rule. PSR-1 is great I guess (I don't know much about any of the established standards). This pattern breaks it. Okay. But we're not going to just change a bunch of stuff and create a new rule because of this.


> I'd like to rename this method, but it's public. Maybe we should deprecate this method and move its code to a private/protected "bootstrap" or "init" function/method? It belongs in the core LifterLMS class and should be called by the init() method.

The main `LifterLMS` class already has this bootstrapping for a bunch of classes and so on. And shortcodes are *already* being bootstrapped there. Edit: but of course that doesn't help since the orders are being loaded via the loader. I realize this. I'm just talking through the solution.

thomasplevy: If the shortcode class initializes itself (willfully breaking the PSR-1 standard but following our pre-established patterns) the require can take place in the `includes()` method without me feeling weird about it and I think we're probably good to go on these changes.

pondermatic: OK.

includes/shortcodes/class.llms.shortcodes.php
```diff
@@ -617,3 +618,5 @@ public static function pricing_table( $atts ) {
 	}
 
 }
+
+new LLMS_Shortcodes();
```
thomasplevy:
```suggestion
return new LLMS_Shortcodes();
```

thomasplevy: To follow the existing pattern more closely

includes/shortcodes/class.llms.shortcodes.php
```diff
@@ -5,7 +5,8 @@
  * @package LifterLMS/Classes/Shortcodes
  *
  * @since 1.0.0
- * @version 6.0.0
+ * @since [version] Added self instantiation.
```
thomasplevy:
```suggestion
```

Coding standards don't allow for a changelog in the fileheader (except for template files).

In this scenario we're okay to not document this change. It's okay.

thomasplevy (changes requested):

includes/class-llms-loader.php
pondermatic: > > I would like to make it a rule that we never instantiate an object in a class definition file. That's a rule I've lived by forever, but I don't remember where I learned it. However, I need to think hard about how this change might affect 3rd parties using these classes.


@thomasplevy wrote 
> This rule introduces a new standard that breaks a substantial (I'd say _most_) files in our codebases. So we're not going to introduce this as a rule. PSR-1 is great I guess (I don't know much about any of the established standards). This pattern breaks it. Okay. But we're not going to just change a bunch of stuff and create a new rule because of this.


When I said, "I would like to make it a rule ...", I didn't mean that I think the rule should start right now. I should have created a feature request issue instead of suggesting it in a PR. It would be a waste of time to change the pattern just because someone "likes" it a certain way, so I'll include justifications. 

Changing how we bootstrap the plugin from the current standard to the PSR-1 standard is a relatively simple change.
1. Remove the `return new CLASS();` from 44 files.
2. Replace `LLMS_Loader::includes()` that includes the files with `LLMS::init_classes()` (or something like that) that instantiates the classes. This method should be called during LifterLMS bootstrapping, possibly by `LLMS::init()`.

includes/shortcodes/class.llms.shortcodes.php
pondermatic: Please don't shoot the messenger.

The file header example changelog in the LifterLMS documentation standards has multiple `@since` tags with one that says "Description of file changes".
https://github.com/gocodebox/lifterlms/blob/trunk/docs/documentation-standards.md#file-headers-1

The commit is changing the file's behavior.

thomasplevy: https://github.com/gocodebox/lifterlms/blob/trunk/docs/documentation-standards.md#file-headers

And run PHPCS (which correctly reports the error. phpcs is not wrong in this case)

File header changelogs don't keep a changelog because it's *generally* redundant... You've found an exception to this rule and I've asked you to ignore it.

The example is probably a hold over from when we *did* (stupidly) keep redundant changelogs. *BUT* we do keep a changelog on file headers for certain files (templates and views, for example).

So.... apply my suggestion, please.

pondermatic: @thomasplevy, I think I understand what happened. When I run PHPCS, I usually use `composer check-cs-errors` instead of `composer check-cs`. With `check-cs-errors`, the `--error-severity=1 --warning-severity=6` options cause PHPCS to not report multiple `@since` tags in the file header as an error.

I probably got it in my head from the [pull request template](https://github.com/gocodebox/lifterlms/blob/trunk/.github/PULL_REQUEST_TEMPLATE.md?plain=1) that `check-cs-errors` should be used. I'll use `check-cs` from now on.

thomasplevy: No worries @pondermatic!

PR #2117, opened 2022-04-13
Introduce "Secure" string anonymization for payment gateways

thomasplevy:
## Description

This PR introduces `llms_anonymize_string()`, a function used to mask a string that might contain sensitive data, for example, convert an email address (`help@lifterlms.com`) to something like `***p@li*********om` which could "safely" recorded in a log file.

The primary use for this function would be to mask API keys that might get stored when debug logging is enabled for a payment gateway and a logged API request contains a plain-text API key in it's request headers (or query string or whatever).

Building on this method, a callback is attached to the filter in `llms_log()` which will allow 3rd parties to register "secure" strings that should be automatically anonymized when they are logged.

I've also added automatic registration of any payment gateway option with a `secure_option` key in the admin field setting. 

The result is that payment gateways will "automatically" have their secure options masked when logged via gateway logging methods.

This also provides an easy-to-use framework for developers (ourselves or others) to easily mask secure strings that might show up within logs using, probably, a single filter.

## How has this been tested?

+ New unit tests

## Types of changes

## Checklist:


includes/abstracts/abstract.llms.payment.gateway.php
```diff
@@ -569,6 +569,57 @@ public function get_logging_enabled() {
...
+	 * This is the callback for the `llms_secure_strings` filter (called via `llms_log()`).
+	 *
+	 * This method will load the value any gateway option with a `secure_option` declaration. Additional
+	 * strings can be added to the list using the `llms_get_gateway_secure_strings` filter.
+	 *
+	 * @since [version]
+	 *
+	 * @param string[] $strings Array of secure strings.
+	 * @param string   $handle  The log handle.
+	 * @return [type] [description]
```
eri-trabiccolo:
```suggestion
	 * @return string[]
```

includes/functions/llms.functions.log.php
```diff
@@ -197,12 +197,66 @@ function llms_log( $message, $handle = 'llms' ) {
...
+	 * Filters a list of "secure" strings which should be anonymized prior to logging.
+	 *
+	 * A plugin or theme that might log potentially sensitive data (such as API keys), the
+	 * API key strings can be registered with this filter to automatically be anonymized
+	 * if they are found within logs.
+	 *
+	 * @since [version]
+	 *
+	 * @param string[] $secure_strings An array of secure strings that should be anonymized.
+	 * @param string[] $handle         The log handle. This can be used to only register strings for a specific log file.
```
eri-trabiccolo:
```suggestion
	 * @param string   $handle         The log handle. This can be used to only register strings for a specific log file.
```

includes/functions/llms.functions.log.php
```diff
@@ -197,12 +197,66 @@ function llms_log( $message, $handle = 'llms' ) {
...
+	}
+
+	$find    = array();
+	$replace = array();
+
+	foreach ( $secure_strings as $string ) {
+		$message = is_array( $message ) || is_object( $message ) ? print_r( $message, true ) : $message; // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_print_r -- This is intentional.
+		if ( false !== strpos( $message, $string ) ) {
+			$find[]  = $string;
+			$replace = llms_anonymize_string( $string );
```
eri-trabiccolo:
```suggestion
			$find[]    = $string;
			$replace[] = llms_anonymize_string( $string );
```
right?
We could even use one associative array only?
```php
$find_replace = array();
foreach ( $secure_strings as $string ) {
[...]
		$find_replace[ $string ] = llms_anonymize_string( $string );
[...]
$message = empty( $find_replace ) ? $message : str_replace( array_keys( $find_replace ), array_values( $find_replace ), $message );
```

includes/functions/llms.functions.log.php
```diff
@@ -197,12 +197,66 @@ function llms_log( $message, $handle = 'llms' ) {
...
+	// Nothing to do.
+	if ( empty( $secure_strings ) ) {
+		return $message;
+	}
+
+	$find    = array();
+	$replace = array();
+
+	foreach ( $secure_strings as $string ) {
+		$message = is_array( $message ) || is_object( $message ) ? print_r( $message, true ) : $message; // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_print_r -- This is intentional.
```
eri-trabiccolo: This doesn't depend on the `foreach` so it can be moved outside, right?
Also we do this anyways here:
https://github.com/gocodebox/lifterlms/blob/6.3.0/includes/functions/llms.functions.log.php#L195-L198

So what if we add another filter just below that line?

`$message = apply_filters( 'llms_log_printable_message', $message, $handle );`

and we hook to that instead, also because I don't like the fact that we transform the type of the `$message` in this callback: we potentially get an array or an object and we always return a string.

eri-trabiccolo (changes requested):
Very nice, we needed this :D 
I remember we talked about this a while back.

includes/functions/llms.functions.log.php
thomasplevy: i thought about adding another filter for exactly this reason but having two filters for essentially the same thing felt stupid to me... but really I think you're right. I didn't consider that accepting an array and returning a string from a filter creates a possible confusing side effect. I'll rework it.

includes/functions/llms.functions.log.php
```diff
@@ -177,32 +177,102 @@ function llms_get_log_path( $handle ) {
...
+		 * before string conversion, use `llms_log_message`.
+		 *
+		 * @since [version]
+		 *
+		 * @see llms_log_message
+		 *
+		 * @param string $message Log message string.
+		 * @param string $handle  Allow creation of multiple log files by handle.
+		 */
+		$message = apply_filters( 'llms_log_message_string', $message, $handle );
```
thomasplevy, line 215: @eri-trabiccolo how about using this instead now?

eri-trabiccolo: Like it the better

includes/functions/llms.functions.log.php
```diff
@@ -177,32 +177,102 @@ function llms_get_log_path( $handle ) {
...
+ *
+ * This function is a callback for the `llms_log_message_string` filter. It loads secure strings
+ * defined in the `llms_secure_strings` filter and automatically anonymizes them when
+ * they are found within the supplied log message.
+ *
+ * @since [version]
+ *
+ * @access private
+ *
+ * @param string $message The data to log.
```
eri-trabiccolo:
```suggestion
 * @param string $message The string to log.
```

eri-trabiccolo (approved):
NIce, I haven't tested though :D

includes/abstracts/abstract.llms.payment.gateway.php
```diff
@@ -569,6 +569,57 @@ public function get_logging_enabled() {
 		return $this->get_option( 'logging_enabled' );
 	}
 
+	/**
+	 * Retrieves a list of "secure" strings which should be anonymized if they're found within debug logs.
+	 *
+	 * This is the callback for the `llms_secure_strings` filter (called via `llms_log()`).
+	 *
+	 * This method will load the value any gateway option with a `secure_option` declaration. Additional
```
pondermatic, line 577: This sentence confuses me.

includes/functions/llms.functions.log.php
```diff
@@ -177,32 +177,102 @@ function llms_get_log_path( $handle ) {
 function llms_log( $message, $handle = 'llms' ) {
 
 	/**
-	 * Filter a log message before it's written to the logger.
+	 * Filter a log data before it's written to the logger.
```
pondermatic, line 180:
```suggestion
	 * Filter log data before it's written to the logger.
```

includes/functions/llms.functions.log.php
```diff
@@ -131,17 +131,17 @@ function llms_get_callable_name( $callable ) {
 
 	if ( is_array( $callable ) && ! empty( $callable ) ) {
 
-		// Class and class method: [ $class, 'method' ].
+		// Class and class method: [ $class, 'method' ]. (phpcs:ignore Squiz.PHP.CommentedOutCode.Found).
```
pondermatic, line 134:
```suggestion
		// A class object + class method: [ $class, 'method' ].
```
Changes the percentage of tokens that look like code from 48% to 31%, which is under the default 40% threshold.

includes/functions/llms.functions.log.php
```diff
@@ -131,17 +131,17 @@ function llms_get_callable_name( $callable ) {
...
 		if ( is_object( $callable[0] ) ) {
 			return get_class( $callable[0] ) . '->' . $callable[1];
 		}
 
-		// Static class + method: [ 'class', 'method' ].
+		// Static class + method: [ 'class', 'method' ]. (phpcs:ignore Squiz.PHP.CommentedOutCode.Found).
```
pondermatic, line 139: The method is static, not the class.
The example should be a string, not an array.
```suggestion
		// A class name and a static method name: 'class' . '::' . 'method'.
```
Changes the percentage of tokens that look like code from 42% to 24%, which is under the default 40% threshold.

includes/functions/llms.functions.log.php
```diff
@@ -131,17 +131,17 @@ function llms_get_callable_name( $callable ) {
...
 		return implode( '::', $callable );
 
 	}
 
-	// Invokable class: $class.
+	// Invokable class: $class. (phpcs:ignore Squiz.PHP.CommentedOutCode.Found).
```
pondermatic, line 144:
```suggestion
	// An invokable class object: $class.
```
Changes the percentage of tokens that look like code from 43% to 20%, which is under the default 40% threshold.

pondermatic (changes requested):

pondermatic:
Dang, missed it by 8 minutes.

PR #2120, opened 2022-04-15
Stop using deprecated `FILTER_SANITIZE_STRING`.

eri-trabiccolo:
## Checklist:

### DIFF
--- a/.changelogs/filter_sanitize_string.yml
+++ b/.changelogs/filter_sanitize_string.yml
@@ -1,3 +1,3 @@
 significance: patch
 type: fixed
-entry: Stop using deprecated `FILTER_SANITIZE_STRING`.
+entry: Stop using the deprecated `FILTER_SANITIZE_STRING` constant.
--- a/.changelogs/filter_sanitize_string.yml
+++ b/.changelogs/filter_sanitize_string.yml
@@ -1,0 +1,3 @@
+significance: patch
+type: fixed
+entry: Stop using deprecated `FILTER_SANITIZE_STRING`.
--- a/includes/admin/post-types/post-tables/class-llms-admin-post-table-awards.php
+++ b/includes/admin/post-types/post-tables/class-llms-admin-post-table-awards.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Admin/PostTypes/PostTables/Classes
  *
  * @since 6.0.0
- * @version 6.0.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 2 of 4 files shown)


.changelogs/filter_sanitize_string.yml
```diff
@@ -0,0 +1,3 @@
+significance: patch
+type: fixed
+entry: Stop using deprecated `FILTER_SANITIZE_STRING`.
```
pondermatic:
```suggestion
entry: Stop using the deprecated `FILTER_SANITIZE_STRING` constant.
```

pondermatic (approved):
Excellent!

In all these cases, the string is used in a comparison or array search, so there's no need to replace `llms_filter_input( INPUT_GET, $var, FILTER_SANITIZE_STRING )` with [`llms_filter_input_sanitize_string( INPUT_GET, $var )`](https://developer.lifterlms.com/reference/functions/llms_filter_input_sanitize_string/).

pondermatic (approved):

pondermatic:
After more thinking, instead of removing `FILTER_SANITIZE_STRING`, should it be replaced with `FILTER_SANITIZE_FULL_SPECIAL_CHARS`? Removing the filter argument causes [`FILTER_DEFAULT`](https://www.php.net/manual/en/function.filter-input.php) to be used, which is the same thing as [`FILTER_UNSAFE_RAW`](https://www.php.net/manual/en/filter.filters.sanitize.php), which does nothing?

eri-trabiccolo:
yes, we're just comparing those input vars to other strings, we're not using them anywhere in the code, so I thought it's just fine so. Am I wrong?

thomasplevy:
@pondermatic it *does* do nothing and it doesn't really matter because the input is used only as a check. If we're storing / processing / etc... we'd want to sanitize but doing nothing is fine in many of these instances. 

Unless I'm wrong and you can provide a PoC displaying how doing nothing poses a security risk or creates a bug in some scenario.

This was discussed in https://github.com/gocodebox/lifterlms/pull/1948

I did a lot of research around the best implementation route and this is what I concluded (Rocco and I discussed this at length in slack conversations)

I'm happy to continue the conversation but I'll need some evidence of what problems we're solving (or creating) before changing code

pondermatic:
When I read that `FILTER_UNSAFE_RAW` does nothing, my first thought was "then why use `filter_input()` instead of `$_GET`? Now I remember that we do that to avoid PHPCS "input not sanitized" errors.

thomasplevy:
We also use `llms_filter_input()` (a wrapper around `filter_input()` so that we can mock things in testing.

And additionally you can call `filter_input()` without having to null coalesce or run `isset( $_GET[...] )` as it will just return `null` if the key isn't available in the super global.

PR #2129, opened 2022-04-22
Allow avoiding error returning when updating an llms post meta with the same value as stored in the db

### LINKED ISSUE #909: Should LLMS_Post_Model property update return false if the previous value equals the new one?
At the moment when updating an llms post model property which **is** a post_meta, if the new value equals the old value the `set` method return `false` (`set_bulk` can return a `WP_Error`).

This desn't happen if we update a model property which **is not** a post_meta, e.g. the `content` or the `title`. This is not consistent given the fact that we use the same "interface".

Ideally I should be able to use that method without concerning about the internal structure of that model, and I would expect a consistent behavior.

What do you think if we still return `true` when there was nothing to update?

eri-trabiccolo:
## Description

Fixes #909 and needed to fix https://github.com/gocodebox/lifterlms-rest/issues/222
This PR, aside from some code reorganization, just adds this piece of code:
https://github.com/gocodebox/lifterlms/pull/2129/files#diff-3f0337a1db5180a72d1b5434e90b08b7b0f4b7ee5997ac4dd50efa7b6d3c21a8R1482-R1497

## How has this been tested?
existing unit tests (doesn't introduce any drawback since the old behavior is the default one).
working on new unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/includes/abstracts/abstract.llms.post.model.php
+++ b/includes/abstracts/abstract.llms.post.model.php
@@ -1532,7 +1532,7 @@ abstract class LLMS_Post_Model implements JsonSerializable {
 			'html',
 			'float',
 		);
-		if ( in_array( $this->get_property_type( $key ), $string_types, true ) ) {
+		if ( in_array( $this->get_property_type( $key ), $scalar_types, true ) ) {
 			$sanitized = (string) $sanitized;
 		}
 
--- a/includes/abstracts/abstract.llms.post.model.php
+++ b/includes/abstracts/abstract.llms.post.model.php
@@ -1281,7 +1281,7 @@ abstract class LLMS_Post_Model implements JsonSerializable {
 			);
 		}
 
-		return $this->set_bulk( $model_array, $allow_same_meta_value );
+		return $this->set_bulk( $model_array, false, $allow_same_meta_value );
 
 	}
 
--- a/includes/abstracts/abstract.llms.post.model.php
+++ b/includes/abstracts/abstract.llms.post.model.php
@@ -1531,6 +1531,9 @@ abstract class LLMS_Post_Model implements JsonSerializable {
 			'yesno',
 			'html',
 			'float',
+			'int',
+			'bool',
+			'boolean',
 		);
 
 		if ( in_array( $this->get_property_type( $key ), $scalar_types, true ) ) {
--- a/includes/abstracts/abstract.llms.post.model.php
+++ b/includes/abstracts/abstract.llms.post.model.php
@@ -1161,7 +1161,7 @@ abstract class LLMS_Post_Model implements JsonSerializable {
 	 */
 	protected function scrub( $key, $val ) {
 		/**
-		 * Filters the property type being scrubbed
+		 * Filters the property type being scrubbed.
 		 *
 		 * The dynamic portion of this hook, `$this->model_post_type`, refers to the model's post type. For example "course",
 		 * "lesson", "membership", etc...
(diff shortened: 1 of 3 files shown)


eri-trabiccolo:
@gocodebox/engineering 
while I work on unit tests, if you got time, may I have your opinion on the strategy?
Thanks

includes/abstracts/abstract.llms.post.model.php
```diff
@@ -1256,54 +1258,91 @@ protected function scrub_field( $val, $type ) {
...
+
+	/**
+	 * Parse the LifterLMS post properties to set.
+	 *
+	 * Logic moved from `set_bulk()` method.
+	 *
+	 * @since [version]
+	 *
+	 * @param array $model_array Associative array of key/val pairs.
+	 * @return array|false False if nothing to set. An array that contains all the post properties and all the metas to set, otherwise.
```
```suggestion
	 * @param array $model_array Associative array of key/val pairs.
	 * @return array|bool Returns `false` if nothing to set or an array that contains all the post properties and all the metas to set.
```

thomasplevy (changes requested):
This looks pretty good to me. This is such a messy area of the codebase that this honestly seems to make it *more* messy to me but refactoring the code into separate methods is a good step towards making all of this a bit more testable and less fragile to maintain longterm.

I haven't physically run any tests but it looks solid given that existing unit tests are passing and the new tests you've added work. However, I'd like to see a bit more test coverage added here, specifically:

1) Usage of the new param in `set()` method isn't covered at all.
2) `empty_data` error condition in `set_bulk()` isn't covered 
3) `invalid_data` error condition in `set_bulk()` isn't covered
4) method `is_meta_value_same_as_stored_value()` is only tested for scalar values (strings). Tests could be written to test this method directly, ideally with every scalar type and at least one non scalar type being passed in. Using `LLMS_Unit_Test_Util::call_private_method()` so you can just pass stuff in directly.
5) You technically did not add anything related to setting post properties but there's no coverage on post properties here (I know other model tests pick it up but it wouldn't be too hard to add some post properties here while you're at it.

eri-trabiccolo:
> This looks pretty good to me. This is such a messy area of the codebase that this honestly seems to make it _more_ messy to me but refactoring the code into separate methods is a good step towards making all of this a bit more testable and less fragile to maintain longterm.

Yeah

> I haven't physically run any tests but it looks solid given that existing unit tests are passing and the new tests you've added work. However, I'd like to see a bit more test coverage added here, specifically:
> 
> 1. Usage of the new param in `set()` method isn't covered at all.

True.

> 2. `empty_data` error condition in `set_bulk()` isn't covered
> 3. `invalid_data` error condition in `set_bulk()` isn't covered

Yeah, I haven't added unit tests for _old_ code.
Tests added in 58b0165

> 4. method `is_meta_value_same_as_stored_value()` is only tested for scalar values (strings). Tests could be written to test this method directly, ideally with every scalar type and at least one non scalar type being passed in. Using `LLMS_Unit_Test_Util::call_private_method()` so you can just pass stuff in directly.

You're right.
Tested it un-directly with non scalar types in [184a13c](https://github.com/gocodebox/lifterlms/pull/2129/commits/184a13c2d90ce8b780714240383e7b216d3e3456)
and improved in [3abbdd9](https://github.com/gocodebox/lifterlms/pull/2129/commits/3abbdd95bb1b367609b8cd26a2afc4b38925f0e3)

> 5. You technically did not add anything related to setting post properties but there's no coverage on post properties here (I know other model tests pick it up but it wouldn't be too hard to add some post properties here while you're at it.

Because this PR didn't change the behavior on post properties: _old_ code.

I can add more tests coverage though <3

thomasplevy:
@eri-trabiccolo I realize that you didn't add tests for old code but since you're moving that old code and writing code all around it it's a good opportunity to add some tests to that old code.

eri-trabiccolo:
Yeah sorry, what I meant was that despite what it looks like the PR introduces very little changes, and those are tested.
But yeah improving the code coverage is always good you're right.

eri-trabiccolo:
@thomasplevy added various tests.

thomasplevy (approved):

PR #2148, opened 2022-05-11
Fix firing of the external engagement filters

pondermatic:
## Description

In LifterLMS 6.0.0, it is possible for the `lifterlms_external_engagement_handler_arguments` and `lifterlms_external_engagement_query_arguments` filters to not be fired.

### [LifterLMS before 6.0.0](https://github.com/gocodebox/lifterlms/blob/5.10.0/includes/class.llms.engagements.php) Logic
1. If the hook action or the engagement event type is not in a hardcoded switch statement, apply the filters, which may result in a parsed array with null values if the filter has no callables.

### [LifterLMS 6.0.0](https://github.com/gocodebox/lifterlms/blob/6.0.0/includes/class.llms.engagements.php#L522) Logic
1. If the hook action or the engagement event type are not registered, apply the filters and return the result, which may be a parsed array with null values if the filter has no callables.
3. Else, parse and return.

### This PR Logic
1. Always apply the filters and return the result if the parsed data is valid.
4. Else, if the hook action or the engagement event type are not registered, return a parsed array with null values.
5. Else parsing continues as it did before.

Fixes [gocodebox/lifterlms-integration-twilio#32](https://github.com/gocodebox/lifterlms-integration-twilio/issues/32).

## How has this been tested?
* Manually with the Twilio add-on.
* New LifterLMS core unit tests.
* Existing LifterLMS core unit tests.

## Types of changes
Bugfix.

## Checklist:

### DIFF
--- a/.changelogs/filter_parsed_engagements.yml
+++ b/.changelogs/filter_parsed_engagements.yml
@@ -1,0 +1,7 @@
+significance: patch
+type: fixed
+links:
+  - "https://github.com/gocodebox/lifterlms-integration-twilio/issues/32"
+entry: Fixed the ability for 3rd party plugins to use the
+  `lifterlms_external_engagement_handler_arguments` and
+  `lifterlms_external_engagement_query_arguments`.
--- a/.changelogs/filter_parsed_engagements.yml
+++ b/.changelogs/filter_parsed_engagements.yml
@@ -1,7 +1,7 @@
 significance: patch
 type: fixed
 links:
-  - "https://github.com/gocodebox/lifterlms-integration-twilio/issues/32"
+  - "gocodebox/lifterlms-integration-twilio#32"
 entry: Fixed the ability for 3rd party plugins to use the
   `lifterlms_external_engagement_handler_arguments` and
-  `lifterlms_external_engagement_query_arguments`.
+  `lifterlms_external_engagement_query_arguments` filters.
--- a/tests/phpunit/unit-tests/class-llms-test-engagements.php
+++ b/tests/phpunit/unit-tests/class-llms-test-engagements.php
@@ -32,6 +32,7 @@ class LLMS_Test_Engagements extends LLMS_UnitTestCase {
 			public $post_type = 'llms_mock_diploma';
 
 			public function __construct() {
+				register_post_type( 'llms_mock_diploma' );
 				add_filter( 'lifterlms_engagement_types', array( $this, 'register_engagement_types' ), 10, 1 );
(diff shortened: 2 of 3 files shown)


.changelogs/filter_parsed_engagements.yml
```diff
@@ -0,0 +1,7 @@
+significance: patch
+type: fixed
+links:
+  - "https://github.com/gocodebox/lifterlms-integration-twilio/issues/32"
```
thomasplevy:
```suggestion
  - "gocodebox/lifterlms-integration-twilio#32"
```

Only short references are considered valid:
+ `#123` (same repo)
+ `gocodebox/repo#123` different repo

Validate with: `npm run dev changelog validate`

.changelogs/filter_parsed_engagements.yml
```diff
@@ -0,0 +1,7 @@
...
+entry: Fixed the ability for 3rd party plugins to use the
+  `lifterlms_external_engagement_handler_arguments` and
+  `lifterlms_external_engagement_query_arguments`.
```
thomasplevy:
```suggestion
entry: Fixed the ability for 3rd party plugins to use the
  `lifterlms_external_engagement_handler_arguments` and
  `lifterlms_external_engagement_query_arguments` filters.
```

tests/phpunit/unit-tests/class-llms-test-engagements.php
```diff
@@ -12,6 +12,81 @@
...
+
+		$mock_engagements = new class {
+
+			public $engagement_action = 'llms_mock_curriculum_completed';
+
+			public $event_type = 'diploma';
+
+			public $handler_action = 'lifterlms_engagement_ship_diploma';
+
+			public $post_type = 'llms_mock_diploma';
```
thomasplevy, line 32: I think tests are failing because this post type isn't registered. Register it during setup and deregister during teardown and I think that'll fix the test errors.

thomasplevy (changes requested):
@pondermatic looks pretty solid, tests are failing though due to usage of an unregistered post type (I think). Notes below:

tests/phpunit/unit-tests/class-llms-test-engagements.php
pondermatic: Dang it! I saw this failure before I pushed the branch and fixed it (by using a known post type) before I opened the PR. Then I found another issue with the test and while fixing that one, re-introduced the first one. I didn't notice it because running PHPUnit on just this file works fine. But it fails when running the full suite. I need to keep a better eye on tests run by GitHub.

Thanks for your suggestion. It saved me some time.

PR #2128, opened 2022-04-21
Add WP Engine cache exclusions.

pondermatic:
## Description
1. Added the `LLMS_Cache_Helper::exclude_page_from_wpe_server_cache()` method that if we're on a live WP Engine site, excludes the page from the WP Engine server-side cache.
2. Updated `LLMS_Cache_Helper::maybe_no_cache()` to call `LLMS_Cache_Helper::exclude_page_from_wpe_server_cache()`.

See https://wpengine.com/support/cache/#Default_Cache_Exclusions.
See https://wpengine.com/support/determining-wp-engine-environment/

Fixes #1717.

## How has this been tested?
Manually locally, manually on WP Engine site, and with a new unit test.

## Types of changes
Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] My code has been tested.
- [x] My code passes all existing automated tests. 
- [x] My code follows the LifterLMS Coding & Documentation Standards.

PR #2154, opened 2022-05-18
Revert "Add WP Engine cache exclusions."

thomasplevy:
Reverts gocodebox/lifterlms#2128


thomasplevy:
@pondermatic I didn't look closely enough after your last changes -- tests are still failing and there's CS issues so I'm reverting #2128. Can you revisit this and open a new PR when tests and CS issues are addressed. Thanks

PR #2160, opened 2022-05-20
Orphaned or Un-purchasable Access Plans

### LINKED ISSUE #994: Checkout notice for already enrolled users should link to the product instead of the dashboard

> It’s just a bit tedious for users to have to go to their Dashboard instead of going straight to their course page (I’m talking about the link in “Visit your dashboard [here]”).

From https://wordpress.org/support/topic/add-filters-for-you-already-have-access-text/

Change the link in the message to link to the product (course or membership) permalink.

We should add a filter here as well (as requested by user in .org thread).

User has requested a filter on the text.

Here's the relevant lines of code:

https://github.com/gocodebox/lifterlms/blob/bd5810133f9ee53c687384aba1a8006d19ba5f33/includes/shortcodes/class.llms.shortcode.checkout.php#L69-L80

Related to (now closed) bug #607

pondermatic:
## Description
Do not display the checkout form if the access plan does not have an existing product, or the product has restrictions that prevent it from being purchasable, e.g. the course enrollment is at capacity.

Refactored `LLMS_Shortcode_Checkout::checkout()` by moving code that printed notices to their own methods and moving check logic to their model classes.

Fixes gocodebox/private-issues#8.
Fixes gocodebox/private-issues#9.
Fixes #994.

## How has this been tested?
Manually and with existing and new unit tests.

## Types of changes
Bugfix (non-breaking change which fixes an issue)

## Checklist:


includes/shortcodes/class.llms.shortcode.checkout.php
```diff
@@ -114,6 +114,26 @@ private static function confirm_payment( $atts ) {
...
+	 *
+	 * The default LifterLMS 'product/not-purchasable.php' template prints error notices if a related course
+	 * is restricted, e.g. if enrollment capacity has been reached or we're not in the enrollment period.
+	 *
+	 * @since [version]
+	 *
+	 * @param LLMS_Product $product The product object.
+	 * @return bool
+	 */
+	private static function does_product_not_purchasable_template_print_error_notices( $product ) {
```
eri-trabiccolo: This is `ingenious` although I must say that I don't really like it ehehe.

I see it's to reduce redundancy, but loading a "front" template, where accidentally there's some business logic we need here, doesn't seem the right solution to me, not only for a formal matter, but because templates are can be overridden by themes and 3rd party plugins, they're not that reliable.
Can't we just move the checks we need into functions/methods and use them in both places (here and the template)?

eri-trabiccolo: This is `ingenious` although I must say that I don't really like it ehehe.

I see it's to reduce redundancy, but loading a "front" template, where accidentally there's some business logic we need here, doesn't seem the right solution to me, not only for a formal matter, but because templates are can be overridden by themes and 3rd party plugins, they're not that reliable.
Can't we just move the checks we need into functions/methods and use them in both places (here and the template)?

eri-trabiccolo (changes requested):
Looks quite good to me, but see my _objection_ for the `does_product_not_purchasable_template_print_error_notices()` logic.

Also, since you're there, can you include a fix for this as well?
https://github.com/gocodebox/lifterlms/issues/994

includes/shortcodes/class.llms.shortcode.checkout.php
thomasplevy: This might be what we're looking for @pondermatic: https://developer.lifterlms.com/reference/functions/llms_notice_count/

pondermatic:
@gocodebox/engineering,

I have refactored all of the logic that verifies if the checkout form should be displayed and prints notices if not. I feel it is much better than my previous attempt, but I am a little bit unhappy that the "verify_X" methods not only return true or false but can also print a notice. Generally, a function that does two things should be split, but then the logic that decides whether to print the notice or not must be duplicated. In this case, it's in the [shortcode checkout class](https://github.com/gocodebox/lifterlms/blob/2198bb77617ac76bbd0e41b05e6a51f7eed44b33/includes/shortcodes/class.llms.shortcode.checkout.php#L393) and the [not-purchasable template](https://github.com/gocodebox/lifterlms/blob/80676a70eb6197b19477fd5f8cf9a94abbc4e991/templates/product/not-purchasable.php#L20).

Do you think the "not purchasable" template should also verify if the user is enrolled, that the product exists and that the plan is available to the user (user is in a required membership)? Right now it only checks the course enrollment period start/stop and the course capacity. If so, the template can call [`LLMS_Shortcode_Checkout::verify_checkout_form_is_displayable()`](https://github.com/gocodebox/lifterlms/blob/2198bb77617ac76bbd0e41b05e6a51f7eed44b33/includes/shortcodes/class.llms.shortcode.checkout.php#L393) and all the little verify methods can be changed to private.

I changed this PR to draft mode while I work on issue #994.

pondermatic:
To help answer the question, "Do you think the `not-purchasable` template should also ...":

#### Flow
The `not-purchasable` template is displayed from the `pricing-table` template, which is displayed from the `content-single-course-after` and `content-single-membership-after` templates (and by the possibly unused `LLMS_Shortcodes::pricing_table()` method).

### Verify User Is Not Enrolled
The `pricing-table` template does not load the `not-purchasable` template if the user is already enrolled. This check is needed by the `not-purchasable` template, but it wouldn't cause a problem if is performed.

### Verify Product Exists
If for some strange reason, `lifterlms_template_pricing_table()` can not get a post ID that exists, we get `PHP Warning:  Attempt to read property "post_type" on null in C:\wordpress-develop\src\wp-content\plugins\lifterlms\includes\abstracts\abstract.llms.post.model.php on line 452` when the `non-purchasble` template tries to get the product type. This check is not normally needed by the `not-purchasable` template, but it wouldn't hurt to have it.

### Plan Is Available
The `pricing-table` template does not load the `not-purchasable` template if the product is purchasable (and the user is not enrolled and the plan has no restrictions). Without this check in the `not-purchasable` template, the enroll button links to the membership page. This check is not needed by the `not-purchasable` template.

**My conclusion is that the extra checks are not really needed by the `not-purchasable` template.**

includes/models/model.llms.course.php
```diff
@@ -598,6 +598,40 @@ public function has_capacity() {
...
+	 * Returns true if the enrollment end date has passed, the course does not have an enrollment period,
+	 * or the course does not have an enrollment end date.
+	 *
+	 * @since [version]
+	 *
+	 * @return bool
+	 */
+	public function has_enrollment_period_ended() {
+
+		if ( 'no' === $this->get( 'enrollment_period' ) || empty( $this->get( 'enrollment_end_date' ) ) ) {
```
thomasplevy, line 611:
```suggestion
		if ( llms_parse_bool( $this->get( 'enrollment_period' ) ) || empty( $this->get( 'enrollment_end_date' ) ) ) {
```

thomasplevy:
```suggestion
		if ( llms_parse_bool( $this->get( 'enrollment_period' ) ) || empty( $this->get( 'enrollment_end_date' ) ) ) {
```

includes/models/model.llms.course.php
```diff
@@ -598,6 +598,40 @@ public function has_capacity() {
 
 	}
 
+	/**
+	 * Returns true if the enrollment end date has passed, the course does not have an enrollment period,
+	 * or the course does not have an enrollment end date.
```
thomasplevy: This description is in conflict with what the method actually does

If the course doesn't have an enrollment period or end date specified it returns false

includes/models/model.llms.course.php
```diff
@@ -598,6 +598,40 @@ public function has_capacity() {
...
+	 *
+	 * @since [version]
+	 *
+	 * @return bool
+	 */
+	public function has_enrollment_period_ended() {
```
eri-trabiccolo, line 609: You're adding here two public methods, shouldn't they be logged in the changelog?

includes/models/model.llms.product.php
```diff
@@ -46,6 +46,20 @@ class LLMS_Product extends LLMS_Post_Model {
...
 	protected $model_post_type = 'product';
 
+	/**
+	 * Returns true if this product's post exists, else returns false.
+	 *
+	 * @since [version]
+	 *
+	 * @return bool
+	 */
+	public function exists() {
```
eri-trabiccolo, line 56: to log in the changelog?

templates/product/not-purchasable.php
```diff
@@ -5,29 +5,20 @@
...
-	<?php if ( ! $course->has_capacity() ) : ?>
-		<?php llms_print_notice( $course->get( 'capacity_message' ), 'error' ); ?>
-	<?php endif; ?>
-
-	<?php
-endif;
+	if ( LLMS_Shortcode_Checkout::verify_course_enrollment_has_started( $course ) ) {
+		LLMS_Shortcode_Checkout::verify_course_enrollment_has_not_ended( $course );
+	}
+	LLMS_Shortcode_Checkout::verify_course_enrollment_has_capacity( $course );
```
eri-trabiccolo: I don't know, I somewhat don't like the idea that those verify methods also print notices.
This template now "hides" the printing logic and makes me wonder whether it has all this sense to keep it.
I know we want to avoid double checks, a reduce complexity but I don't like this, but I'm not in the position to say that it shouldn't be done this way, I think we have to wait for Thomas.

Maybe the best thing would be a method/function like [`llms_page_restricted`](https://developer.lifterlms.com/reference/functions/llms_page_restricted/) that does all the checks and returns an array with the not-purchasable reason and message, and this template only checks whether the reason is within `{not_ended|has_capacity}` and then print the notice. What do you think?

I'm not even sure why the various `verify_course_enrollment_*` methods should be part of the `LLMS_Shortcode_Checkout` class. They're used by that shortcode but also here in the `not-purchasable.php` template that is included from the pricing table. Can't they be `products`'s methods?

includes/shortcodes/class.llms.shortcode.checkout.php
```diff
@@ -298,6 +392,165 @@ protected static function setup_plan_and_form_atts( $plan_id, $atts ) {
...
+	 *
+	 * @since [version]
+	 *
+	 * @param LLMS_Course $course The course object.
+	 * @return bool
+	 */
+	public static function verify_course_enrollment_has_not_ended( $course ) {
+
+		if ( $course->has_enrollment_period_ended() ) {
+			llms_print_notice( $course->get( 'enrollment_closed_message' ), 'error' );
```
eri-trabiccolo: This kind of message contains a shortcode, e.g. 
`Enrollment in this course closed on [lifterlms_course_info id="123" key="enrollment_end_date"].`
So shortcodes need to be parsed.
Basically if you use the pricing table shortcode from outside a course, e.g. in a page, shortcodes inside that message are not parsed, same thing when you display that notice on the checkout page (at least from my tests).
The reason why the shortcodes in the notices are parsed when from within a course content is because:
1) (if the pricing table is added via a shortcode in a course content) see `do_shortcode()` here https://github.com/gocodebox/lifterlms/blob/6.6.0/includes/functions/llms-functions-content.php#L59
2) (if the pricing table is added via a block) [`do_blocks` (priority 9)](https://github.com/WordPress/WordPress/blob/6.0/wp-includes/default-filters.php#L184) applied before `[do_shortcodes` (priority 11)](https://github.com/WordPress/WordPress/blob/6.0/wp-includes/default-filters.php#L603)

eri-trabiccolo (changes requested):

eri-trabiccolo:
Oh another thing, I wonder if `admins` should always see the checkout form, even when the course is not purchasable because it's closed ecc. Like if a course enrollment is not opened yet, should the admin be able to reach the checkout anyways? I guess so, then the [view manager](https://lifterlms.com/docs/lifterlms-preview-management/) can be used to switch context.

eri-trabiccolo:
> Do you think the "not purchasable" template should also verify if the user is enrolled, that the product exists and that the plan is available to the user (user is in a required membership)? Right now it only checks the course enrollment period start/stop and the course capacity. If so, the template can call [LLMS_Shortcode_Checkout::verify_checkout_form_is_displayable()](https://github.com/gocodebox/lifterlms/blob/2198bb77617ac76bbd0e41b05e6a51f7eed44b33/includes/shortcodes/class.llms.shortcode.checkout.php#L393) and all the little verify methods can be changed to private.

> To help answer the question, "Do you think the `not-purchasable` template should also ...":
> 
> #### Flow
> The `not-purchasable` template is displayed from the `pricing-table` template, which is displayed from the `content-single-course-after` and `content-single-membership-after` templates (and by the possibly unused `LLMS_Shortcodes::pricing_table()` method).
> 
> ### Verify User Is Not Enrolled
> The `pricing-table` template does not load the `not-purchasable` template if the user is already enrolled. This check is needed by the `not-purchasable` template, but it wouldn't cause a problem if is performed.
> 
> ### Verify Product Exists
> If for some strange reason, `lifterlms_template_pricing_table()` can not get a post ID that exists, we get `PHP Warning: Attempt to read property "post_type" on null in C:\wordpress-develop\src\wp-content\plugins\lifterlms\includes\abstracts\abstract.llms.post.model.php on line 452` when the `non-purchasble` template tries to get the product type. This check is not normally needed by the `not-purchasable` template, but it wouldn't hurt to have it.
> 
> ### Plan Is Available
> The `pricing-table` template does not load the `not-purchasable` template if the product is purchasable (and the user is not enrolled and the plan has no restrictions). Without this check in the `not-purchasable` template, the enroll button links to the membership page. This check is not needed by the `not-purchasable` template.
> 
> **My conclusion is that the extra checks are not really needed by the `not-purchasable` template.**

I'm not sure.
I think that the logic, until now, was to made the required checks before calling the template. So this is the status-quo, we can change it of course, but why?
Once you know the rules, if you want to call the function `lifterlms_template_pricing_table()`, you have to make sure to run the required checks before.

includes/models/model.llms.course.php
pondermatic: The class changelog? I thought that new class methods were **not** mentioned in the class changelog because it is redundant with the new method's changelog.

eri-trabiccolo: nope, I meant in `.changelogs/`

thomasplevy: No, this kind of change should *not* be recorded.

The published changelogs are meant for users, primarily. We *can* add things like "introduced new filter" or function / method. But really that clutters up the changelog with information users don't care about.

A good changelog describes *what new features were added* and what *bugs were fixed*

includes/models/model.llms.product.php
thomasplevy: no, see above

includes/shortcodes/class.llms.shortcode.checkout.php
pondermatic: I added a call to `do_shortcode()` in `LLMS_Course::is_enrollment_restricted()`.

pondermatic:
@eri-trabiccolo wrote,
> Oh another thing, I wonder if `admins` should always see the checkout form, even when the course is not purchasable because it's closed ecc. Like if a course enrollment is not opened yet, should the admin be able to reach the checkout anyways? I guess so, then the [view manager](https://lifterlms.com/docs/lifterlms-preview-management/) can be used to switch context.

I believe that admins **should** be able to preview the checkout form, even when the checkout form is not displayable because of restrictions.

@thomasplevy, should we roll this feature into this PR and or create a new feature request issue?

eri-trabiccolo:
@pondermatic 
is this ready for the review or not? I guess you're waiting for thomas' feedback on the admins privileges right? If so shall we turn this into Draft until it's ready?

thomasplevy:
I'm far out of context of this PR and I won't have time to fully review or make decisions around this for a bit. If you two can't come up with a solution together let's put this on the back burner and I'll revisit in a few weeks.

includes/models/model.llms.course.php
```diff
@@ -663,6 +697,42 @@ public function is_enrollment_open() {
...
+		$restricted_message = false;
+		if ( ! $this->has_enrollment_period_started() ) {
+			$restricted_message = $this->get( 'enrollment_opens_message' );
+		}
+		if ( $this->has_enrollment_period_ended() ) {
+			$restricted_message = $this->get( 'enrollment_closed_message' );
+		}
+		if ( ! $this->has_capacity() ) {
+			$restricted_message = $this->get( 'capacity_message' );
+		}
```
thomasplevy, line 719: I think the behavior here is different than previously found in the `not-purchaseable` template.

The first two conditions change from an if / elseif to an if / if, which doesn't really matter but the last condition could be true in addition to one of the two enrollment period conditions.

The template *could* have two different messages shown -- one for one of the two enrollment period conditions and a second message if the course capacity is also exhausted.

After this change though the first message would get lost and only the latter (capacity) message would be displayed.

brianhogg:
Closing, can re-open if this comes up.

PR #2176, opened 2022-06-07
Implement new `gateway feature: modify_recurring_payments`

### LINKED ISSUE #2173: Implement new gateway feature: `modify_recurring_payments`

eri-trabiccolo:
## Description
Fixes #2173

## How has this been tested?
manually (unit tests on the way)

## Types of changes

## Checklist:

### DIFF
--- a/includes/admin/post-types/meta-boxes/class.llms.meta.box.order.submit.php
+++ b/includes/admin/post-types/meta-boxes/class.llms.meta.box.order.submit.php
@@ -126,22 +126,18 @@ class LLMS_Meta_Box_Order_Submit extends LLMS_Admin_Metabox {
 			}
 		}
 
-		/**
-		 * Order is important -- if both trial and next payment are updated
-		 * they should be saved in that order since next payment date
-		 * is automatically recalculated by trial end date update.
-		 */
 		$editable_dates = array(
-			'_llms_date_trial_end',
-			'_llms_date_next_payment',
 			'_llms_date_access_expires',
 		);
 
-		// Do not save recurring payments related dates if order's gateway do not support recurring payments modification.
-		if ( ! $order->supports_modify_recurring_payments() ) {
-			$editable_dates = array(
-				'_llms_date_access_expires',
-			);
+		// Save recurring payments related dates if order's gateway supports recurring payments modification.
+		if ( $order->supports_modify_recurring_payments() ) {
+			/**
+			 * Order is important -- if both trial and next payment are updated
+			 * they should be saved in that order since next payment date
+			 * is automatically recalculated by trial end date update.
+			 */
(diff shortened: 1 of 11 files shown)


eri-trabiccolo:
@thomasplevy 
So I did as discussed:
https://github.com/gocodebox/lifterlms-gateway-paypal/issues/52#issuecomment-1148825190
but I kept https://github.com/gocodebox/lifterlms/blob/1579e1185bc406e6d0dab0464ec4a35e6e2016ed/includes/models/model.llms.order.php#L1884
I think it makes sense to keep it, as that it's both a gateway and order "support".

Fine for you?

thomasplevy:
@eri-trabiccolo yea that makes sense

includes/admin/post-types/meta-boxes/class.llms.meta.box.order.submit.php
```diff
@@ -136,6 +137,13 @@ public function save( $post_id ) {
 			'_llms_date_access_expires',
 		);
 
+		// Do not save recurring payments related dates if order's gateway do not support recurring payments modification.
+		if ( ! $order->supports_modify_recurring_payments() ) {
+			$editable_dates = array(
+				'_llms_date_access_expires',
+			);
+		}
+
```
thomasplevy, line 142: I'd rather see the array start as the dates that are always editable (access_expires) and then add the conditional ones later if the order supports recurring payments in favor of overwriting / redeclaring the array with only the expires date when the order cannot modify recurring payments.

thomasplevy (changes requested):
This looks good to me, a minor piece of feedback. I haven't thoroughly reviewed all of the tests but the code seems solid.

includes/admin/post-types/meta-boxes/class.llms.meta.box.order.submit.php
eri-trabiccolo: Ok, makes sense.

PR #2182, opened 2022-06-10
AJAX Payment Source Switch action

thomasplevy:
## Description

Builds on #2102 to introduce an AJAX endpoint for gateways to utilize for AJAX payment source switching.

Additionally refactors a bit of code related to payment source switching.

## How has this been tested?

+ Unit tests

I'll be utilizing the new AJAX endpoint for PayPal and will do further manual testing as I write the integration / js code.

## Types of changes

+ New features
+ Deprecation (with replacement) of the location of a handler method. This shouldn't affect anyone but it is backwards compatible.

## Checklist:


includes/controllers/class-llms-controller-checkout.php
```diff
@@ -373,10 +380,211 @@ public function get_url( $action ) {
...
+		$order_id = llms_filter_input( INPUT_POST, 'order_id', FILTER_SANITIZE_NUMBER_INT );
+		if ( ! $order_id ) {
+			return new WP_Error( 'switch-source-order-missing', __( 'Missing order information.', 'lifterlms' ), 'error' );
+		}
+
+		$order = llms_get_post( $order_id );
+		if ( ! is_a( $order, 'LLMS_Order' ) || get_current_user_id() !== $order->get( 'user_id' ) ) {
+			return new WP_Error( 'switch-source-order-invalid', __( 'Invalid order.', 'lifterlms' ), 'error' );
+		}
+
```
eri-trabiccolo, line 523: Should we add a check on whether the current user can switch payment: current use is the order "owner"?

includes/models/model.llms.order.php
```diff
@@ -432,19 +432,26 @@ public function can_be_retried() {
 	}
 
 	/**
-	 * Determine if an order can be resubscribed to
+	 * Determines if a user can resubscribe to an inactive recurring payment order.
```
eri-trabiccolo: May I be nitpicking here?
There's actually no user involved in the logic here.

eri-trabiccolo (changes requested):

includes/controllers/class-llms-controller-checkout.php
thomasplevy: you mean like the second condition on this if statement?

`|| get_current_user_id() !== $order->get( 'user_id' )`

eri-trabiccolo: Yeah that one, idiot me. Sorry.

thomasplevy: heheh <3

PR #2184, opened 2022-06-15
Prerelease Version 7.0.0

thomasplevy:


includes/models/model.llms.order.php
```diff
@@ -1708,17 +1708,17 @@ public function maybe_schedule_retry() {
 
 		}
 
-		$datetime = date_i18n( 'Y-m-d H:i:s', current_time( 'timestamp' ) + $current_rule['delay'] );
+		$timestamp = current_time( 'timestamp' ) + $current_rule['delay'];
 
-		$this->set_date( 'next_payment', date( 'Y-m-d H:i:s', $datetime ) );
+		$this->set_date( 'next_payment', date( 'Y-m-d H:i:s', $timestamp ) );
```
eri-trabiccolo: Because of my wrong suggestion... See https://github.com/gocodebox/lifterlms/commit/f370fbbe4f411b212bbde7390547ce2dc5a0e710#r78813530

eri-trabiccolo: Also consider that this is potentially a breaking change...
We used to save the `date_i18n()`, now we save the `date()`?

thomasplevy: yea yea yea okay

eri-trabiccolo (approved):
I tested this a bit with some purchases and so on. It works good to me.

PR #2185, opened 2022-06-17
Array settings fields

thomasplevy:
## Description

Adds handling for admin settings fields that nest options into an associative array.

We already have support for checkboxes to do this, this PR introduces the ability to do this with "regular" fields (like email, text, selects, etc...).

This will allow settings field code like this:

```php
$fields = array(
	array(
		'type' => 'text',
		'id'   => 'my_field[one]',
	),
	array(
		'type' => 'text',
		'id'   => 'my_field[two]',
	)
);
```

To save an associative array option:

```php
get_option( 'my_field' ) === array( 'one' => 'val_one', 'two' => 'val_two' )
```

This can be useful when creating options which will generally be used together, they can be retrieved together with a single database request / function call.

I've deprecated a (previously undocumented) action as that action should have likely been initially introduced as a filter since the action itself cannot act on the supplied argument.

## How has this been tested?

+ New unit tests -- I wrote a bunch of tests before modifying the code and then improved the code to ensure tests continue to pass for unaffected options

+ I've tested the checkbox code against the PDF plugin options which utilize the checkbox functionality already

+ I've added new options to the PayPal gateway which utilize the new code for a series of select fields.

## Types of changes

## Checklist:

### DIFF
--- a/includes/admin/class.llms.admin.settings.php
+++ b/includes/admin/class.llms.admin.settings.php
@@ -999,10 +999,9 @@ class LLMS_Admin_Settings {
 					break;
 
 				default:
-
 					/**
 					 * Action run for external field types.
-					 * 
+					 *
 					 * @since Unknown
 					 * @deprecated [version] Use `llms_update_option_{$type}` filter hook instead.
 					 *
@@ -1011,14 +1010,14 @@ class LLMS_Admin_Settings {
 					do_action_deprecated( "lifterlms_update_option_{$type}", array( $field ), '[version]' );
 
 			}
-			
+
 			/**
 			 * Filters the value of a settings field after it has been parsed and sanitized
 			 * and before it is saved to the database.
 			 *
 			 * The dynamic portion of this hook, `{$type}` refers to the setting field type:
 			 * email, text, checkbox, etc...
-			 * 
+			 *
 			 * @since [version]
 			 *
 			 * @param string|null $option_value The sanitized option value or `null`.
@@ -1059,7 +1058,7 @@ class LLMS_Admin_Settings {
 			 *
 			 * An update isn't guaranteed after this action if the method's logic can't
 			 * find a valid posted valued to persist to the database.
-			 * 
+			 *
 			 * @since Unknown
 			 *
 			 * @param array $field The admin setting field array to be updated.
@@ -1080,7 +1079,7 @@ class LLMS_Admin_Settings {
 	}
 
 	/**
-	 * Retrieves the posted value for an array type setting field. 
+	 * Retrieves the posted value for an array type setting field.
 	 *
 	 * @since [version]
 	 *
--- a/includes/admin/class.llms.admin.settings.php
+++ b/includes/admin/class.llms.admin.settings.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Admin/Classes
  *
  * @since 1.0.0
- * @version 5.9.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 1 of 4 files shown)


includes/admin/class.llms.admin.settings.php
```diff
@@ -902,13 +902,14 @@ public static function get_option( $option_name, $default = '' ) {
 	/**
 	 * Save admin fields.
 	 *
-	 * Loops though the lifterlms options array and outputs each field.
+	 * Loops though a LifterLMS settings field options array and saves the values via `update_option()`.
```
eri-trabiccolo:
```suggestion
	 * Loops through a LifterLMS settings field options array and saves the values via `update_option()`.
```

includes/admin/class.llms.admin.settings.php
```diff
@@ -982,74 +976,125 @@ public static function save_fields( $settings ) {
...
+	 * @since [version]
+	 *
+	 * @param string $id The field ID, eg: "my_setting[field_one]".
+	 * @return string Returns the (sanitized) posted value or an empty string if it wasn't posted.
+	 */
+	private static function get_array_field_posted_value( $id ) {
+
+		parse_str( $id, $parsed_id );
+		$opt_id  = current( array_keys( $parsed_id ) );
+		$opt_key = key( $parsed_id[ $opt_id ] );
```
eri-trabiccolo:
```suggestion
		$opt_id  = key( $parsed_id );
		$opt_key = key( $parsed_id[ $opt_id ] );
```

includes/admin/class.llms.admin.settings.php
```diff
@@ -982,74 +976,125 @@ public static function save_fields( $settings ) {
...
-					break;
+					/**
+					 * Action run for external field types.
+					 *
+					 * @since Unknown
+					 * @deprecated [version] Use `llms_update_option_{$type}` filter hook instead.
+					 *
+					 * @param type $arg Description.
+					 */
+					do_action_deprecated( "lifterlms_update_option_{$type}", array( $field ), '[version]' );
```
eri-trabiccolo, line 1010: I guess you haven't added the "replacement" here because it's not a true replacement since the new hook is a filter hook. Right?

tests/phpunit/unit-tests/admin/class-llms-test-admin-settings.php
```diff
@@ -0,0 +1,281 @@
...
+		$this->mockPostRequest( array(
+			'mock' => '1',
+		) );
+		$res = LLMS_Admin_Settings::save_fields( $fields );
+		$this->assertSame( $actions, did_action( 'lifterlms_update_option' ) );
+
+	}
+
+
+}
```
eri-trabiccolo:
```suggestion

}

```
well this just because you know, I was there.

eri-trabiccolo (changes requested):
Minor change requested, nothing crucial.
Looks good to me.

includes/admin/class.llms.admin.settings.php
thomasplevy: yea i thought three things when I read through here:

Oh m**k...
That should be a filter
It should run against all types, not just custom types.

Realistically the action works just fine and doesn't need to be deprecated. But it will require a developer to perform their own save logic and call update_action() on their own in favor of just returning a value and letting us do it.

eri-trabiccolo: Yeah.

eri-trabiccolo: Sorry, I meant just:
Yeah

PR #2204, opened 2022-07-11
Convert `LLMS.Spinner` to an `@lifterlms/components` module and reduce necessity of jQuery when using

thomasplevy:
## Description

Moves the javascript found at `LLMS.Spinner` to a component in the `@lifterlms/components` package.

The new component can be imported: `import { Spinner } from `@lifterlms/components`

It can be accessed (backwards compatible safe) via `window.LLMS.Spinner`.

The functions (`LLMS.Spinner.start()`, etc...) can now be used without the necessity of passing in a jQuery selection. For example, `LLMS.Spinner.start( '.my-spinner-wrapper' )` will now function. Previously you'd have to do `LLMS.Spinner.start( $( '.my-spinner-wrapper' ) )`. This is a subtle change but makes it so newer JS we write which increasingly does not utilize any jQuery doesn't have to include jQuery to use the built-in spinner. The component will continue to require jQuery though. I was trying to figure out a way to make it not need jQuery and remain backwards compatible but I can't really figure that one out. There's probably a pretty easy way to do it but it's not necessary for now.

Additionally this moves all the CSS for the HTML elements into JS. The CSS is automatically "loaded" and appended to the `<head>` as a `<style>` element when a spinner component is loaded. This is a technically problematic technique for any custom code which might rely on CSS precedence to override the default styles. If the custom code simply loads it's CSS overrides *after* the LifterLMS core stylesheet where this CSS was previously located it would override. Now they'd need to add more specific selectors to override. But loading the CSS this way makes it so that we can unbundle the component from the LifterLMS core styles and use the component in places where the LifterLMS core css isn't loaded. It makes it more generally useable while sacrificing a potential issue for themers. I'd intend to release this with 7.0 so we can note the breaking change with the major release. I think it's a niche problem.

## How has this been tested?

+ Manually

## Types of changes

+ Update / enhancement

## Checklist:


packages/components/src/spinner/index.js
```diff
@@ -0,0 +1,90 @@
...
+ *                                                  array of DOM Elements. Each element in the list will have it's spinner started. If a spinner doesn't
+ *                                                  exist within the element, it will be appended and then started.
+ * @param {string}                         size     Size or the spinner element. Accepts "default" (40px) or "small" (20px).
+ * @return {void}
+ */
+export function start(selector, size = SIZE_DEFAULT) {
+	ensureElementList(selector).forEach((el) => {
+		const spinner = get(el, size, false);
+		if (spinner) {
+			spinner.style.display = 'block';
```
eri-trabiccolo, line 71: I like this smooth animation!

packages/components/src/spinner/styles.js
```diff
@@ -0,0 +1,53 @@
+import { WRAPPER_CLASSNAME, CLASSNAME, SIZE_SMALL } from './constants';
+
+/**
+ * CSS Styles for the components.
+ *
+ * @type {string}
+ */
+export const STYLES = `
```
eri-trabiccolo, line 8: this is cool but it's not "autoprefixed", you know...

packages/components/src/spinner/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+	const nodeList = ensureElementList(selector);
+	if (!nodeList.length) {
+		return null;
+	}
+
+	const wrapper = nodeList[0],
+		spinner =
+			wrapper.querySelector(`.${WRAPPER_CLASSNAME}`) ||
+			create(wrapper, size);
```
eri-trabiccolo: This is not "enough": `wrapper.querySelector(`.${WRAPPER_CLASSNAME}`)`
This way we get all the descendants, but we only want the direct child.
Just checked that, with the PayPal spinner already there, if you apply a coupon, this code will start that spinner, instead of creating a new one as direct child of form.llms-checkout.

```suggestion
	const wrapper = nodeList[0];
	let spinner =
		wrapper.querySelector(`.${WRAPPER_CLASSNAME}`) ||
		create(wrapper, size);
	// Make sure the found spinner is a direct child of the `wrapper`, otherwise create a new one.
	spinner = wrapper === spinner.parentNode ? spinner : create(wrapper, size);
```

What do you think about this solution?

packages/components/src/spinner/styles.js
```diff
@@ -0,0 +1,53 @@
...
+	.${WRAPPER_CLASSNAME} {
+		background: rgba( 250, 250, 250, 0.7 );
+		bottom: 0;
+		display: none;
+		left: 0;
+		position: absolute;
+		right: 0;
+		top: 0;
+	}
```
eri-trabiccolo: can we add a z-index here? Even just `2`. When applying a coupon the new PayPal button was not overlapped by the form spinner.
```suggestion
		top: 0;
		z-index: 2;
	}
```

eri-trabiccolo (changes requested):
Minor changes. Looks very good overall. <3

packages/components/src/spinner/styles.js
thomasplevy: what prefixes do we need here?

eri-trabiccolo: the old browsers compatibility, I was referring to the old "autoprefixer" npm:
https://www.npmjs.com/package/autoprefixer
postcss now, dunno which plugin takes care of that in webpack :D

thomasplevy: yea you're right, I didn't notice that.

but i don't think we actually need any of that anymore

box-sizing, keyframes, animation, etc... are all widely supported without prefixes.

eri-trabiccolo: I agree, I was more talking about this approach in general.

PR #2218, opened 2022-07-27
Don't override headings' font-weight site wide

### LINKED ISSUE #2217: Unwanted headings font-weight override?
https://github.com/gocodebox/lifterlms/blob/7932ff1e9ab160a66d52d01d838a0a99dec52e71/assets/scss/frontend/_main.scss#L74-L77

I'm not sure why we have that rule in place, but I think we should get rid of it.
Defining the headings font-weight sitewide, it's outside of our scope (and creates me problems with the new theme :P).

Of course changing this might have some drawback but to me it's something we should get rid of.

eri-trabiccolo:
## Description
Fixes #2217 

## How has this been tested?
manually

## Types of changes
Breaking change (fix or feature that would cause existing functionality to not work as expected)?

## Checklist:

### DIFF
--- a/.changelogs/dev.yml
+++ b/.changelogs/dev.yml
@@ -2,4 +2,4 @@ significance: minor
 type: changed
 links:
   - "#2217"
-entry: Don't override headings font-weight site-wide.
+entry: Removed site-wide font-weight styles targeting `<h1>` through `<h6> elements.
--- a/.changelogs/dev.yml
+++ b/.changelogs/dev.yml
@@ -1,0 +1,5 @@
+significance: minor
+type: changed
+links:
+  - "#2217"
+entry: Don't override headings font-weight site-wide.
--- a/assets/scss/frontend/_main.scss
+++ b/assets/scss/frontend/_main.scss
@@ -71,11 +71,6 @@
   text-align: center;
 }
 
-/* Genesis Overrides */
-h1, h2, h3, h4, h5, h6 {
-  font-weight: 300;
-}
-
 #main-content .llms-payment-options p {
   margin: 0;
   font-size: 16px;


thomasplevy:
@eri-trabiccolo did you happen to test this against a few "major" themes, astra, kadence, 2022, etc...?

I'm going to make an assumption that since the offending code has been in the stylesheet for *seven* years most themes aren't affected due to this rule being very general and they likely have even slightly more specific selectors that overrule it.

If any themes are affected by this, the custom css (into the customizer) to fix it is simple. I'm not worried, really, but I want to make sure the big recommended themes don't have problems before approving.

.changelogs/dev.yml
```diff
@@ -0,0 +1,5 @@
+significance: minor
+type: changed
+links:
+  - "#2217"
+entry: Don't override headings font-weight site-wide.
```
thomasplevy:
```suggestion
entry: Removed site-wide font-weight styles targeting `<h1>` through `<h6> elements.
```

thomasplevy (approved):
Assuming testing is "thorough" per my above comment I approve with this minor change to the changelog entry

eri-trabiccolo:
> @eri-trabiccolo did you happen to test this against a few "major" themes, astra, kadence, 2022, etc...?
> 
> I'm going to make an assumption that since the offending code has been in the stylesheet for _seven_ years most themes aren't affected due to this rule being very general and they likely have even slightly more specific selectors that overrule it.
> 
> If any themes are affected by this, the custom css (into the customizer) to fix it is simple. I'm not worried, really, but I want to make sure the big recommended themes don't have problems before approving.

Well, no, this is basically a breaking change for some of them..
I tested it with:
- 2022: headings in the post/page content were not bold before this PR, back to be bold with this PR. This is because the CSS defined in the theme.json of the FSE themes is printed inline earlier than plugin's (another inconsistency of the FSE?)
- 2021: no effect, bold before, bold after: safe change.
- 2020: no effect, bold before, bold after: safe change.
- Divi: no effect, bold before, bold after: safe change.
- Kadence: no effect, bold before, bold after: safe change.
- Astra: no effect, not bold before, not bold after: quite safe change?. Astra has this rule:
```
h1, h2, h3, h4, h5, h6 {
font-weight: inherit;
}
```
that we were overriding. Astra loads its main style before the plugin styles.
- OceanWP: no effect, bold before, bold after: safe change.

PR #2219, opened 2022-08-05
Enqueue styles

donaldafeith:
4 fonts for testing loaded
'no frills' just enqueue version

## Description

Fixes #

## Checklist:


thomasplevy (commented):
Initially I was in favor of adding only woff2 fonts, but I think we should include woff also, just to be safe for some older browsers.

Font support looks pretty solid to me across the board with woff and woff2.

We don't need to support IE at all anymore and the legacy iOS can absolutely be dropped.

@eri-trabiccolo do you have any thoughts on this?

lifterlms.php
```diff
@@ -48,6 +48,7 @@
 require_once LLMS_PLUGIN_DIR . 'vendor/autoload.php';
 require_once LLMS_PLUGIN_DIR . 'includes/class-llms-loader.php';
 
+
```
thomasplevy, line 51:
```suggestion
```

package.json
```diff
@@ -18,7 +18,8 @@
   },
   "dependencies": {
     "@emotion/styled": "^11.6.0",
-    "classnames": "^2.3.1"
+    "classnames": "^2.3.1",
+    "registry.npmjs.org": "^1.0.1"
```
thomasplevy, line 22:
```suggestion
```
Looks like this may have been an accidental commit unrelated to this changeset.

Can you drop this (and the package-lock.json) changes please

includes/functions/llms.functions.certificate.php
```diff
@@ -486,4 +486,13 @@ function llms_register_certificate_image_size() {
...
 }
-add_action( 'after_setup_theme', 'llms_register_certificate_image_size' );
+/**
+ * Register and enqueue a local font cache of Google Fonts
+ */
+function llms_local_fonts() {
+	$plugin_url = ( plugins_url( '/lifterlms/', $file ) );
+
+wp_enqueue_style( 'style',  $plugin_url . "assets/css/localstyles.css");
+}
```
thomasplevy:
```suggestion
```

We should drop this... I think we can avoid enqueuing directly through the WP apis for now (knowing that a true webfont API is coming in a future WP core release).

If we drop this we can load them through the existing loader in the dynamic-styles.php (with a few other changes noted there)

However, the `llms_get_certificate_fonts()` function in this file needs to be adjusted to include the CSS file for each font independently.

I'd like to *keep* the google preloading as a fallback through a filter so I'm thinking that if you add a filter to the top of the function, something like:

```php
$use_g_fonts = apply_filters( 'llms_use_google_webfonts', false );
```

Then each `href` prop of the fonts array can be switched to a ternary, eg:

`'href' => $use_g_fonts ? {google-url} : LLMS_PLUGIN . {localurl}`

templates/certificates/dynamic-styles.php
```diff
@@ -23,15 +23,15 @@
 $gfonts_preconnet = false;
 ?>
 
-<!-- Certificates Dynamic Styles -->
+<!-- Certificates  -->
```
thomasplevy:
```suggestion
<!-- Certificates Dynamic Styles -->
```

No need to change the comment?

templates/certificates/dynamic-styles.php
```diff
@@ -23,15 +23,15 @@
...
 	<?php if ( ! empty( $font['href'] ) ) : ?>
 		<?php
-		if ( ! $gfonts_preconnet && false !== strpos( $font['href'], 'fonts.googleapis.com' ) ) :
+		if ( ! $gfonts_preconnet && false !== strpos( $font['href'], LLMS_PLUGIN_URL . 'assets/css/localstyles.css'  ) ) :
 			$gfonts_preconnet = true;
 			?>
-			<link rel="preconnect" href="https://fonts.googleapis.com">
-			<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+			<link rel="preconnect" href="<?LLMS_PLUGIN_URL ?>">
+			<link rel="preconnect" href="<?LLMS_PLUGIN_URL ?>" crossorigin>
```
thomasplevy: With the other changes I've suggested for loading methodology we should keep this to preconnect to google. That way if a use opts in to loading from google directly (via a filter) the preconnect will still happen.

assets/css/localstyles.css
```diff
@@ -0,0 +1,52 @@
+/* PirataOne/pirata-one-regular - latin */
```
thomasplevy, line 1: I think this file can be dropped entirely in favor of loading each font indepedently.

This isn't the *ideal* method since it adds additional unnecessary HTTP connections but the way the forthcoming webfonts api works means this will problem that will go away once we switch APIs (pending core changes of course).

assets/css/fonts/DancingScript/dancing-script.css
```diff
@@ -0,0 +1,12 @@
+@font-face {
+    font-family: 'Dancing Script';
+    font-style: normal;
+    font-weight: 400;
+    src: url('./fonts/dancing-script-v24-latin-regular.eot'); /* IE9 Compat Modes */
+    src: local('Dancing Script'),
```
thomasplevy: This is a cool function, I wasn't aware of it's existence

assets/css/fonts/DancingScript/dancing-script.css
```diff
@@ -0,0 +1,12 @@
...
+         url('./fonts/dancing-script-v24-latin-regular.eot?#iefix') format('embedded-opentype'), /* IE6-IE8 */
+         url('./fonts/dancing-script-v24-latin-regular.woff2') format('woff2'), /* Super Modern Browsers */
+         url('fonts/dancing-script-v24-latin-regular.woff') format('woff'), /* Modern Browsers */
+         url('fonts/dancing-script-v24-latin-regular.ttf') format('truetype'), /* Safari, Android, iOS */
+         url('fonts/dancing-script-v24-latin-regular.svg#DancingScript') format('svg'); /* Legacy iOS */
```
thomasplevy: I think the paths are messed up here

thomasplevy (changes requested):

thomasplevy:
@donaldafeith Welcome to the team! I think this is a great first pull into LifterLMS but it's not ready yet. I've made some "suggestions". Please have a look through and let me know if you have any questions!

thomasplevy (commented):
another note I just noticed is the fonts are all added inside `assets/css`, I'd rather see these in the existing `assets/fonts` directory and then the stylesheets can moved into the root of the `assets/css` directory.

lifterlms.php
thomasplevy: This empty line (resulting in an irrelevant additional file in the changeset) is still present

includes/admin/class.llms.admin.assets.php
```diff
@@ -34,6 +34,7 @@ public function __construct() {
 		add_action( 'admin_print_scripts', array( $this, 'admin_print_scripts' ) );
 		add_action( 'admin_print_footer_scripts', array( $this, 'admin_print_footer_scripts' ) );
 		add_action( 'enqueue_block_editor_assets', array( $this, 'block_editor_assets' ) );
+		add_action( 'wp_enqueue_scripts', 'local_styles' ); //Add action for local styles
```
thomasplevy, line 37:
```suggestion
```

includes/admin/class.llms.admin.assets.php
```diff
@@ -501,7 +502,17 @@ public static function register_quill( $modules = array() ) {
...
+	//  * Initialize Local Styles
+	//  *
+	//  * @since Unknown.
+	//  *
+	//  * @return void
+	//  */
+	// 	public function local_styles() {
+        
+    //     wp_enqueue_style( plugin_dir_url . 'assets/css/localstyles.css' );
+	// }
```
thomasplevy, line 515:
```suggestion
```

templates/certificates/dynamic-styles.php
thomasplevy: This is still outstanding

assets/fonts/UnifrakturMaguntia/UnifrakturMaguntia.css
```diff
@@ -0,0 +1,13 @@
+/* unifrakturmaguntia-regular - latin */
```
thomasplevy, line 1: the changes made to the other fonts (use only woff & woff2) need to be made to this font as well

includes/admin/class.llms.admin.assets.php
thomasplevy, line 37:
```suggestion
```

This will cause an error

includes/admin/class.llms.admin.assets.php
thomasplevy, line 515:
```suggestion
```

No commented out good without good reason <3

templates/certificates/dynamic-styles.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /**
- * Single certificate dynamic styles.
+ * Single certificate .
```
thomasplevy, line 3:
```suggestion
 * Single certificate dynamic styles.
```

I don't see any reason to change the comment here.

eri-trabiccolo:
> Initially I was in favor of adding only woff2 fonts, but I think we should include woff also, just to be safe for some older browsers.
> 
> Font support looks pretty solid to me across the board with woff and woff2.
> 
> We don't need to support IE at all anymore and the legacy iOS can absolutely be dropped.
> 
> @eri-trabiccolo do you have any thoughts on this?

Ok for woff too, it'll be loaded as fallback only so ok.

includes/functions/llms.functions.certificate.php
```diff
@@ -110,6 +110,66 @@ function llms_get_certificate_fonts() {
 
 	$serif = '"Iowan Old Style", "Apple Garamond", Baskerville, "Times New Roman", "Droid Serif", Times, "Source Serif Pro", serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol"';
 
+	$fonts = array(
```
thomasplevy, line 113: this change doesn't really do what we'd like it to do. It does provide devs with a way to opt back into using google fonts, kind of, but not really in the way that I envisioned it (it results in a more convoluted switch rather than using a `__return_true` callback on `add_filter()`

Please refer back to my original suggestion on using a ternary on the `href` property of each font.

Let me know if you have any questions on my desired implementation

templates/certificates/dynamic-styles.php
```diff
@@ -27,11 +27,11 @@
...
 	<?php if ( ! empty( $font['href'] ) ) : ?>
 		<?php
-		if ( ! $gfonts_preconnet && false !== strpos( $font['href'], 'fonts.googleapis.com' ) ) :
+		if ( ! $gfonts_preconnet && false !== strpos( $font['href'], LLMS_PLUGIN_URL . '/assets/css/localstyles.css'  ) ) :
 			$gfonts_preconnet = true;
 			?>
-			<link rel="preconnect" href="https://fonts.googleapis.com">
-			<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+			<link rel="preconnect" href="<?LLMS_PLUGIN_URL ?>">
+			<link rel="preconnect" href="<?LLMS_PLUGIN_URL ?>" crossorigin>
```
thomasplevy, line 34: This is problematic still.

First, we don't need to preconnect to the site as we're already connected and second this is designed to preconnect to google when google fonts are being used so the change actually should be undone.

Also, if you got the `<link>` elements to output they'd have an empty `href` since the constant isn't being echoed.

Finally, short open tags are generally frowned upon in WordPress since not all server environment support them.

thomasplevy (changes requested):

Issue #2189, opened 2022-06-21
Include google webfonts directly in favor of loading remotely


thomasplevy:
In light of this recommendation change (for themes) we should probably get out in front of this and load certificate fonts locally in favor of loading them from Google directly. 

https://make.wordpress.org/themes/2022/06/18/complying-with-gdpr-when-using-google-fonts/

Source: https://github.com/gocodebox/lifterlms/blob/f40e79dcb75486e0581d19e0d46c462e1801b3f1/includes/functions/llms.functions.certificate.php#L96-L164

I think that a safe bet is to just switch to loading them locally. Though by doing so we'll create a "performance issue" for anyone who considers loading the fonts from the local site to be detrimental to page load speeds.

Perhaps a filter flag could be used to switch back to the google-hosted URL for anyone who might wish to continue loading remotely.

donaldafeith:
The best way I can find to do this would be to have the fonts downloaded in to a folder /fonts/ or some other way and then called to.
Example:
``
 		'imperial-script'     => array( 
 			'name'       => 'Imperial Script', 
 			'href'       => 'https://the-website-using-them.com/css2?family=Imperial+Script&display=swap', 
 			'fontFamily' => 'Imperial Script', ' . $serif, 
 		), 
``
Or (Not tested just brainstorming but calling using the get_site_url() to the plugin and then /fonts/and the font css family there.)
``
 		'imperial-script'     => array( 
 			'name'       => 'Imperial Script', 
 			'href'       =>(get_site_url())'/fonts/css2?family=Imperial+Script&display=swap', 
 			'fontFamily' => 'Imperial Script', ' . $serif, 
 		), 
``

I know we could do an @import but it would still call to Google. If they totally don't want that anymore the best way I figure is to include the fonts (I will check to see if that is allowed... I believe it is.) and then package them as well.

Any thoughts on this? I'm going to do some different ways of doing it and then we can pick the best way in terms of speed, bandwidth usage, and user loading... That's the only thing I worry about is user browser loading. Google is fast and usually very good about maintaining their codebases (but not their documentation... :D ) But I could write up a fallback (if WordPress is okay with that) in that we could have it... In theory if it was taking too long to load (page hosts vary...) it could then fallback to Google to import/download/etc.

thomasplevy:
Hey @donaldafeith 

There's this: https://github.com/WPTT/webfont-loader

But... I'm not sure using that is necessary for our project here.

[...]

PR #2220, opened 2022-08-08
Df #2189

donaldafeith:
## Description
Added the ability to use local fonts for the certificates, and also the ability to shut off local and turn on google fonts by using a true/false switch.

Fixes # Using local fonts instead of using the google fonts.

## How has this been tested?
Tested on local LAMP server viewing a certificate. Changing the styles and making sure Google Fonts were not being called.

Called font, changed, style, called another font

Affects certificate

## Types of changes
Local font support ahead of the WordPress font API. 

Functionality would not work so well if the theme used was using Google fonts already or using their own local cache. It may still call to the local fonts when viewing the certificate, but it would depend on how the theme was laid out.

## Checklist:
- [X ] My code has been tested.


thomasplevy:
Closing in favor of preexisting #2220

PR #2221, opened 2022-08-10
@lifterlms/scripts update for easier compilation of blocks

thomasplevy:
## Description

This update to [`@lifterlms/scripts`](https://github.com/gocodebox/lifterlms/tree/trunk/packages/scripts) introduces a new webpack config file aimed at building blocks. The intended project structure is outlined in updates to the README.md included this PR.

This PR additionally updates the LifterLMS core plugin's webpack config to utilize the new config.

There is a change to the preexisting webpack config configuration generator which removes a copy routine that copies block.json files, this is redundant when using the new config. This is a breaking change.

Upgrades the `@wordpress/scripts` dependency to the latest. This pulls upstream breaking changes into this package.

~Finally, this PR introduces a (potentially controversial) change: removal of the blocks build/dist directory from the git repository. The logic here is that in general repositories for software projects do not include dependencies and distributed files *except* with release tags / branches. Since the LifterLMS repository is a source repository it is unnecessary to include distributed / compiled static assets. This is a change I wish to make generally across the codebase (and other plugins) and I'm starting here. We can fight about this if you'd like, @eri-trabiccolo~

## How has this been tested?

+ Manually tested the blocks build process (and the existing webpack config) with these changes.
+ Manually tested the new config on the new aircraft project (which currently has 3 completed blocks).

## Types of changes
+ New features
+ Breaking changes

## Checklist:


eri-trabiccolo (approved):
Seems to be good but, correct me if I'm wrong, we still need to specify the js files to be removed for non blocks.

thomasplevy:
> Seems to be good but, correct me if I'm wrong, we still need to specify the js files to be removed for non blocks.

What do you mean @eri-trabiccolo ?

eri-trabiccolo:
> > Seems to be good but, correct me if I'm wrong, we still need to specify the js files to be removed for non blocks.
> 
> What do you mean @eri-trabiccolo ?

I mean:
In the new theme we're dropping gulp, so webpack will generate the CSS too. To do that we have js files in src/js that just load the scss files in src/scss, then in the webpack config I take care of removing the useless js files files created in `assets/`. I will still have to do that, nothing will change on that respect. Right?

thomasplevy:
> In the new theme we're dropping gulp, so webpack will generate the CSS too. To do that we have js files in src/js that just load the scss files in src/scss, then in the webpack config I take care of removing the useless js files files created in `assets/`. I will still have to do that, nothing will change on that respect. Right?

You *will* not have to do that anymore, actually.

To generate CSS without an associated JS file you will do the following:

```
const { resolve } = require( 'path' ),
	generate = require( '@lifterlms/scripts/config/webpack.config' ),
	config = generate( {} );

config.entry.styles = resolve( './src/scss/styles.scss' );

module.exports = config;
```

This *should* work at least, I haven't tested it exactly.

this would create two files: `assets/css/styles.css` and `assets/css/styles-rtl.css`

eri-trabiccolo:
> > In the new theme we're dropping gulp, so webpack will generate the CSS too. To do that we have js files in src/js that just load the scss files in src/scss, then in the webpack config I take care of removing the useless js files files created in `assets/`. I will still have to do that, nothing will change on that respect. Right?
> 
> You _will_ not have to do that anymore, actually.
> 
> To generate CSS without an associated JS file you will do the following:
> 
> ```
> const { resolve } = require( 'path' ),
> 	generate = require( '@lifterlms/scripts/config/webpack.config' ),
> 	config = generate( {} );
> 
> config.entry.styles = resolve( './src/scss/styles.scss' );
> 
> module.exports = config;
> ```
> 
> This _should_ work at least, I haven't tested it exactly.
> 
> this would create two files: `assets/css/styles.css` and `assets/css/styles-rtl.css`

Ok great, gonna test it tomorrow and tell you <3

thomasplevy:
@eri-trabiccolo I'm going to merge (and ship) this. If you have any issues with the above we can update the package accordingly. I think it should work but we may need to do a revision to make it work as I'm expecting it to.

eri-trabiccolo:
Yeah go ahead but with love

PR #2222, opened 2022-08-11
Font Awesome

thomasplevy:
## Description

Adds a Font Awesome wrapper / package

Requires #2221

## How has this been tested?

+ Manually

## Types of changes
New package

## Checklist:


includes/assets/llms-assets-styles.php
```diff
@@ -31,6 +31,7 @@
  * @since 4.8.0 Added llms-admin-importer.
  * @since 5.0.0 Added llms-select2.
  * @since 5.5.0 Added llms-admin-add-ons.
+ * @since [version] Added llms-fontawesome.
```
eri-trabiccolo: mmm where?

webpack.config.js
```diff
@@ -42,6 +42,8 @@ config.plugins.push( new CleanWebpackPlugin( {
 
 } ) );
 
+// config.entry.fontawesome = resolve( './src/scss/fontawesome.scss' );
```
eri-trabiccolo, line 45: shall we uncomment this?

eri-trabiccolo (changes requested):
this basically only adds the package?

thomasplevy:
> this basically only adds the package?

Yes. At some point we'll update the LifterLMS core and drop the existing FA usage (4.x) and upgrade to using this. But that needs to happen over time. Many add-ons use FA and it's possible 3rd parties / themes might be relying on them from us too (possibly). We'll need to add some sort of transition the existing FA shims or whatever... Anyway... 

This PR introduces the package but is not meant to actually introduce the code into LifterLMS itself. We'll be using this in Aircraft first and later in the core...

I had committed code I intended to remove (I left some of it commented out intentionally)

webpack.config.js
thomasplevy: No, since the LifterLMS core won't be using FA 6 (yet) we can leave it out. We can uncomment it later when we decide to transition from FA 4 -> 6 in the core. For now I'll just leave it commented out.

includes/assets/llms-assets-styles.php
thomasplevy: I MADE A MISTAKE!

thomasplevy:
```suggestion
```

webpack.config.js
eri-trabiccolo: ok fine

PR #2242, opened 2022-08-24
added docs to the 'llms_certificate_merge_data' filter'

eri-trabiccolo:
docs


includes/models/model.llms.user.certificate.php
```diff
@@ -491,6 +491,22 @@ protected function get_merge_data() {
 			'llms_certificate_merge_data'
 		);
 
+		/**
+		 * Filters the certificate merge data.
+		 *
+		 * @since 6.0.0
+		 *
+		 * @param array $codes      {
+		 *
```
thomasplevy:
```suggestion
```

PR #2266, opened 2022-09-15
Issue/2189

### LINKED ISSUE #2189: Include google webfonts directly in favor of loading remotely

nrherron92:
## Description
Adds local call to fonts avaliable in the certificate builder. 

Fixes #2189 

## How has this been tested?
Ran 'composer run check-cs' and manually checked the fonts worked and were loading from the local link in dev tools.

## Types of changes
Bug fix?

## Checklist:


thomasplevy (changes requested):
the RTL css files can be dropped since they're identical to the regular versions (and the loader doesn't check RTL so it'll never load them anyway)

nrherron92:
on it

thomasplevy (changes requested):
I think we should drop the version from the `.woff` filenames.

If we update the fonts later we'll have to remove a file and replace it (in favor of just updating the file itself).

Deleting a file is something we don't take lightly so it would have to be saved for a major release to remove.

It'll be easier to update in the future if the version number isn't in the file name.

For cache-busting and version control purposes a query string variable can be added to the URL in the `llms_get_certificate_fonts()` function.

For example: `LLMS_PLUGIN_URL . 'assets/css/unifraktur-maguntia.css'` becomes: `LLMS_PLUGIN_URL . 'assets/css/unifraktur-maguntia.css?ver=v19'`

Then if we update the font to version 20, we update the CSS and woff files and then bump the version number in the query string variable without having to remove any files, this can be done in a minor release without worry for backwards compatibility stuff.

includes/functions/llms.functions.certificate.php
```diff
@@ -107,7 +107,7 @@ function llms_get_certificate_content( $id = 0 ) {
  * }
  */
 function llms_get_certificate_fonts() {
-
+	$use_g_fonts = apply_filters( 'llms_use_google_webfonts', false );
```
thomasplevy, line 118: Needs a filter doc:


```suggestion
	/**
	 * Determines whether or not webfonts are loaded from Google CDNs.
	 *
	 * @since [version]
	 *
	 * @param bool $use_g_fonts If `true`, fonts are loaded from Google, otherwise they are loaded from the local site.
	 */
	$use_g_fonts = apply_filters( 'llms_use_google_webfonts', false );
```

includes/functions/llms.functions.certificate.php
```diff
@@ -107,7 +107,7 @@ function llms_get_certificate_content( $id = 0 ) {
  * }
  */
 function llms_get_certificate_fonts() {
```
thomasplevy, line 110: function needs a new `@since [version] description...` changelog added to it

and the file's `@version` tag should be replaced with the `[version]` placeholder

thomasplevy (changes requested):
overall looks very good, a few minor pieces of feedback regarding versioning and RTL files (see above)

A few doc suggestions.

Overall the code looks sound but I haven't tested it yet.

Also, needs a `.changelog/` file to include with the PR

includes/functions/llms.functions.certificate.php
nrherron92: Added!

thomasplevy (changes requested):
@nrherron92 I think you accidentally deleted all the fonts, the whole `assets/fonts/` dir is missing from the PR now!

nrherron92:
alright I didn't force push this time lets see if I managed to stop deleting things now 🤦🏼‍♀️

thomasplevy (changes requested):
ahhhh @nrherron92 you're so close!

The woff2 file for UnifrakturMaguntia is missing a hyphen in the name!

PR #2282, opened 2022-10-03
Full Site Editing: wrap custom template's header and footer into proper tags

### LINKED ISSUE #2281: FSE: header and footer in our custom templates don't have proper tags:
When Header and Footer templates are included, they result wrapped into a standard div:
`<div class="wp-block-template-part">`

While they should be more properly been included into `header` and `footer` tags.
`<header class="wp-block-template-part">`
`<footer class="wp-block-template-part">`

This is of course a general rule, different themes might require different treatments.

**Note:**
FSE it's a bit tricky on this.
`get_header()` (`get_footer()`) ensured we always used the theme's header and footer as a whole , but with FSE, header and footer templates only define their inner content, which, when included, is always wrapped into another element (`<div class="wp-block-template-part">` by default). You can set the wrapper tag and classes when you include them, e.g.:
`<!-- wp:template-part {"slug":"header","tagName":"header","className":"site-header"} /-->`

This is a quite big difference because it means that themes are required to define the header partly in the header template part and partly in the template that includes it. This means that plugins' custom templates will basically never show header and footer that are exactly like the other pages.

eri-trabiccolo:
## Description
Fixes #2281 

## How has this been tested?
manually

## Types of changes
Enhancement

## Checklist:

### DIFF
--- a/.changelogs/issue_2281.yml
+++ b/.changelogs/issue_2281.yml
@@ -1,6 +1,5 @@
-significance: minor
+significance: major
 type: changed
 links:
   - "#2281"
-entry: "Full Site Editing: wrap custom template's header and footer into proper
-  tags."
+entry: "Full Site Editing: **[BREAKING]** The wrappers in the custom header and footer templates have been changed to the semantic HTML tags `<header>` and `<footer>` in favor of default `<div>` tags."
--- a/.changelogs/issue_2281.yml
+++ b/.changelogs/issue_2281.yml
@@ -1,0 +1,6 @@
+significance: minor
+type: changed
+links:
+  - "#2281"
+entry: "Full Site Editing: wrap custom template's header and footer into proper
+  tags."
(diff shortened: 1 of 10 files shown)


thomasplevy:
@eri-trabiccolo Does changing the default tag from `div` to the semantic equivalents create a backwards incompatible change?

I think we're probably safe here, noting as such, since FSE is still in "beta" in the core...

Thoughts?

eri-trabiccolo:
Yeah I think that as long as FSE is in beta we're safe.

.changelogs/issue_2281.yml
```diff
@@ -0,0 +1,6 @@
+significance: minor
```
thomasplevy:
```suggestion
significance: major
```

.changelogs/issue_2281.yml
```diff
@@ -0,0 +1,6 @@
...
+type: changed
+links:
+  - "#2281"
+entry: "Full Site Editing: wrap custom template's header and footer into proper
+  tags."
```
thomasplevy:
```suggestion
entry: "Full Site Editing: **[BREAKING]** The wrappers in the custom header and footer templates have been changed to the semantic HTML tags `<header>` and `<footer>` in favor of default `<div>` tags."
```

.changelogs/issue_2281.yml
thomasplevy: per our slack conversation let's include this in v7 and call it a major change

eri-trabiccolo: <3

PR #2290, opened 2022-10-28
Introduce "enum-link" pattern and allow customization of the instructor admin menu hack

thomasplevy:
## Description

A dev asked about the instructor admin menu hack -- still an open trac bug: https://core.trac.wordpress.org/ticket/16808

This dev is adding the ability for an instructor to edit posts, our "hack" doesn't differentiate between explicitly granting the `edit_posts` capability, and having the capability for the purposes of this hack.

I looked into removing it or implementing a different solution. And found adding a filter probably the simplest way to achieve the dev's needs. The `instructor` role can be removed via filter and then the hack won't affect them anymore and they'll be able to edit posts due to being granted the additional permissions explicitly.

Additionally, I've been working on this enum concept for a bit and felt this was a simple place to introduce it for review and inclusion.

Neither of these are high priority and I'm open to feedback on a better implementation.

I included the capabilities "enum" because I initially planned to add a new capability that would be used in favor of `edit_posts`. If they had this cap, we'd "pseudo add" `edit_posts` conditionally via the `user_has_cap` filter... But this got complicated quick and I started to question the removal of the `edit_posts` cap. It would require a database migration and possibly could break existing customizations. But I'd already written the code so I included it.

## How has this been tested?

+ Manually
+ New & existing unit tests

## Types of changes
New feature (non-breaking change which adds functionality

## Checklist:

### DIFF
--- a/.changelogs/dev-1.yml
+++ b/.changelogs/dev-1.yml
@@ -1,0 +1,4 @@
+significance: minor
+type: deprecated
+entry: Deprecated public constant `LLMS_Roles::MANAGE_EARNED_ENGAGEMENT_CAP` in
+  favor of `LLMS_Capabilities::MANAGE_EARNED_ENGAGEMENT`.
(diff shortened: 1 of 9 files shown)


thomasplevy:
ooof sometimes I'm silly.

PR #2294, opened 2022-10-31
Fix export file names with special characters

### LINKED ISSUE #1540: Bug in export students report
### Reproduction Steps

+ Install Lifterlms
+ Switch the WordPress language to non latin language like Arabic as me
+ Make sure the lifterlms translation is pulled
+ Open the Report section in the Lifterlms dashboard and try to export the Students report

### Expected Behavior

+ Should work and install the csv file as expected

### Actual Behavior

+ The file download fail and show server problem
Screenshot from 2021-02-22 10-44-14

### Bug Info
After deep looking i figured the reason of this bug, the generated file name has a URL encoded characters because it uses the title of the table we export, i mean here 'students' but because we pulled the arabic translations the translation text of 'students' will be used, see the image below:
Screenshot from 2021-02-22 10-33-59

and then when we request the file in the browser the server will evaluate the URL encoded characters so it will not match the file name so i need to URL encode the word 'الطلاب' (students) twice then use it in the request to get it work

maybe it's better to not use the title of the table at all and just use the id of the table

https://github.com/gocodebox/lifterlms/blob/trunk/includes/abstracts/llms.abstract.exportable.admin.table.php#L75

**Note:** other report section like courses is working as expected because it's not using the table title so the file name is always begin with 'llms-courses'

### Error Messages / Logs

```

```

<details>

```


```

</details>

This issue has be recreated:
+ [x] Locally
+ [x] On a staging site
+ [x] On a production website

+ Browser name and version

thomasplevy:
## Description

Fixes #1540 

## How has this been tested?

+ New unit tests
+ Manually

## Types of changes
Bug fix

## Checklist:

### DIFF
--- a/.changelogs/export-special-chars.yml
+++ b/.changelogs/export-special-chars.yml
@@ -1,0 +1,6 @@
+significance: patch
+type: fixed
+links:
+  - "#1540"
+entry: Use `sanitize_file_name()` in favor of `sanitize_title()` for generating
+  the file name of reporting table export files.
--- a/.changelogs/export-special-chars-1.yml
+++ b/.changelogs/export-special-chars-1.yml
@@ -1,0 +1,6 @@
+significance: patch
+type: dev
+entry: A stub method, `get_title()` has been added to the
+  `LLMS_Abstract_Exportable_Admin_Table` abstract class. This method should be
+  defined by any extending classes and will throw a `_doing_it_wrong()` error
+  when undefined.
--- a/includes/abstracts/llms.abstract.exportable.admin.table.php
+++ b/includes/abstracts/llms.abstract.exportable.admin.table.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Abstracts/Classes
  *
  * @since 3.28.0
- * @version 4.0.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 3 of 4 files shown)


.changelogs/export-special-chars.yml
```diff
@@ -0,0 +1,6 @@
+significance: patch
```
eri-trabiccolo, line 1: We're introducing an abstract method here, which means that all the classes that extend `LLMS_Abstract_Exportable_Admin_Table` must now implement that otherwise we have a fatal. This is a breaking change, isn't it?

eri-trabiccolo (changes requested):
Except for the significance, it looks fine.

.changelogs/export-special-chars.yml
thomasplevy: well... I added it because it's a bug that it didn't already exist since other methods in the class reference the method already.

It's possible that other classes extending this didn't happen to use the method(s) that reference the missing abstract method.

You're right though. it's a breaking change this way.

If we added it as a stub with an empty string return it would fix the bug I'm seeing without making any breaking changes for other classes, right?

eri-trabiccolo: yeah with a warning yeah

thomasplevy: what warning? like a warninig in the function?

eri-trabiccolo: like a doing it wrong. E.g.
https://github.com/gocodebox/lifterlms-rest/blob/trunk/includes/abstracts/class-llms-rest-controller-stubs.php#L90

thomasplevy: @eri-trabiccolo updated with `_doing_it_wrong()` per your suggestion

eri-trabiccolo (approved):

PR #2297, opened 2022-11-09
Introduce something like a database ORM

thomasplevy:
## Description

This PR introduces a new set of classes aimed to reduce repetition in database table creation and management. It also includes schema files for each existing database table in the LifterLMS core plugin.

This PR introduces functionality for registering and managing tables.

The next steps would be to utilize the `LLMS_Database_Table` class within object models to setup getters and setters based on the already-defined schema... Also query classes could utilize this data to automatically generate query classes based off the data available in the table's schema.

This is likely not completely ready for production yet. I wanted to open the PR to review and improve.

This PR also includes a simple PHP CLI script for automatically generating markdown docs describing database tables. I'd like to package that script up a bit differently so that add-ons could also easily utilize it to generate documentation. I'm considering bundling it into an WP (LLMS) CLI command so that an add-on (which already depends on LifterLMS) can easily generate it's own documentation.

My plan for this markdown file would be to (eventually) pull it into developer.lifterlms.com via the incomplete "handbook" importer: https://github.com/gocodebox/developer.lifterlms.com/pull/29

Other add-ons that have their own markdown database docs could be added the the handbook(s) as well later on.

This was inspired, pretty heavily, by BerlinDB. I initially wished to use BerlinDB directly but I had a few problems with it and ultimately decided that we could manage this internally in a way that's better for us without having to worry about another plugin's requirements causing issues for us.

## How has this been tested?
+ Manually
+ New and existing tests

## Types of changes
+ New feature / enhancement 

## Checklist:


includes/class-llms-database-table.php
```diff
@@ -0,0 +1,299 @@
...
+	 * @var array
+	 */
+	protected array $schema;
+
+	/**
+	 * References the global `$wpdb` instance.
+	 *
+	 * @var wpdb
+	 */
+	protected $wpdb;
```
eri-trabiccolo, line 48: I don't know whether this makes all this sense here, does it?
We're referencing to a global in a protected property for which we would have to ensure bw compatibility?!
Let's use the global where needed.

thomasplevy: This makes testing "easier"

When we want to mock a particular portion of the global we can simply overwrite the class property in favor of mocking the whole global.

EG:

```
$mock_wpdb = new class {
  function has_cap( $cap ) {
     return false;
  }
}
$table = new LLMS_Database_Table( ... );
LLMS_Unit_Test_Util::set_private_property_value( $table, 'wpdb', $mock_wpdb );
```

This feels like a better approach than actually physically mocking the global. We could of course mock the global too and maybe that's better but I don't like that as much.

As far as backwards compatibility is concerned, if we drop it we drop it. But... I don't forsee `wpdb` ever being dropped from WordPress so we'll probably never need to drop it either. And if we decide we don't need it anymore it's not hard to deprecate.

Maybe my thinking is shortsighted though?

eri-trabiccolo: I don't know, I feel it's not a good practice overdoing something for "more easy testing" purposes only. But ok it's fine.

includes/class-llms-database-table.php
```diff
@@ -0,0 +1,299 @@
...
+	 *                  and explicitly return default values which may be left
+	 *                  undefined on the raw schema.
+	 * @return array
+	 */
+	public function get_schema( bool $raw = true ): array {
+
+		if ( $raw ) {
+			return $this->schema;
+		}
+
```
eri-trabiccolo, line 207: shall we cache the computed schema?

includes/schemas/database/quiz-attempts.php
```diff
@@ -0,0 +1,75 @@
+<?php
+/**
+ * Table Schema Definition: lifterlms_user_postmeta
```
eri-trabiccolo, line 3:
```suggestion
 * Table Schema Definition: lifterlms_quiz_attempts
```

includes/schemas/database/sessions.php
```diff
@@ -0,0 +1,43 @@
...
+			'allow_null'  => false,
+		),
+		'data'        => array(
+			'description' => 'The session data, stored as a JSON object.',
+			'type'        => 'longtext',
+			'allow_null'  => false,
+		),
+		'expires'     => array(
+			'description' => 'The session data, stored as a JSON object.',
+			'type'        => 'id',
```
eri-trabiccolo, line 36: this is a UNIX timestamp actually...

includes/schemas/database/sessions.php
```diff
@@ -0,0 +1,43 @@
...
+			'type'        => 'primary_id',
+		),
+		'session_key' => array(
+			'description' => 'The unique session key.',
+			'type'        => 'char',
+			'length'      => 32,
+			'allow_null'  => false,
+		),
+		'data'        => array(
+			'description' => 'The session data, stored as a JSON object.',
```
eri-trabiccolo, line 30: isn't actually a serialized array?

eri-trabiccolo (changes requested):
Looks a good starting point

eri-trabiccolo:
we're using php8 code here, let's not forget to raise the min required PHP version.

Issue #940, opened 2019-09-19
Delete User data when a user is deleted


thomasplevy:
Migrated from #263 

- [ ] Unschedule recurring payments, delete orders
- [ ] Delete data from lifterlms_user_postmeta table
- [ ] Delete quiz attempts
- [ ] Delete events from the lifterlms_events table
- [ ] Delete notifications from the lifterlms_notifications table

eri-trabiccolo:
I'm possibly going to break this in several PR so that the review process would be easier.

thomasplevy:
Sounds good

PR #2315, opened 2022-12-30
llms_is_user_enrolled always return false for non existing or unlogged users

eri-trabiccolo:
## Description
While looking at https://github.com/gocodebox/private-issues/issues/63
I realized that we "try" to query the DB looking for enrollments even for not logged in users, and generally for non existing users (e.g. deleted users). In this PR I just return immediately false when checking for enrollments of non existing users when using `llms_is_user_enrolled()`. 
This is technically a breaking change (so will ideally push the next LifterLMS version to 8.0), as we're changing the default behavior of a public function although we internally always used that assuming that the user was existing: with that function you could theoretically query the enrollment status of a student in a course even if the student was deleted.
Also consider that in the future we  want to delete the enrollment related data of a user upon its removal (see https://github.com/gocodebox/lifterlms/issues/940 bullet  _Delete data from lifterlms_user_postmeta table_). Maybe we want take the occasion and implement a fix for #940 so to release a new major version including this PR?

This PR will also basically fix https://github.com/gocodebox/private-issues/issues/63

Also this PR has some perfomance impact (improvements) on not logged in users visiting e.g. courses or membership catalogs: for each course/membership in the current catalog page we check whether or not the current user (user with ID 0 for non logged in users) is enrolled in the course/membership performing a custom and not simple DB query (although on an index of the table).

## How has this been tested?
manually and existing unit tests

## Types of changes
Breaking change (fix or feature that would cause existing functionality to not work as expected) -->

## Checklist:

### DIFF
--- a/.changelogs/check_user_existence_on_checking_enrollment.yml
+++ b/.changelogs/check_user_existence_on_checking_enrollment.yml
@@ -1,0 +1,5 @@
+significance: minor
+type: dev
+entry: The function `llms_is_user_enrolled()` will always return `false` for non
+  existing users. While, before, it could return `true` if a now removed user
+  was enrolled into a the given course or membership.
--- a/.changelogs/enrollment-no-user.yml
+++ b/.changelogs/enrollment-no-user.yml
@@ -1,0 +1,5 @@
+significance: major
+type: performance
+entry: Immediately return false when running `llms_is_user_enrolled()` on logged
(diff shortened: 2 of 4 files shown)


eri-trabiccolo:
@ideadude 
changelog entries done and rebased to the current dev: ready to be merged.

PR #2320, opened 2023-01-05
Correctly record manually placed trial transactions

### LINKED ISSUE #2308: Manually recorded first transaction on an access plan with unpaid trial, is not recorded as trial transaction
### Reproduction Steps

+ Create an access plan with trial
+ as visitor, purchase the access plan with manual payment
+ as admin head on the order and manually record a transaction, ideally for the trial.

### Expected Behavior

+ I'd expect the transaction is recorded for the trial

### Actual Behavior

+ The transaction is actually recorded as a "normal" transaction, this also means that it if you have a plan with a trial and a recurring plan with 5 payments due, this transaction will decrease the number of remaining recurring transactions from 5 to 4.

LifterLMS 7.0.1

+ [x] Locally
+ [x] On a staging site
+ [ ] On a production website

eri-trabiccolo:
## Description
Fixes #2308 

## How has this been tested?
manually

## Types of changes

## Checklist:

### DIFF
--- a/includes/admin/post-types/meta-boxes/class.llms.meta.box.order.transactions.php
+++ b/includes/admin/post-types/meta-boxes/class.llms.meta.box.order.transactions.php
@@ -5,13 +5,13 @@
  * @package LifterLMS/Admin/PostTypes/MetaBoxes/Classes
  *
  * @since 3.0.0
- * @version 3.35.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
 
 /**
- * LLMS_Meta_Box_Order_Transactions class
+ * LLMS_Meta_Box_Order_Transactions class.
  *
  * @since 3.0.0
  * @since 3.35.0 Verify nonces and sanitize `$_POST` data.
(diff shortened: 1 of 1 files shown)


brianhogg:
Need to verify this fixes the bug and that we want the transaction as "trial" in this case instead of "single".

Also add a changelog.

includes/admin/post-types/meta-boxes/class.llms.meta.box.order.transactions.php
```diff
@@ -198,7 +199,9 @@ private function save_transaction( $post_id ) {
 			return $this->add_error( __( 'Refund Error: Missing or invalid payment amount', 'lifterlms' ) );
 		}
 
-		$order = new LLMS_Order( $post_id );
+		$order        = new LLMS_Order( $post_id );
+		$payment_type = $order->has_trial() && ! $order->get_last_transaction( 'llms-txn-succeeded', 'trial' ) ?
```
brianhogg, line 203: I think we also need to check that the trial has a payment amount. Otherwise for a free trial, this is applied against the trial vs. single incorrectly.

brianhogg: Actually since there is already a transaction added during the initial creation of the Order during checkout, it may be fixing that so it's marked as a `trial` payment type vs. single.

brianhogg:
We can pitch this for a future sprint if we want to address this edge case.

PR #2328, opened 2023-01-19
[dev cli] Account for spaces in the file paths

eri-trabiccolo:
Should fix the issue with the spaces in file paths occurring e.g. on Windows.

I tested it trying to reproduce the same "environment": both `npm` executable and the project directory inside directories with a spaces in their name.

Seems to work fine with my tests,
I tested the following `release` commands:
- `prepare` (the one that failed)
- `archive`

I didn't thoroughly tested the `create` command to avoid inadvertently pushing on the repo, but it should work fine.


ideadude:
Looks good. I'm going to merge and use. Will let you know if I run into issues.

PR #2345, opened 2023-02-10
Improve performance when counting course's lessons on the course post list table

eri-trabiccolo:
## Description
About the lessons count in the courses post list table:
The lessons count the way is done right now is not performant:
That is acceptable in most of the cases because we use that on a single course template ecc. But in a list of N courses, that could be pretty expensive if there are many courses with many sections and lessons each.
We’re now using `count( $course->get_lessons('ids') );`, but if you look at how course’s get_lessons() [method works](https://github.com/gocodebox/lifterlms/blob/7.0.1/includes/models/model.llms.course.php#L283), and down the rabbit hole at how section’s get_lessons() [method works](https://github.com/gocodebox/lifterlms/blob/7.0.1/includes/models/model.llms.section.php#L208), you’ll see what I’m talking about.
So I propose this as a viable alternative.

## How has this been tested?
manually

## Types of changes
performance

## Checklist:

### DIFF
--- a/.changelogs/improve-lessons-count-perf.yml
+++ b/.changelogs/improve-lessons-count-perf.yml
@@ -1,0 +1,4 @@
+significance: minor
+type: dev
+entry: Added new `LLMS_Course::get_lessons_count()` method. It can be used in
+  place of `count( LLMS_Course::get_lessons() )` to improve performance.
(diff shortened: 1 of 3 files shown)


ideadude:
This is better than using the full get_lessons query. We may still need some improvements for sites with many many lessons and many other posts in the wp_posts table/etc, but merging for now for the 7.1 release. We can improve this further later if needed.

PR #2377, opened 2023-03-08
Fix notices woocommerce screens

### LINKED ISSUE #2373: Notice appearance borked in WooCommerce screens
### Reproduction Steps

Install WooCommerce and trigger a LifterLMS notice. The appearance of the notice within WooCommerce screens has extra padding.
Screenshot 2023-03-07 at 8 52 38 AM

### Expected Behavior
We should reset padding for our notice to maintain consistent appearance of our notices throughout.

This issue has be recreated:
+ [x] Locally

kimcoleman:
## Description

Fixes #2373

## Screenshots 
Before:
Screenshot 2023-03-07 at 9 08 14 AM

After:
Screenshot 2023-03-08 at 9 35 05 AM

## Checklist:

### DIFF
--- a/.changelogs/fix-notices-woocommerce-screens.yml
+++ b/.changelogs/fix-notices-woocommerce-screens.yml
@@ -1,0 +1,5 @@
+significance: patch
+type: fixed
+links:
+  - "#2373"
+entry: Fixed notice display on WooCommerce dashboard pages.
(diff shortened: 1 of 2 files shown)


eri-trabiccolo (approved):

ideadude:
Note here that we decided to use an ! important in our CSS code here. GASP!

In reality, we are using it exactly as one should. We don't want anything else to override these styles. There is no need to override these styles, vs styles on the frontend that are meant to be cascaded. There is a small edge case of themes for the WP admin, but they are not generally focused on restyling LifterLMS notices, and even if they are, there is still plenty of room to change the styling of these notices. Merging in.

PR #2380, opened 2023-03-09
LifterLMS Resources Page

nrherron92:
## Description
Added a new LifterLMS Resources page to the WordPress backend.

## How has this been tested?
Manually and with run styles

## Types of changes

Added a new page and styling for that page.

## Checklist:


ideadude:
I am removing the 7.2 milestone.

The links and features added here are redundant with existing links on the updated dashboard and in the Addons & More page. The video is nice, but we can perhaps show that in a dismissible box on the dashboard page. If resources are missing, we can add them to the dashboard.

If we keep this, we want to think about a few UI/UX related things:
* The Resources menu item should go further down in the menu, perhaps at the end after Addons & More.
* The framing could be more similar to the updated dashboard page.
* Do we want to redirect to the Resources page after the wizard finishes?

eri-trabiccolo:
closed in favor of #2534

PR #2388, opened 2023-03-16
Making the LifterLMS logo in admin header link to the LifterLMS.com site

kimcoleman:
## Description

This is a small update to have our logo be linked to our website in the admin header. This does not warrant a changlog entry in my opinion.

I also had a small fix to the UTM tracking on the Get Support link when that link is in the header - it should use the "utm_medium" Admin Header.


eri-trabiccolo (approved):
Good for me

PR #2397, opened 2023-03-23
Avoid use of inline CSS in reviews

### LINKED ISSUE #410: [Reviews] Avoid use of Inline CSS.
Hello,

We are trying to make our theme compatible with LifterLMS but the problem is, you have used inline CSS at many places, like in Review List, Review Box etc., you can look at here: 
https://github.com/gocodebox/lifterlms/blob/master/includes/class.llms.review.php#L65
https://github.com/gocodebox/lifterlms/blob/master/includes/class.llms.review.php#L111

Used multiple style attribute in single tag: https://github.com/gocodebox/lifterlms/blob/master/includes/class.llms.review.php#L66

In our theme, we are not able to override your inline CSS and we don't want to use `!important`.

I haven't checked your plugin fully but It may possible you used inline in other places, I'll suggest you to don't use inline CSS.

seothemes:
## Description

This PR removes most inline styles from course reviews HTML and allows remaining CSS to be removed with the `llms_review_custom_styles` filter.

It also adds escaping to all attributes and HTML output previously missing. 

Fixes #410 

## How has this been tested?

Manually

## Types of changes

## Checklist:


includes/class.llms.review.php
```diff
@@ -42,58 +43,82 @@ public function __construct() {
...
 			);
+
 			$posts_array = get_posts( $args );
 
-			$styles = array(
-				'background-color' => '#EFEFEF',
-				'title-color'      => 'inherit',
-				'text-color'       => 'inherit',
-				'custom-css'       => '',
+			$styles = apply_filters(
```
eri-trabiccolo, line 85: Can you add a docblock here? <3

includes/class.llms.review.php
```diff
@@ -42,58 +43,82 @@ public function __construct() {
...
 
 		/**
-		 * Check to see if we are supposed to output the code at all
+		 * Check to see if we are supposed to output the code at all.
 		 */
 		if ( get_post_meta( get_the_ID(), '_llms_display_reviews', true ) ) {
 			?>
 			<div id="old_reviews">
-			<h3><?php echo apply_filters( 'lifterlms_reviews_section_title', __( 'What Others Have Said', 'lifterlms' ) ); ?></h3>
+			<h3><?php echo esc_html( apply_filters( 'lifterlms_reviews_section_title', __( 'What Others Have Said', 'lifterlms' ) ) ); ?></h3>
```
eri-trabiccolo: Can you add a docblock here too?
I think it's a good occasion to improve the documentation. What do you think?

includes/class.llms.review.php
```diff
@@ -42,58 +43,82 @@ public function __construct() {
...
+
+			foreach ( $posts_array as $post ) {
 				?>
-				<div class="llms_review" style="margin:20px 0px; background-color:<?php echo $styles['background-color']; ?>; padding:10px">
-					<h5 style="font-size:17px; color:<?php echo $styles['title-color']; ?>;" style="margin:3px 0px"><strong><?php echo get_the_title( $post->ID ); ?></strong></h5>
-					<h6 style="font-size:13px; color:<?php echo $styles['text-color']; ?>;"><?php echo sprintf( __( 'By: %s', 'lifterlms' ), get_the_author_meta( 'display_name', get_post_field( 'post_author', $post->ID ) ) ); ?></h6>
-					<p style="font-size:15px; color:<?php echo $styles['text-color']; ?>;"><?php echo get_post_field( 'post_content', $post->ID ); ?></p>
+				<div class="llms_review">
+					<h5><strong><?php echo get_the_title( $post->ID ); ?></strong></h5>
+					<h6><?php echo esc_html( __( 'By: ', 'lifterlms' ) . get_the_author_meta( 'display_name', get_post_field( 'post_author', $post->ID ) ) ); ?></h6>
```
eri-trabiccolo: I'd prefer not to change the localized string here, so it doesn't require an update, is there a reason why you prefer string concatenation over `sprintf`?
Also I'd consider that while something like "By: author" makes perfectly sense it might be not that obvious for some locales. See https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#best-practices-for-writing-strings 

```suggestion
					<h6>
					<?php
						// Translators: %s = The author display name.
						echo esc_html( sprintf( __( 'By: %s', 'lifterlms' ), get_the_author_meta( 'display_name', get_post_field( 'post_author', $post->ID ) ) ) );
					?>
					</h6>
```

includes/class.llms.review.php
```diff
@@ -116,26 +141,26 @@ public static function output() {
 			if ( get_post_meta( get_the_ID(), '_llms_multiple_reviews_disabled', true ) && $posts_array ) {
 				?>
 				<div id="thank_you_box">
-					<h2><?php echo apply_filters( 'llms_review_thank_you_text', __( 'Thank you for your review!', 'lifterlms' ) ); ?></h2>
+					<h2><?php echo esc_html( apply_filters( 'llms_review_thank_you_text', __( 'Thank you for your review!', 'lifterlms' ) ) ); ?></h2>
```
eri-trabiccolo: Mind to add a filter docblock?

includes/class.llms.review.php
```diff
@@ -116,26 +141,26 @@ public static function output() {
...
 					<input name="post_ID" value="<?php echo get_the_ID(); ?>" type="hidden" id="post_ID">
-					<input type="submit" class="button" value="<?php _e( 'Leave Review', 'lifterlms' ); ?>" id="llms_review_submit_button">
-				<!--</form>	-->
+					<input type="submit" class="button" value="<?php esc_attr_e( 'Leave Review', 'lifterlms' ); ?>" id="llms_review_submit_button">
+					<!--</form>	-->
 				</div>
-				<div id="thank_you_box" style="display:none;">
-					<h2><?php echo apply_filters( 'llms_review_thank_you_text', __( 'Thank you for your review!', 'lifterlms' ) ); ?></h2>
+				<div class="thank_you_box" id="thank_you_box">
+					<h2><?php echo esc_html( apply_filters( 'llms_review_thank_you_text', __( 'Thank you for your review!', 'lifterlms' ) ) ); ?></h2>
```
eri-trabiccolo: Mind to add a filter docblock?

.changelogs/410-reviews-avoid-use-of-inline-css.yml
```diff
@@ -0,0 +1,5 @@
+significance: patch
```
eri-trabiccolo: I'm not sure this is just a `patch`, it might have impacts on the frontend as we're moving some inline style into a stylesheet, which means less specificity.
```suggestion
significance: minor
```

eri-trabiccolo (changes requested):
Thanks for this fix and for taking the time to "beautify" the code a bit :D 
I just requested some minor stuff like adding filter docblocks and prefer phrases over words translation. <3

seothemes:
New PR #2403

PR #2403, opened 2023-03-28
Avoid use of inline CSS in course reviews

### LINKED ISSUE #410: [Reviews] Avoid use of Inline CSS.

seothemes:
## Description

This PR removes most inline styles from course reviews HTML and allows remaining CSS to be removed with the `llms_review_custom_styles` filter.

It also adds escaping to all attributes and HTML output previously missing. 

Fixes #410 

## How has this been tested?

Manually

## Types of changes

## Checklist:

### DIFF
--- a/.changelogs/reviews-avoid-use-of-inline-css.yml
+++ b/.changelogs/reviews-avoid-use-of-inline-css.yml
@@ -1,4 +1,4 @@
-significance: minor
+significance: patch
 type: changed
 links:
   - "#410"
--- a/.changelogs/reviews-avoid-use-of-inline-css.yml
+++ b/.changelogs/reviews-avoid-use-of-inline-css.yml
@@ -1,0 +1,5 @@
+significance: minor
+type: changed
+links:
+  - "#410"
+entry: Avoid use of inline styles in course reviews.
--- a/includes/class.llms.review.php
+++ b/includes/class.llms.review.php
@@ -3,12 +3,12 @@
  * LifterLMS Course reviews
  *
  * This class handles the front end of the reviews. It is responsible
- * for outputting the HTML on the course page (if reviews are activated)
+ * for outputting the HTML on the course page (if reviews are activated).
  *
  * @package LifterLMS/Classes
  *
- * @since Unknown
- * @version 5.9.0
+ * @since 1.2.7
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -16,9 +16,10 @@ defined( 'ABSPATH' ) || exit;
 /**
  * LLMS_Reviews class
  *
- * @since Unknown
+ * @since 1.2.7
  */
 class LLMS_Reviews {
+
 	/**
 	 * This is the constructor for this class. It takes care of attaching
 	 * the functions in this file to the appropriate actions. These actions are:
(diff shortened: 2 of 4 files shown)


includes/class.llms.review.php
```diff
@@ -3,31 +3,32 @@
...
 class LLMS_Reviews {
+
 	/**
 	 * This is the constructor for this class. It takes care of attaching
 	 * the functions in this file to the appropriate actions. These actions are:
 	 * 1) output after course info
 	 * 2) output after membership info
 	 * 3 & 4) Add function call to the proper AJAX call
 	 *
+	 * @version 3.1.3
```
eri-trabiccolo:
```suggestion
	 * @since 3.1.3
```

includes/class.llms.review.php
```diff
@@ -3,31 +3,32 @@
...
+ * @since 1.2.7
  */
 class LLMS_Reviews {
+
 	/**
 	 * This is the constructor for this class. It takes care of attaching
 	 * the functions in this file to the appropriate actions. These actions are:
 	 * 1) output after course info
 	 * 2) output after membership info
 	 * 3 & 4) Add function call to the proper AJAX call
```
eri-trabiccolo:
```suggestion
	 * This is the constructor for this class. 
	 * 
	 * It takes care of attaching the functions in this file to the 
	 * appropriate actions.
	 * These actions are:
	 * 1) output after course info
	 * 2) output after membership info
	 * 3 & 4) Add function call to the proper AJAX call.
```

includes/class.llms.review.php
```diff
@@ -40,64 +41,110 @@ public function __construct() {
...
+					?>
+					<div class="llms_review">
+						<h5><strong><?php echo get_the_title( $post->ID ); ?></strong></h5>
+						<h6>
+							<?php
+							// Translators: %s = The author display name.
+							echo esc_html( sprintf( __( 'By: %s', 'lifterlms' ), get_the_author_meta( 'display_name', get_post_field( 'post_author', $post->ID ) ) ) );
+							?>
+						</h6>
+						<p><?php echo esc_html( get_post_field( 'post_content', $post->ID ) ); ?></p>
```
eri-trabiccolo, line 135: mmm `get_post_field()` should sanitize the field already, can you check we need the `esc_html()` here?

eri-trabiccolo (changes requested):

includes/class.llms.review.php
seothemes: @eri-trabiccolo I think you're right, `get_post_field()` does sanitize the value

eri-trabiccolo (approved):
Thanks!

.changelogs/reviews-avoid-use-of-inline-css.yml
```diff
@@ -0,0 +1,5 @@
+significance: minor
```
ideadude:
```suggestion
significance: patch
```

PR #2407, opened 2023-04-03
Added helper function for floating values shown in average column in reporting.

### LINKED ISSUE #2237: [Reporting Section] We can eliminate zero decimals for faster readability
The reporting section for the assignment grades column shows 100.000%. I am not sure if that is mandatory. I think we can just show 100% for clarity and readability. Also, 85.3% seems more appropriate to me than 85.381%.

Please let me know if this is possible and seems logical.

actuallyakash:
## Description
Added a helper function for returning precision with a filter for floating values in the reporting section for quizzes and assignment's average column.

Fixes #2237

## How has this been tested?
Manually.

## Types of changes
Bug fix

## Checklist:

### DIFF
--- a/.changelogs/issue_2237.yml
+++ b/.changelogs/issue_2237.yml
@@ -1,0 +1,6 @@
+significance: patch
+type: changed
+comment: Added a helper function for returning precision with filter for
+  floating values in the reporting section for quizzes and assignment's average
+  column.
+entry: Added helper function for precision for floating values in reporting section.
--- a/.changelogs/issue_2237.yml
+++ b/.changelogs/issue_2237.yml
@@ -1,4 +1,4 @@
-significance: minor
+significance: patch
 type: dev
 comment: Added a helper function `llms_get_floats_rounding_precision()` for returning precision with filter `lifterlms_floats_rounding_precision` for rounding off floating values in the reporting section for quizzes average column.
 links:
--- a/.changelogs/issue_2237.yml
+++ b/.changelogs/issue_2237.yml
@@ -1,8 +1,7 @@
 significance: minor
 type: dev
 comment: Added a helper function `llms_get_floats_rounding_precision()` for returning precision with filter for
-  rounding off floating values in the reporting section for quizzes and
-  assignment's average column.
+  rounding off floating values in the reporting section for quizzes average column.
 links:
   - "#2237"
 entry: Added helper function `llms_get_floats_rounding_precision()` to return precision for rounding off floating
--- a/.changelogs/issue_2237.yml
+++ b/.changelogs/issue_2237.yml
@@ -1,7 +1,6 @@
 significance: minor
 type: dev
-comment: Added a helper function `llms_get_floats_rounding_precision()` for returning precision with filter for
-  rounding off floating values in the reporting section for quizzes average column.
+comment: Added a helper function `llms_get_floats_rounding_precision()` for returning precision with filter `lifterlms_reporting_average_precision` for rounding off floating values in the reporting section for quizzes average column.
 links:
   - "#2237"
 entry: Added helper function `llms_get_floats_rounding_precision()` to return precision for rounding off floating
--- a/.changelogs/issue_2237.yml
+++ b/.changelogs/issue_2237.yml
@@ -1,7 +1,7 @@
 significance: minor
 type: dev
-comment: Added a helper function `llms_get_floats_rounding_precision()` for returning precision with filter `lifterlms_reporting_average_precision` for rounding off floating values in the reporting section for quizzes average column.
+comment: Added a helper function `llms_get_floats_rounding_precision()` for returning precision with filter `lifterlms_floats_rounding_precision` for rounding off floating values in the reporting section for quizzes average column.
 links:
   - "#2237"
 entry: Added helper function `llms_get_floats_rounding_precision()` to return precision for rounding off floating
-  values in reporting section.
+  values with filter hook `lifterlms_floats_rounding_precision` to filter precision value for reporting section.
--- a/.changelogs/issue_2237.yml
+++ b/.changelogs/issue_2237.yml
@@ -1,6 +1,9 @@
-significance: patch
-type: changed
-comment: Added a helper function for returning precision with filter for
-  floating values in the reporting section for quizzes and assignment's average
-  column.
-entry: Added helper function for precision for floating values in reporting section.
+significance: minor
+type: dev
+comment: Added a helper function `llms_get_floats_rounding_precision()` for returning precision with filter for
+  rounding off floating values in the reporting section for quizzes and
+  assignment's average column.
+links:
+  - "#2237"
+entry: Added helper function `llms_get_floats_rounding_precision()` to return precision for rounding off floating
+  values in reporting section.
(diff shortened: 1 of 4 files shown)


.changelogs/issue_2237.yml
```diff
@@ -0,0 +1,6 @@
+significance: patch
+type: changed
+comment: Added a helper function for returning precision with filter for
+  floating values in the reporting section for quizzes and assignment's average
+  column.
+entry: Added helper function for precision for floating values in reporting section.
```
eri-trabiccolo: I would see the name of the function here.
The entry test will appear in the plugin's CHANGELOG.md and in the release note's post, e.g.:
https://make.lifterlms.com/2023/03/23/lifterlms-version-7-1-2/

You can also add a reference to the new filter added.

.changelogs/issue_2237.yml
```diff
@@ -0,0 +1,6 @@
+significance: patch
```
eri-trabiccolo, line 1:
```suggestion
significance: minor
```
This can be a minor change rather than a `patch`

.changelogs/issue_2237.yml
```diff
@@ -0,0 +1,6 @@
...
+type: changed
```
eri-trabiccolo:
```suggestion
type: dev
```

.changelogs/issue_2237.yml
eri-trabiccolo: Also you can link this entry to the issue 2237.

Additionally, I would add another changelog entry (so another file) because we're changing the default precision on the quizzes table.

While you're there can you check where else we're using that 3 digits rounding?

includes/functions/llms-functions-content.php
```diff
@@ -112,6 +112,24 @@ function llms_get_post_sales_page_content( $post, $default = '' ) {
...
+ * By default the database return floating-point values and  function
+ * percision to round off the decimal value.
+ *
+ * @since [version]
+ *
+ * @return int
+ */
+function llms_get_reporting_round_precision() {
+
+	$precision = apply_filters( 'lifterlms_average_decimal', 1 );
```
eri-trabiccolo: I would use a more contextualized yet less specific (average? can't we use it in general?) hook name. Dunno, something like:

`lifterlms_reporting_floats_rounding_precision`

or something of the sort.
What do you think?

Also, could you please add a docblock for this new filter?

What do you think if we run this filter once? Like only the first time we call this function?

includes/admin/reporting/tables/llms.table.quizzes.php
```diff
@@ -199,7 +199,7 @@ protected function get_data( $key, $data ) {
 						$grade += $attempt->get( 'grade' );
 					}
 
-					$value = round( $grade / $attempts, 3 ) . '%';
+					$value = round( $grade / $attempts, llms_get_reporting_round_precision() ) . '%';
```
eri-trabiccolo: This change would require a docblock changelog update to me.
Also the file header's `version` tag should be updated as well.

eri-trabiccolo (changes requested):

includes/functions/llms-functions-content.php
```diff
@@ -112,6 +112,24 @@ function llms_get_post_sales_page_content( $post, $default = '' ) {
...
+ * Retrieve the decimal value for average column in Reporting.
+ *
+ * By default the database return floating-point values and  function
+ * percision to round off the decimal value.
+ *
+ * @since [version]
+ *
+ * @return int
+ */
+function llms_get_reporting_round_precision() {
```
eri-trabiccolo: Also, I don't feel this is the right file for this helper.
In this file we have helpers related to page/post's "post_content" alteration.

This function should be moved somewhere else.

includes/functions/llms-functions-content.php
```diff
@@ -112,6 +112,24 @@ function llms_get_post_sales_page_content( $post, $default = '' ) {
 
 }
 
+/**
+ * Retrieve the decimal value for average column in Reporting.
+ *
+ * By default the database return floating-point values and  function
+ * percision to round off the decimal value.
```
eri-trabiccolo: this sentence sound a bit weird to me, also two spaces between `and` and `function` are too many :D

.changelogs/issue_2237.yml
actuallyakash: Done. I've created the separate changelog in the lifterlms-assignments repo, where it belongs.

includes/functions/llms-functions-content.php
actuallyakash: done! thanks 👍

includes/functions/llms-functions-content.php
actuallyakash: done!

includes/functions/llms-functions-content.php
actuallyakash: my bad, I forgot to update this sentence. Now fixed.

includes/admin/llms.functions.admin.php
```diff
@@ -288,3 +288,33 @@ function llms_merge_code_button( $target = 'content', $echo = true, $codes = arr
...
+
+	if ( is_null( $precision ) ) {
+		/**
+		 * Filters the precision for round function for floating values.
+		 *
+		 * @since [version]
+		 *
+		 * @param int $precision Precision for round function for floating values.
+		 */
+		$precision = apply_filters( 'lifterlms_reporting_average_precision', 2 );
```
eri-trabiccolo:
```suggestion
		$precision = apply_filters( 'lifterlms_floats_rounding_precision', 2 );
```

.changelogs/issue_2237.yml
```diff
@@ -0,0 +1,7 @@
+significance: minor
+type: dev
+comment: Added a helper function `llms_get_floats_rounding_precision()` for returning precision with filter `lifterlms_reporting_average_precision` for rounding off floating values in the reporting section for quizzes average column.
+links:
+  - "#2237"
+entry: Added helper function `llms_get_floats_rounding_precision()` to return precision for rounding off floating
```
eri-trabiccolo, line 6: Can you add a reference to the newly added filter hook too? Thanks! <3
Like:
Added filter hook `lifterlms_floats_rounding_precision`, to filter bla bla bla ?

eri-trabiccolo (changes requested):

.changelogs/issue_2237.yml
actuallyakash: Done! 👍

includes/admin/llms.functions.admin.php
```diff
@@ -288,3 +288,33 @@ function llms_merge_code_button( $target = 'content', $echo = true, $codes = arr
 	return $html;
 
 }
+
+/**
+ * Retrieve the precision for round function for floating values.
+ *
+ * The function returns precision to round off the decimal values
+ * for 'average' column in reporting tables.
```
eri-trabiccolo: I missed this, can we have a more generic description here? We're gonna use it in several places.
Or even remove it.
Thanks

eri-trabiccolo (changes requested):

includes/admin/llms.functions.admin.php
actuallyakash: Removed 👍

eri-trabiccolo (approved):
thanks

PR #2409, opened 2023-04-13
Fixes #2191

### LINKED ISSUE #2191: [7.x] Add new data retention settings for orders and user accounts
Introduces a collection of admin settings designed to automatically delete abandoned orders from the database after a specified time period (days/weeks/months/years)

+ Inactive accounts - an account which has no enrollments
+ Pending orders - orders with then "llms-pending" status

When running the cron *pending orders* should be deleted first and then the inactive account check should be run.

The user settings should be a number input accepting any whole number greater than 0 and a select drop down to select the time period (days/weeks/months/years)

A cron should run for each setting to query orders/users that match the setting criteria and delete them accordingly.

The settings can be left blank (no number value) to disable the cron (the data is retained)

The default value of both settings should be blank

With 7.0 and gateways that subscribe to the new AJAX checkout flow (currently on PayPal CP) a pending order will *not* have an associated user account but with other gateways the user account will likely exist and be considered an "inactive" account. Deleting an order should *never* delete the user, instead the user should be deleted independently by the inactive account setting.

---

I think this should be added shortly after 7.0 release if not with the 7.0 release.

actuallyakash:
## Description
Added feature for deletion of pending orders and inactive users after held duration.

Fixes #<2191

## How has this been tested?
Manually.

## Types of changes
New feature

## Checklist:

### DIFF
--- a/includes/admin/llms.functions.admin.php
+++ b/includes/admin/llms.functions.admin.php
@@ -288,3 +288,60 @@ function llms_merge_code_button( $target = 'content', $echo = true, $codes = arr
 	return $html;
 
 }
+
+/**
+ * Deletes all pending orders after held duration.
+ *
+ * @since [version]
+ *
+ * @return null
+ */
+function llms_delete_pending_orders() {
+
+	$days = get_option( 'lifterlms_pending_orders_deletion' );
+
+	if ( ! $days ) {
+		return;
+	}
+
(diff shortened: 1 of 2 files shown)


actuallyakash:
Reopened in new PR.
https://github.com/gocodebox/lifterlms/pull/2410

PR #2417, opened 2023-04-20
Let's use strpos instead of str_starts_with to avoid fatal errors

### LINKED ISSUE #2415: Customers websites are crashing due to undefined function str_starts_with()
### Reproduction Steps
1. Spin up an environment that uses PHP 7 (not 8).
2. Create a fresh install of WordPress install (any version **below 5.6**)
3. Make sure WooCommerce is not installed.
4. Install the latest version of LifterLMS core plugin.

### Expected Behavior
The plugin must install.

### Actual Behavior
The plugin will not install. Or, the website will crash.

### Error Messages / Logs
```
PHP Fatal error: Uncaught Error: Call to undefined function str_starts_with()
```

PHP 7
WordPress version below 5.6
No WooCommerce installed.

### Other Notes
WordPress 5.6 onwards polyfills this in `wp-includes\compat.php`.
WooCommerece also polyfills this in `wp-content/plugins/woocommerce/vendor/symfony/polyfill-php80/bootstrap.php`.

ideadude:
Even though `str_starts_with()` should be available when using our minimum recommended versions of PHP and WP, using `strpos(...) !== false` is easy enough and avoids fatal errors for some of our users.

In general, we want to avoid using new functions, methods, and features of PHP/WP until we have too. We don't want to go too far out of our way to support older versions, but if the required change is small, won't hinder our development, and is not at risk of being deprecated, we should lean toward using the older code to maximize compatibility.

This is a different stance than the LifterLMS dev team has taken in the past.

## Description

Fixes #2415

## How has this been tested?
Locally. After the change, the LifterLMS admin header and footer are still loading properly on pages they should and not loading on pages they shouldn't.

## Types of changes
Bug Fix

## Checklist:

### DIFF
--- a/.changelogs/issue_2415.yml
+++ b/.changelogs/issue_2415.yml
@@ -1,0 +1,5 @@
+significance: patch
+type: fixed
+links:
+  - "#2415"
+entry: Using `strpos()` instead of `str_starts_with()` for compatibility.
--- a/includes/admin/class-llms-admin-header.php
+++ b/includes/admin/class-llms-admin-header.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Admin/Classes
  *
  * @since 7.1.0
- * @version 7.1.2
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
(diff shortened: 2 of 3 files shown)


eri-trabiccolo (changes requested):
As per slack conversation

eri-trabiccolo (approved):
Looks good to me!

PR #2437, opened 2023-05-17
Added support for multiple difficulties.

### LINKED ISSUE #2433: Display multiple course difficulties when multiple course difficulties are added.
### Reproduction Steps
We enabled multiple course difficulties as shown [in this screenshot](https://share.getcloudapp.com/4guNok78) when using the Gutenburg editor. However, as you can see in here [lifterlms/includes/models/model.llms.course.php](https://github.com/gocodebox/lifterlms/blob/9db9175eb5a70f2fcf05410898f52d13dd8ffb55/includes/models/model.llms.course.php#LL227C2-L244C3), the `get_difficulty` function only returns one level of difficulty instead of all available levels of difficulty.

### Expected Behavior
It should show multiple levels of difficulty on the front-end [as you can see here](https://share.getcloudapp.com/o0uY2rEZ). 

### Actual Behavior
It only shows one level of difficulty, [as you can see here](https://share.getcloudapp.com/Z4ungQov).

+ [x] Locally
+ [ ] On a staging site
+ [x] On a production website

### Slack Discussion
Conversation before creating this ticket can be accessed [here](https://teamcodebox.slack.com/archives/C0AM6AGF4/p1684145201877879). 

### Affected tickets
An example customer that will benefit from this change is HS-216321.

actuallyakash:
## Description
Added support for showing multiple difficulties when using Gutenberg Editor.

Fixes #2433

## How has this been tested?
Manually.

## Types of changes
New feature

## Checklist:

### DIFF
--- a/includes/models/model.llms.course.php
+++ b/includes/models/model.llms.course.php
@@ -5,7 +5,7 @@
  * @package LifterLMS/Models/Classes
  *
  * @since 1.0.0
- * @version 7.1.0
+ * @version [version]
  */
 
 defined( 'ABSPATH' ) || exit;
@@ -218,6 +218,7 @@ class LLMS_Course extends LLMS_Post_Model implements LLMS_Interface_Post_Instruc
 	 *
 	 * @since 1.0.0
 	 * @since 3.24.0 Unknown.
+	 * @since [version] Added support for showing multiple difficulties.
 	 *
 	 * @param string $field Optional. Which field to return from the available term fields.
 	 *                      Any public variables from a WP_Term object are acceptable: term_id, name, slug, and more.
@@ -234,11 +235,15 @@ class LLMS_Course extends LLMS_Post_Model implements LLMS_Interface_Post_Instruc
 
 		} else {
 
+			$difficulties = array();
+
 			foreach ( $terms as $term ) {
 
-				return $term->$field;
+				$difficulties[] = $term->$field;
 
 			}
+
+			return implode( ', ', $difficulties );
 		}
 
 	}
--- a/includes/models/model.llms.course.php
+++ b/includes/models/model.llms.course.php
@@ -223,22 +223,18 @@ class LLMS_Course extends LLMS_Post_Model implements LLMS_Interface_Post_Instruc
 	 * @param string $field Optional. Which field to return from the available term fields.
 	 *                      Any public variables from a WP_Term object are acceptable: term_id, name, slug, and more.
 	 *                      Default is 'name'.
-	 * @return string|array
+	 * @return string
 	 */
 	public function get_difficulty( $field = 'name' ) {
 
 		$terms = get_the_terms( $this->get( 'id' ), 'course_difficulty' );
 
 		if ( false === $terms ) {
-
 			return '';
-
-		} else {
-
-			$difficulties = wp_list_pluck( $terms, $field );
-
-			return implode( ', ', $difficulties );
 		}
+
+		$difficulties = wp_list_pluck( $terms, $field );
+		return implode( ', ', $difficulties );
 
 	}
 
--- a/includes/models/model.llms.course.php
+++ b/includes/models/model.llms.course.php
@@ -223,7 +223,7 @@ class LLMS_Course extends LLMS_Post_Model implements LLMS_Interface_Post_Instruc
 	 * @param string $field Optional. Which field to return from the available term fields.
 	 *                      Any public variables from a WP_Term object are acceptable: term_id, name, slug, and more.
 	 *                      Default is 'name'.
-	 * @return string | array
+	 * @return string|array
 	 */
 	public function get_difficulty( $field = 'name' ) {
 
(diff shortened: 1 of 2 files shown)


includes/models/model.llms.course.php
```diff
@@ -234,11 +235,15 @@ public function get_difficulty( $field = 'name' ) {
 
 		} else {
 
+			$difficulties = array();
+
 			foreach ( $terms as $term ) {
 
-				return $term->$field;
+				$difficulties[] = $term->$field;
```
eri-trabiccolo: we're not doing any check on the existence of the `$term->field`, so this will raise a fatal if the requested field doesn't exist.
I think we can live with it, but in this case then we can use [`wp_list_pluck()`](https://developer.wordpress.org/reference/functions/wp_list_pluck/).

I wonder though if we should add the separator as new parameter.
I also wonder if we want to add an additional parameter like return type... array or string.
Your opinion?

eri-trabiccolo (changes requested):

includes/models/model.llms.course.php
actuallyakash: we can use `property_exist` to check if `$term->$field` exists.
Like this:

```
else {

	foreach ( $terms as $term ) {
			
		if( property_exists( $term, $field ) ) {
			$difficulties = wp_list_pluck( $terms, $field );
		}

	}

	return implode( ', ', $difficulties );
}
```

> I wonder though if we should add the separator as new parameter.

parameter as in? where?

> I also wonder if we want to add an additional parameter like return type... array or string.

yeah, it would be good. Like this maybe:

` * @return string | array`

includes/models/model.llms.course.php
```diff
@@ -218,11 +218,12 @@ public function get_categories( $args = array() ) {
 	 *
 	 * @since 1.0.0
 	 * @since 3.24.0 Unknown.
+	 * @since [version] Added support for showing multiple difficulties.
 	 *
 	 * @param string $field Optional. Which field to return from the available term fields.
 	 *                      Any public variables from a WP_Term object are acceptable: term_id, name, slug, and more.
 	 *                      Default is 'name'.
-	 * @return string
+	 * @return string | array
```
eri-trabiccolo:
```suggestion
	 * @return string|array
```

includes/models/model.llms.course.php
```diff
@@ -236,9 +237,14 @@ public function get_difficulty( $field = 'name' ) {
 
 			foreach ( $terms as $term ) {
 
-				return $term->$field;
+				if ( property_exists( $term, $field ) ) {
 
+					$difficulties = wp_list_pluck( $terms, $field );
+
+				}
 			}
```
eri-trabiccolo: why are we keeping the foreach?
You're doing the same `wp_list_pluck( $terms, $field );` for each `$term` in `$terms`.
I see you want to make sure the property exists, but we didn't do that check before so, as I already said, we can live not doing that check. The whole foreach then becomes just:
`$difficulties = wp_list_pluck( $terms, $field );`

eri-trabiccolo (changes requested):

includes/models/model.llms.course.php
actuallyakash: thanks! fixed now

includes/models/model.llms.course.php
eri-trabiccolo: @actuallyakash 
> parameter as in? where?

Like a new parameter of the function:
`	public function get_difficulty( $field = 'name', $separator = ', '  ) {`


> yeah, it would be good. Like this maybe:

>  `* @return string | array`

Yeah like that, but then you have to return an array, otherwise what that  stays for?
The idea would be to optionally return an array instead of a string, depending on an additional parameter. But let's not do it for now.

includes/models/model.llms.course.php
```diff
@@ -218,11 +218,12 @@ public function get_categories( $args = array() ) {
 	 *
 	 * @since 1.0.0
 	 * @since 3.24.0 Unknown.
+	 * @since [version] Added support for showing multiple difficulties.
 	 *
 	 * @param string $field Optional. Which field to return from the available term fields.
 	 *                      Any public variables from a WP_Term object are acceptable: term_id, name, slug, and more.
 	 *                      Default is 'name'.
-	 * @return string
+	 * @return string|array
```
eri-trabiccolo: as discussed, since we don't return an array, there's no reason to modify this, we always return a string.
```suggestion
	 * @return string
```

includes/models/model.llms.course.php
```diff
@@ -234,11 +235,9 @@ public function get_difficulty( $field = 'name' ) {
...
 		} else {
 
-			foreach ( $terms as $term ) {
+			$difficulties = wp_list_pluck( $terms, $field );
 
-				return $term->$field;
-
-			}
+			return implode( ', ', $difficulties );
 		}
```
eri-trabiccolo: The whole method then can be simplified as following:
```
public function get_difficulty( $field = 'name' ) {

	$terms = get_the_terms( $this->get( 'id' ), 'course_difficulty' );
	if ( false === $terms ) {
		return '';
	}

	$difficulties = wp_list_pluck( $terms, $field );
	return implode( ', ', $difficulties );

}
```

do you agree?

eri-trabiccolo (changes requested):

eri-trabiccolo (approved):

includes/models/model.llms.course.php
actuallyakash: yes, it's better.

PR #2452, opened 2023-06-05
Enhancement/removing old course builder button

ideadude:
## Description
In version 7.2 we moved the course builder links into the header and sidebar of the block editor. So let's remove this redundant meta box.

Related: https://github.com/gocodebox/lifterlms-blocks/pull/186/files

## How has this been tested?

Manually on local. The button is removed. When I run with the latest lifterlms-blocks updates, I see the other 2 new buttons.

Don't think we need a changelog since we will have the one from lifterlms-blocks.

## Checklist:


ideadude:
This PR is no good based on how the button is being moved. The issue seems to be related to running Sky Pilot (maybe other FSE themes). We will have to address that bug directly.

PR #2457, opened 2023-06-10
Disable cron

thomasplevy:
Please disable crons since they fail nightly and send me fail notifications since I was the last commit on the workflow (I guess).

Thank you.


eri-trabiccolo:
Thanks for this @thomasplevy.
We're gonna re-enable it once fixed, and push another commit on the workflow.

PR #2458, opened 2023-06-10
Disable failing workflow cron

thomasplevy:
Please disable failing workflow crons so I stop getting nightly notifications for failed work flows.

Thank you <3


eri-trabiccolo:
Thanks for this @thomasplevy 
We're gonna re-enable it once fixed.

PR #2466, opened 2023-06-22
Issue/2432

### LINKED ISSUE #2432: Update the order links when BuddyPress is active.
With BuddyPress and LifteLMS active, when you visit `/members/admin/courses/orders/` and click on any order link (including the "View" button), it redirects you to the `/dashboard/orders/<order number>` instead of `/members/admin/courses/<order number>`. You can [watch this video](https://share.getcloudapp.com/Z4un0peB) (sent to the customer) to see what I mean.

When you visit `/members/admin/courses/orders/` and click on any order link (including the "View" button), it should redirect you to  `/members/admin/courses/<order number>` instead of `/dashboard/orders/<order number>`.

You can override this with a child theme, but it's better to fix this from within LifterLMS.

**Side effects**
This should also fix the problem with BuddyBoss.

**Affected tickets**
Affected customers include [HS-216423](https://secure.helpscout.net/search/?query=216423).

actuallyakash:
## Description
Updated the order links (including the "View" button) with the Order ID so the link will be `/members/admin/courses/<order number>` instead of `/dashboard/orders/<order number>`.

Fixes #2432

## How has this been tested?
Manually.

## Types of changes
Bug fix

## Checklist:

### DIFF
--- a/templates/myaccount/my-orders.php
+++ b/templates/myaccount/my-orders.php
@@ -31,7 +31,7 @@ defined( 'ABSPATH' ) || exit;
 			<?php foreach ( $orders['orders'] as $order ) : ?>
 				<tr class="llms-order-item <?php echo $order->get( 'status' ); ?>" id="llms-order-<?php $order->get( 'id' ); ?>">
 					<td data-label="<?php _e( 'Order', 'lifterlms' ); ?>: ">
-						<a href="<?php echo $order->get_view_link(); ?>">#<?php echo $order->get( 'id' ); ?></a>
+						<a href="<?php echo $order->get( 'id' ); ?>">#<?php echo $order->get( 'id' ); ?></a>
 						<span class="llms-status <?php echo $order->get( 'status' ); ?>"><?php echo $order->get_status_name(); ?></span>
 					</td>
 					<td data-label="<?php _e( 'Date', 'lifterlms' ); ?>: "><?php echo $order->get_date( 'date', 'F j, Y' ); ?></td>
@@ -50,7 +50,7 @@ defined( 'ABSPATH' ) || exit;
 						<?php endif; ?>
 					</td>
 					<td>
-						<a class="llms-button-primary small" href="<?php echo $order->get_view_link(); ?>"><?php _e( 'View', 'lifterlms' ); ?></a>
+						<a class="llms-button-primary small" href="<?php echo $order->get( 'id' ); ?>"><?php _e( 'View', 'lifterlms' ); ?></a>
 					</td>
 				</tr>
 			<?php endforeach; ?>
--- a/templates/myaccount/my-orders.php
+++ b/templates/myaccount/my-orders.php
@@ -9,6 +9,8 @@
  */
 
 defined( 'ABSPATH' ) || exit;
+
+$base_url = ! empty( $_SERVER['REQUEST_URI'] ) ? esc_url( home_url( wp_unslash( $_SERVER['REQUEST_URI'] ) ) ) : admin_url(); // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
 ?>
 
 <div class="llms-sd-section llms-my-orders">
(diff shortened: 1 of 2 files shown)


eri-trabiccolo (changes requested):
I'm not totally convinced of this even if it makes sense :D
For example for sure this doesn't work with plain permalinks (in the LifterLMS dashboard page). Now I see there are various things that do not work with plain permalinks (e.g. advanced videos), or rest api, so we could skip that case and make pretty permalinks a requirement.
But your solution won't work either with Numeric permalinks, or any custom permalink structure which misses the trailing slash.

Anyways:
1) I'd use `esc_url()`.
2) I'd check everything works as expected, e.g. can you, from members/orders/1234, switch payment source of a recurring payment? Can you cancel a subscription etc? (test with payment gateways)
3) Also, as **always**:
a) add a changelog item
b) update the `@version` file tag.

templates/myaccount/my-orders.php
```diff
@@ -31,7 +31,7 @@
 			<?php foreach ( $orders['orders'] as $order ) : ?>
 				<tr class="llms-order-item <?php echo $order->get( 'status' ); ?>" id="llms-order-<?php $order->get( 'id' ); ?>">
 					<td data-label="<?php _e( 'Order', 'lifterlms' ); ?>: ">
-						<a href="<?php echo $order->get_view_link(); ?>">#<?php echo $order->get( 'id' ); ?></a>
+						<a href="<?php echo esc_url( home_url( "/members/admin/courses/orders/{$order->get('id')}/" ) ); ?>">#<?php echo $order->get( 'id' ); ?></a>
```
eri-trabiccolo: mmm but this way this link will only work on the buddyboss/buddypress profile page.
So here we're replacing a problem with another :D
I think this commit doesn't have so much sense and should be reverted, but feel free to tell me I'm wrong, I might have missed something...

eri-trabiccolo (changes requested):
(to be done after the cycle)

templates/myaccount/my-orders.php
actuallyakash: What do you suggest I should use?

By using esc_url() like this `esc_url( $order->get( 'id' )` it doesn't work and converts the ID into unexpected output.

templates/myaccount/my-orders.php
```diff
@@ -9,6 +9,8 @@
  */
 
 defined( 'ABSPATH' ) || exit;
+
+$base_url = ! empty( $_SERVER['REQUEST_URI'] ) ? esc_url( home_url( wp_unslash( $_SERVER['REQUEST_URI'] ) ) ) : admin_url(); // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
```
eri-trabiccolo: why `admin_url()`?

eri-trabiccolo: also, on the student dashboard, this has exactly the same problem when using permalinks without trailing slash:



or using ugly permalinks (?p=124) - although we don't really care of this.

templates/myaccount/my-orders.php
eri-trabiccolo: I'm reviewing the code and finding it doesn't fix the issue 100%.
I don't have a ready solution at the moment, I should spend some time on it, otherwise I would have already done a PR :P 

If you want to spend some more time on it it's fine, but not that much (a couple of hours?).
If you don't come up with a solution **it's fine**, you can add a comment to the issue with your findings etc, and remove the assignation to yourself and change the Development status from "In Progress" to "To Do", so that someone else would pick it up. <3

eri-trabiccolo: @actuallyakash news on this?

actuallyakash: @eri-trabiccolo 

> I'm reviewing the code and finding it doesn't fix the issue 100%.

What issues are you referring to?

templates/myaccount/my-orders.php
actuallyakash: sorry, didn't noticed the `admin_url()`, it was meant to be empty.

Fixed it. Rocco, I think the image isn't uploaded correctly, can you upload it again?

eri-trabiccolo: Schermata 2023-07-21 alle 17 15 04

brianhogg:
I don't think this implementation makes sense (replacing `get_view_link()` with a mash of the server `REQUEST_URI` which is not reliable, or blank if it doesn't exists), though we would like to solve the root issue in some way.

PR #2486, opened 2023-07-18
Use dashboard as default value for navigation link block

### LINKED ISSUE #2465: LifterLMS Navigation Link Block doesn't work properly when the default link is used
### Reproduction Steps

+ using the Site Editor add a LifterLMS Link block to the menu
+ Do not modify the link selection: the Dashboard should be selected by default
+ Save and check the front-end

### Expected Behavior

+ I would expect to see the Dashboard menu item

### Actual Behavior

+ No Dashboard menu item

I think we can _safely_ consider the attributes `'label' = 'Dashboard'` and `'page' = 'dashboard'` when they're not set (yet).
Workaround:
select any page, then select the Dashboard one and save.

### Error Messages / Logs

```
[20-Jun-2023 06:58:33 UTC] PHP Warning:  Undefined array key "page" in /Users/rock/Dev/plugins/lifterlms/includes/class.llms.nav.menus.php on line 383

```

LifterLMS 7.2.1

+ [x] Locally
+ [x] On a staging site
+ [x] On a production website

n/a

seothemes:
## Description

Adds default value (Dashboard) to LLMS Navigation Link block. Previously no default value.

Fixes #2465

## How has this been tested?

Manually

## Types of changes

## Checklist:

### DIFF
--- a/.changelogs/nav-link-block-default.yml
+++ b/.changelogs/nav-link-block-default.yml
@@ -1,0 +1,3 @@
+significance: minor
+type: changed
+entry: Use dashboard as default value for navigation link block.
--- a/.changelogs/nav-link-block-default.yml
+++ b/.changelogs/nav-link-block-default.yml
@@ -1,3 +1,5 @@
-significance: minor
-type: changed
+significance: patch
+type: fixed
+links:
+  - "#2465"
 entry: Use dashboard as default value for navigation link block.
--- a/.changelogs/nav-link-block-default.yml
+++ b/.changelogs/nav-link-block-default.yml
@@ -2,4 +2,4 @@ significance: patch
 type: fixed
 links:
   - "#2465"
-entry: Use dashboard as default value for navigation link block.
+entry: Use student dashboard as default value for navigation link block.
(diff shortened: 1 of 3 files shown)


.changelogs/nav-link-block-default.yml
```diff
@@ -0,0 +1,3 @@
+significance: minor
+type: changed
+entry: Use dashboard as default value for navigation link block.
```
eri-trabiccolo:
```suggestion
significance: patch
type: fixed
links:
  - "#2465"
entry: Use dashboard as default value for navigation link block.
```
I'd say this is more a bugfix than a minor change.
Also added a link to the issue.

eri-trabiccolo (changes requested):
(changelog item to change)
This works, but still in the editor the Link is gray, as if it was not set, is there anything we can do for that?
Schermata 2023-07-18 alle 09 53 21

.changelogs/nav-link-block-default.yml
seothemes: @eri-trabiccolo have just fixed the changelog entry. For the link in the editor, it should work now but requires JS to be rebuilt since the default values are set in block.json

eri-trabiccolo: @seothemes sorry, I missed the rebuild :D my bad!

eri-trabiccolo (approved):
Thanks <3

PR #2493, opened 2023-07-21
Added support for all Post Statuses for sales reporting.

### LINKED ISSUE #2490: Private course/ membership is not visible for filtering in sales page
### Reproduction Steps

1. Create a private course.
2. Visit LifterLMS > Reporting > Sales tab.
3. Search for that private course in the course box, it will not visible.

screely-1689758923592

4. Buy that private course.
5. Now visit LifterLMS > Reporting > Sales tab. The private course price will be count as a Net sale and number of sales but the course will not visible in the filter area.

screely-1689760069863

### Expected Behavior

The private course/ membership should be found in the filter area when we will search for that private course. Because when we are purchasing the private course it's visible on the net sale but not on the filter area(course box).

### Actual Behavior

When we purchase the private course the price is counting on the sale page but when we searching for that course on the sale page filter area it's showing "Not found"

LifterLMS 7.2.1

+ [x] Locally
+ [ ] On a staging site
+ [x] On a production website

HS-220173

actuallyakash:
## Description
The user was not able to filter sales reporting for a course with Private | Draft | Pending post statuses. This PR fixes the issue.

Fixes #2490 

## How has this been tested?
Manually.

## Types of changes
Bug fix

## Checklist:


assets/js/llms-admin.js
```diff
@@ -58,7 +59,7 @@
 				multiple: false,
 				placeholder: undefined !== LLMS.l10n ? LLMS.l10n.translate( 'Select a Course/Membership' ) : 'Select a Course/Membership',
 				post_type: self.attr( 'data-post-type' ) || 'post',
-				post_statuses: self.attr( 'data-post-statuses' ) || 'publish',
+				post_statuses: self.attr( 'data-post-statuses' ) || 'publish,private',
```
eri-trabiccolo: this is a bit more than allowing support for private posts only in the reporting/sales page.
I'd like a more specific approach, or if you think we should extend this default behavior we can discuss about that.

eri-trabiccolo (changes requested):

assets/js/llms-admin.js
actuallyakash: What do you suggest?

Should I make a 'data-post-statuses' attribute in the `Courses` select field with all the `post_statuses`?
Perhaps, like this: `data-post-statuses="<?php echo implode( ',', array_keys( get_post_statuses() ) ); ?>"`
Both for courses and memberships?

OR

Should I return all the default `post_statuses` of WordPress for all cases like I just did for Private posts?

eri-trabiccolo: > What do you suggest?

I actually wanted your opinion on this :D
I'm suggesting the current solution isn't specific and it's partial.

> Should I make a 'data-post-statuses' attribute in the Courses select field with all the post_statuses?

isn't there already such attribute? I think you should update it with the desired statuses if it exists, or add it if it doesn't exist, yeah.

> all the `post_statuses`?

All the required post statuses needed to filter the sales yeah.

> Should I return all the default post_statuses of WordPress for all cases?

see above

actuallyakash: There are too many fields that are using `.llms-select2-post` so I chose to use the 1st option (using data attribute) for this.

eri-trabiccolo (changes requested):
Looks good to me, I'd only take this occasion to escape the stuff we print:
1. `echo implode( ',', array_keys( get_post_statuses() ) );`
as `echo esc_attr( implode( ',', array_keys( get_post_statuses() ) ) );`.
2. escape all the urls with `esc_url`.
3. `esc_attr` for course and membership ids printed as values.
4. esc_html (esc_html__) for all the localized strings.
I'd say we can keep the title (`get_title(...)`) as it is without escaping its html.

Can you please take care of this? <3

actuallyakash:
done escaping.

.changelogs/issue_2490.yml
```diff
@@ -0,0 +1,5 @@
+significance: patch
+type: changed
+links:
+  - "#2490"
+entry: Added support for showing Private Courses in Reporting > Sales Page.
```
eri-trabiccolo: can you update this?
the change impacts memberships too and not only private status.

templates/admin/reporting/nav-filters.php
```diff
@@ -5,7 +5,7 @@
  * @package LifterLMS/Templates/Admin
  *
  * @since Unknown
- * @version 3.29.0
+ * @version [version]
```
eri-trabiccolo, line 9: can you add a small changelog here?
e.g.
```suggestion
 * @since [version] What I did.
 * @version [version]
```

templates/admin/reporting/nav-filters.php
```diff
@@ -76,29 +76,28 @@
...
 			<li class="llms-nav-item llms-analytics-form">
 
-				<label><?php _e( 'Courses', 'lifterlms' ); ?></label>
-
-				<select class="llms-select2-post" data-placeholder="<?php _e( 'Filter by Course(s)', 'lifterlms' ); ?>" data-post-type="course" id="llms-course-ids-filter" name="course_ids[]" multiple="multiple">
+				<label><?php esc_html_e( 'Courses', 'lifterlms' ); ?></label>
+				<select data-post-statuses="<?php echo esc_attr( implode( ',', array_keys( get_post_statuses() ) ) ); ?>" class="llms-select2-post" data-placeholder="<?php esc_html_e( 'Filter by Course(s)', 'lifterlms' ); ?>" data-post-type="course" id="llms-course-ids-filter" name="course_ids[]" multiple="multiple">
 					<?php foreach ( $current_courses as $course_id ) : ?>
-						<option value="<?php echo $course_id; ?>" selected><?php echo get_the_title( $course_id ); ?> <?php printf( __( '(ID# %d)', 'lifterlms' ), $course_id ); ?></option>
+						<option value="<?php echo esc_attr( $course_id ); ?>" selected><?php echo get_the_title( $course_id ); ?> <?php printf( __( '(ID# %d)', 'lifterlms' ), $course_id ); ?></option>
```
eri-trabiccolo: ``` printf( __( '(ID# %d)', 'lifterlms' ), $course_id );```
that translation needs to be escaped, also can you add `Translators` suggestions? E.g.
https://github.com/gocodebox/lifterlms/blob/94903b54585ebd572e533d25698fd9b1aa2cb347/templates/loop/enroll-date.php#L23

templates/admin/reporting/nav-filters.php
```diff
@@ -67,7 +67,7 @@
 					?>
 					<?php foreach ( $current_students as $id ) : ?>
 						<?php $s = get_user_by( 'id', $id ); ?>
-						<option value="<?php echo $id; ?>" selected="selected"><?php echo $s->display_name; ?> &lt;<?php echo $s->user_email; ?>&gt;</option>
+						<option value="<?php echo esc_attr( $id ); ?>" selected="selected"><?php echo $s->display_name; ?> &lt;<?php echo $s->user_email; ?>&gt;</option>
```
eri-trabiccolo: I would escape display name and email too.

templates/admin/reporting/nav-filters.php
```diff
@@ -17,33 +17,33 @@
...
+				<a class="llms-nav-link" data-range="last-7-days" href="<?php echo esc_url( admin_url( 'admin.php?page=llms-reporting&tab=' . $current_tab . '&range=last-7-days' ) ); ?>"><?php esc_html_e( 'Last 7 Days', 'lifterlms' ); ?></a>
 			</li>
 
 
 			<li class="llms-nav-item llms-analytics-form<?php echo ( 'custom' === $current_range ) ? ' llms-active' : ''; ?>">
 
-				<label><?php _e( 'Custom', 'lifterlms' ); ?></label>
+				<label><?php esc_html_e( 'Custom', 'lifterlms' ); ?></label>
 				<input type="text" name="date_start" class="llms-datepicker" placeholder="yyyy-mm-dd" value="<?php echo $date_start; ?>"> -
 				<input type="text" name="date_end" class="llms-datepicker" placeholder="yyyy-mm-dd" value="<?php echo $date_end; ?>">
```
eri-trabiccolo: let's escape the dates (`esc_attr`)

eri-trabiccolo (changes requested):
Looks good, can you just escaping more stuff?
This only to improve our past code you know... :D
Thanks <3

.changelogs/issue_2490.yml
actuallyakash: done

templates/admin/reporting/nav-filters.php
actuallyakash: done

templates/admin/reporting/nav-filters.php
actuallyakash: done.

templates/admin/reporting/nav-filters.php
actuallyakash: done

actuallyakash:
> Looks good, can you just escaping more stuff? This only to improve our past code you know... :D Thanks <3

it's done 👌

templates/admin/reporting/nav-filters.php
```diff
@@ -76,29 +77,44 @@
...
-
-				<select class="llms-select2-post" data-placeholder="<?php _e( 'Filter by Course(s)', 'lifterlms' ); ?>" data-post-type="course" id="llms-course-ids-filter" name="course_ids[]" multiple="multiple">
+				<label><?php esc_html_e( 'Courses', 'lifterlms' ); ?></label>
+				<select data-post-statuses="<?php echo esc_attr( implode( ',', array_keys( get_post_statuses() ) ) ); ?>" class="llms-select2-post" data-placeholder="<?php esc_html_e( 'Filter by Course(s)', 'lifterlms' ); ?>" data-post-type="course" id="llms-course-ids-filter" name="course_ids[]" multiple="multiple">
 					<?php foreach ( $current_courses as $course_id ) : ?>
-						<option value="<?php echo $course_id; ?>" selected><?php echo get_the_title( $course_id ); ?> <?php printf( __( '(ID# %d)', 'lifterlms' ), $course_id ); ?></option>
+						<option value="<?php echo esc_attr( $course_id ); ?>" selected><?php echo get_the_title( $course_id ); ?>
+							<?php 
+							printf(
+								// Translators: %s = Course ID.
```
eri-trabiccolo:
```suggestion
								// Translators: %d = Course ID.
```

templates/admin/reporting/nav-filters.php
```diff
@@ -76,29 +77,44 @@
...
+				<label><?php esc_html_e( 'Memberships', 'lifterlms' ); ?></label>
 
-				<select class="llms-select2-post" data-placeholder="<?php _e( 'Filter by Memberships(s)', 'lifterlms' ); ?>" data-post-type="llms_membership" id="llms-membership-ids-filter" name="membership_ids[]" multiple="multiple">
+				<select data-post-statuses="<?php echo esc_attr( implode( ',', array_keys( get_post_statuses() ) ) ); ?>" class="llms-select2-post" data-placeholder="<?php esc_html_e( 'Filter by Memberships(s)', 'lifterlms' ); ?>" data-post-type="llms_membership" id="llms-membership-ids-filter" name="membership_ids[]" multiple="multiple">
 					<?php foreach ( $current_memberships as $membership_id ) : ?>
-						<option value="<?php echo $membership_id; ?>" selected><?php echo get_the_title( $membership_id ); ?> <?php printf( __( '(ID# %d)', 'lifterlms' ), $membership_id ); ?></option>
+						<option value="<?php echo esc_attr( $membership_id ); ?>" selected><?php echo get_the_title( $membership_id ); ?>
+							<?php 
+							printf(
+								// Translators: %s = Membership ID.
```
eri-trabiccolo:
```suggestion
								// Translators: %d = Membership ID.
```

eri-trabiccolo (changes requested):

templates/admin/reporting/nav-filters.php
actuallyakash: oops, sorry. Fixed now. Thanks.

eri-trabiccolo (commented):
Hey @actuallyakash 
this is good - except for coding standards issues to fix.
Also, have you checked for all the possible post statuses then, even for scheduled posts?

eri-trabiccolo:
@actuallyakash bump

actuallyakash:
> Hey @actuallyakash this is good - except for coding standards issues to fix. Also, have you checked for all the possible post statuses then, even for scheduled posts?

Hey @eri-trabiccolo 

by "scheduled posts" did you mean "pending posts"?

eri-trabiccolo:
> > Hey @actuallyakash this is good - except for coding standards issues to fix. Also, have you checked for all the possible post statuses then, even for scheduled posts?
> 
> Hey @eri-trabiccolo
> 
> by "scheduled posts" did you mean "pending posts"?

hey
@actuallyakash 

I mean when you schedule a post to be published in the future:
screenshot-20230803_163447_909760316
screenshot-20230803_163515_417089140

actuallyakash:
@eri-trabiccolo 

sorry for the late response.

> I mean when you schedule a post to be published in the future:

No, the scheduled posts (courses/memberships) are not visible. We're currently using `get_post_statuses()` which only returns `draft,pending,private,publish`. Though, we can add a `future` tag by appending it to the array.

Should we show scheduled posts as old posts can be again rescheduled? Let me know your thoughts.

eri-trabiccolo:
@actuallyakash 
What I think we should do is to allow users to filter by all the (existing) courses/memberships that produced a sale that is visible in the graph/numbers.
So check this: buy a course, turn its status to "whatever status" (except for trash/deleted), if the sale appears in that reporting screen (its amount is counted, the graphs shows it) etc, then we should be able to filter by that course too.

eri-trabiccolo:
@actuallyakash let's not forget about this :D

actuallyakash:
> @actuallyakash let's not forget about this :D

Didn't forget :D 

I worked on it in between the tasks and still working on finding the solution. Will wrap it up soon.

actuallyakash:
hey @eri-trabiccolo 

I've added `future` post-status to `data-post-statuses` for showing scheduled posts (courses/memberships).

I don't think showing the sold posts is feasible at the moment as we are passing the `courses` and `memberships` post type [here](https://github.com/gocodebox/lifterlms/blob/trunk/includes/class.llms.ajax.handler.php#L946). We need to fetch the orders and then fetch the courses associated with them. Too long a process. Perhaps, there must be a shorter one which I'm not able to find without changing the logic in this [query](https://github.com/gocodebox/lifterlms/blob/trunk/includes/class.llms.ajax.handler.php#L941).

I think for now, adding `future` will be sufficient. Let me know your thoughts. 😁

ideadude:
Did we accidentally undo the change to include private posts in the second commit here: https://github.com/gocodebox/lifterlms/pull/2493/commits/c4778166d67429cd921806f987a9fd1f5af62514#diff-62976cfc458475b0fd1f70759a37960b64b21da658718ede3884e3897062b4a6L62

Let's double check that. Test. Review and merge.

brianhogg:
Private posts were being included by default as part of the PR, but reverted. No changes to what it was before in that file.

PR #2472, opened 2023-06-27
Add student dashboard section shortcodes

seothemes:
## Description

Adds new `lifterlms_dashboard` shortcode to allow users to customize the layout of the student dashboard sections.

#### Dashboard

Displays the student dashboard navigation. 

`[lifterlms_dashboard section="navigation" layout="vertical"]`

*Attributes:*

- section: {{section_name}}
- layout: vertical/horizontal
- show_title: true/false

## How has this been tested?

Locally

### Initial concept (subject to change)

dashboard section shortcodes

### Sample HTML block data

```html
<!-- wp:shortcode -->
[lifterlms_dashboard]
<!-- /wp:shortcode -->

<!-- wp:columns -->
<div class="wp-block-columns"><!-- wp:column {"width":"33.33%"} -->
<div class="wp-block-column" style="flex-basis:33.33%"><!-- wp:shortcode -->
[lifterlms_dashboard section="navigation"]
<!-- /wp:shortcode --></div>
<!-- /wp:column -->

<!-- wp:column {"width":"66.66%"} -->
<div class="wp-block-column" style="flex-basis:66.66%"><!-- wp:shortcode -->
[lifterlms_dashboard section="content" show_title="no"]
<!-- /wp:shortcode --></div>
<!-- /wp:column --></div>
<!-- /wp:columns -->

<!-- wp:shortcode -->
[/lifterlms_dashboard]
<!-- /wp:shortcode -->
```

## Types of changes

New feature (non-breaking change which adds functionality)

## Checklist:
- [ ] This PR requires and contains at least one changelog file.
- [ ] My code has been tested.
- [ ] My code passes all existing automated tests.
- [ ] My code follows the LifterLMS Coding & Documentation Standards.

PR #2531, opened 2023-10-05
Feature/dashboard shortcodes

eri-trabiccolo:
see #2472


brianhogg:
We're not looking to take the approach of splitting the dashboard into piece (navigation vs. content) at this time. We can reference this PR if we ever decide to take this approach.

PR #2569, opened 2023-11-08
Add filter to prevent current user with an "administrator" or lms_manager" role from being changed to an "instructor"

### LINKED ISSUE #1100: It should not be possible for an administrator to change their role to "Instructor"
### Reproduction Steps

+ Login to admin panel as an administrator
+ Visit Admin->Users
+ Select yourself using the checkbox
+ Select "Instructor" from the bulk role change dropdown on the top of the page
+ Click "Change" button

### Expected Behavior

+ The WP core prevents admins from changing their role to any role without the `edit_users` capability and this behavior would be expected trying to switch yourself to an instructor.
+ This expected can be observed by performing the same steps above but choosing "Editor" instead of "Instructor"

### Actual Behavior

+ The role is changed.

### More Information

Since instructors are able to edit users (assistant instructors) they *kind of* have this capability but only for specific users.

We should hook into this bulk edit role change and add additional checks when changing yourself from admin (or lms_manager) to instructor.

bsetiawan88:
## Description
Upon closer examination of the role change process in WordPress core, it becomes evident that the default WordPress check for changing a user's role from `administrator` to `editor` fails at https://github.com/WordPress/WordPress/blob/master/wp-admin/users.php#L148. It's crucial to recognize that the correct capability name for this check is `promote_users` rather than `edit_users`. Consequently, relying on the `handle_bulk_actions-users.php` filter to enforce this restriction is not a viable approach. To achieve the same outcome as observed in the WordPress core, the `role_has_cap` filter is utilized as an alternative. Within this filter, the necessary checks and adjustments are made, ensuring that `administrator` and `lms_manager` are prevented from changing their roles to `instructor`.

Fixes #1100

## How has this been tested?
Manually change current user role to `instructor`.

## Checklist:

### DIFF
--- a/includes/admin/class-llms-admin-users-table.php
+++ b/includes/admin/class-llms-admin-users-table.php
@@ -48,7 +48,7 @@ class LLMS_Admin_Users_Table {
 		add_filter( 'views_users', array( $this, 'modify_views' ) );
 
 		add_filter( 'user_row_actions', array( $this, 'add_actions' ), 20, 2 );
-
+		add_filter( 'role_has_cap', array( $this, 'role_has_cap' ), 10, 3 );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


ideadude:
Closed in favor of a core WP update that we will push for.

Some discussion here:
https://github.com/gocodebox/lifterlms/issues/1100#issuecomment-2035509215

Thanks for the contribution anyway @bsetiawan88

PR #2592, opened 2024-02-01
Added nonce for reviews and checking user limits.

ideadude:
## Description

The reviews form did not have a nonce. Adding one makes sure people are on the site when submitting a review.

Further, while the form was hidden if the one review per person setting was enabled and the user already submitted a review, the code that processed the AJAX calls did not also check this setting or the user limits.

Together, the above could allow someone to create unlimited reviews from offsite by posting to the admin-ajax.php with the correct values.

This PR adds the nonce and also abstracts the limit check into it's own method and then uses that to see if the form should be shown AND uses it again on the backend before creating the review posts.

## How has this been tested?

Make sure to build llms.js or the nonce check will not work.

This has been tested locally by creating a course and enabling reviews and the prevent multiple reviews setting. You can open the course page in multiple tabs before submitting a review. Then submit a review in the first tab. A review should show up in the dashboard. Then submit a review in the second tab. While it still says "Thanks" on the frontend, there will be no review in the backend.

I also tested using Postman and sending authenticated POST requests directly to admin-ajax to create reviews. They do not work without the nonce.

We could consider showing better error messages when the limit is hit, but I don't think this is important in these edge cases.

## Checklist:


includes/class.llms.review.php
```diff
@@ -212,23 +194,81 @@ public static function output() {
...
+		// Make sure the user is logged in.
+		if ( ! is_user_logged_in() ) {
+			return false;
+		}
+
+		// Make sure we have a post ID to check.
+		if ( empty( $parent_id ) ) {
+			$post_id = get_the_ID();
+		}
+		if ( empty( $parent_id ) ) {
```
brianhogg: Since this is the same check as above, couldn't we just return false there? Or should this line be checking for an empty `$post_id`?

includes/class.llms.review.php
```diff
@@ -212,23 +194,81 @@ public static function output() {
...
 	 */
 	public function process_review() {
+		// Check the nonce.
+		if ( ! isset( $_POST['llms_review_nonce'] ) || ! wp_verify_nonce( sanitize_key( $_POST['llms_review_nonce'] ), 'llms-review' ) ) {
+			return;
+		}
+
+		$parent_id = llms_filter_input_sanitize_string( INPUT_POST, 'pageID' );
+
+		// Make sure the curren user can write reviews yet.
```
brianhogg: Small typo (curren vs current)

includes/class.llms.review.php
```diff
@@ -212,23 +194,81 @@ public static function output() {
...
+		}
 
 		$post = array(
 			'post_content' => llms_filter_input_sanitize_string( INPUT_POST, 'review_text' ), // The full text of the post.
 			'post_name'    => llms_filter_input_sanitize_string( INPUT_POST, 'review_title' ), // The name (slug) for your post.
 			'post_title'   => llms_filter_input_sanitize_string( INPUT_POST, 'review_title' ), // The title of your post.
 			'post_status'  => 'publish',
 			'post_type'    => 'llms_review',
-			'post_parent'  => llms_filter_input_sanitize_string( INPUT_POST, 'pageID' ), // Sets the parent of the new post, if any. Default 0.
+			'post_parent'  => $parent_id, // Sets the parent of the new post, if any. Default 0.
```
brianhogg, line 219: The comment says we're defaulting to 0, but in `current_user_can_write_review` we're fetching `get_the_ID()`? Though we're not using it, possibly because of a typo (ie. `$parent_id = get_the_ID()` instead of `$post_id = get_the_ID()`

includes/class.llms.review.php
```diff
@@ -212,23 +194,81 @@ public static function output() {
...
+			'author'           => get_current_user_id(),
+			'suppress_filters' => true,
+		);
+		$posts_array = get_posts( $args );
+		if ( get_post_meta( $parent_id, '_llms_multiple_reviews_disabled', true ) && $posts_array ) {
+			return false;
+		}
+
+		// If we got here, we can write a review.
+		return true;
```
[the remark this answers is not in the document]
brianhogg: I disagree, why not do early returns :)

brianhogg (changes requested):

includes/class.llms.review.php
ideadude:
```suggestion
		// Make sure the current user can write reviews yet.
```

includes/class.llms.review.php
ideadude: Updated to require a parent_id. https://github.com/gocodebox/lifterlms/pull/2592/commits/88ccd31f58175b7f2e63d2da7ac58677d0c5160b

includes/class.llms.review.php
ideadude: Always have a parent_id now. https://github.com/gocodebox/lifterlms/pull/2592/commits/88ccd31f58175b7f2e63d2da7ac58677d0c5160b

PR #2623, opened 2024-04-11
Updating lifter tests version, fixing PHPCS errors

brianhogg:
## Types of changes
Updating tests and PHPCS checks.


brianhogg:
Will avoid PHPCS for the time being.

PR #2643, opened 2024-04-30
Make blocks unsynced by default

brianhogg:
## Description
Since a couple version of WP ago, reusable blocks are now Synced by default. Make them unsynced on new installs so they can be edited more easily.

Note this would only fix things on a fresh install of WordPress. To fix on an existing install we'd need some kind of migration script to change things like this in Forms:

```
<!-- wp:block {"ref":7} /-->
```

to this:

```
<!-- wp:llms/form-field-confirm-group {"llms_visibility":"logged_out"} -->
<!-- wp:llms/form-field-user-password {"description":"","field":"password","required":true,"label":"Password","label_show_empty":false,"match":"password_confirm","options":[],"options_preset":"","placeholder":"","columns":6,"last_column":false,"name":"password","id":"password","data_store":"users","data_store_key":"user_pass","html_attrs":{"minlength":8},"isConfirmationField":false,"isConfirmationControlField":true,"meter":true,"meter_description":"A strong password is required with at least 8 characters. To make it stronger, use both upper and lower case letters, numbers, and symbols.","min_strength":"strong","llms_visibility":"off"} /-->

<!-- wp:llms/form-field-text {"description":"","field":"password","required":true,"label":"Confirm Password","label_show_empty":false,"match":"password","options":[],"options_preset":"","placeholder":"","columns":6,"last_column":true,"name":"password_confirm","id":"password_confirm","data_store":"usermeta","data_store_key":"password_confirm","html_attrs":{"minlength":8},"isConfirmationField":true,"isConfirmationControlField":false,"llms_visibility":"off"} /-->
<!-- /wp:llms/form-field-confirm-group -->

```

while taking into account any changes that have been made to the reusable block.

This doesn't create the reusable blocks anymore since it doesn't run through `get_reusable_block`, but reusable blocks seem very buggy right now (ie. we can't uncheck "Confirmation field" for the Email Address block due to an error).

## How has this been tested?
Manually

## Checklist:


brianhogg:
The legacy blocks don't include things like the Confirmation email field on by default. Need to go through and compare.

brianhogg:
Workaround is to detach the group but would be nice if they were broken up by default

brianhogg:
No longer needed now that we fixed the "email required" error.

PR #2651, opened 2024-05-08
Added Block setting and Shortcode attribute to display Student Account Dashboard as stacked or columns

kimcoleman:
## Description
Now allowing people to define a shortcode attribute 'layout' for our Student Dashboard / My Account page shortcode OR for the My Account block.

The attribute accepts 'columns' or 'stacked'. Stacked will maintain the current appearance of the My Account page where the navigation is above the content of the tab.

Columns will display the layout with the navigation in the left column. We can provide people with CSS recipes to have the columns swapped so the nav is on the right side.  I feel this is better than supporting more options for the layout attribute, but I am open to it. It isn't a lot of effort to change this to "Navigation Location: Top | Left | Right".

This update also includes adding proper core LifterLMS forms CSS selectors to the "Redeem a Voucher" tab of the My Account page so that the form field and buttons are in line with other forms throughout the frontend.

## How has this been tested?
Locally

Note: We should test a bit more with Kadence and Astra - right now they are looking fine when the default attribute is set to stacked. Kadence will not look good with the attribute "columns" since our CSS is getting mixed up with the styling in their customizations
.
Sky Pilot will also need an update. I am going to do that PR and make it dependent on this update being released.

## Screenshots 
Screenshot 2024-05-08 at 7 26 30 AM
Screenshot 2024-05-08 at 7 34 47 AM

## Types of changes

## Checklist:

### DIFF
--- a/includes/class.llms.install.php
+++ b/includes/class.llms.install.php
@@ -286,7 +286,7 @@ class LLMS_Install {
 					'docs_url'     => 'https://lifterlms.com/docs/checkout-page/?utm_source=LifterLMS%20Plugin&utm_campaign=Plugin%20to%20Sale&utm_medium=Wizard&utm_content=LifterLMS%20Checkout%20Page',
 				),
 				array(
-					'content'      => '[lifterlms_my_account]',
+					'content'      => '[lifterlms_my_account layout="columns"]',
 					'option'       => 'lifterlms_myaccount_page_id',
 					'slug'         => 'dashboard',
 					'title'        => __( 'Dashboard', 'lifterlms' ),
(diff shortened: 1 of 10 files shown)


includes/class.llms.install.php
```diff
@@ -286,7 +286,7 @@ public static function get_pages() {
 					'docs_url'     => 'https://lifterlms.com/docs/checkout-page/?utm_source=LifterLMS%20Plugin&utm_campaign=Plugin%20to%20Sale&utm_medium=Wizard&utm_content=LifterLMS%20Checkout%20Page',
 				),
 				array(
-					'content'      => '[lifterlms_my_account]',
+					'content'      => '[lifterlms_my_account layout="columns"]',
```
brianhogg: The default is stacked, but the thought is to make it columns for new installs going forward?

kimcoleman: Yes - I have a conversation open on this in Slack. If we default to columns we break sites with themes (Kadence) that have custom styling for LifterLMS Student Dashboard. The admin would have to go in an explicitly set to "stacked" on an existing site. We can decide the best path in the Slack conversation.

kimcoleman: I've updated to default to columns. We can add to our CSS adjustments to Kadence when we release this.

kimcoleman:
Sky Pilot related PR that will need merge and release when this goes out (this will break Sky Pilot sites): https://github.com/gocodebox/sky-pilot/pull/115

PR #2660, opened 2024-05-15
v7.7

brianhogg:


brianhogg:
Ideally we'd like to avoid modifying the HTML/CSS for existing sites unless they opt in (ie. adding the `layout` option to the shortcode or selecting the new layout in the block). Currently this is switching to a left-side nav for default themes, and the navigation is hidden completely in Sky Pilot.

brianhogg:
To echo a note from Slack, I believe we'll need to coordinate with themes that need updates to work with the Dashboard changes as-is since there's HTML changes. I’m not sure what the status of that coordination is, and I haven’t done thorough testing to see what changes will be required with what themes.

PR #2677, opened 2024-06-14
Feature/account security

ideadude:
## Description

We are going to add some spam/security features:
[x] IP Block at checkout if there are more than 10 failures within 15 minutes.
[x] Built in support for captcha's like recaptcha and turnstile.
[x] Built in integration with Akismet to check emails at checkout/registration.

## How has this been tested?
Manually

## Checklist:


includes/llms.spam.functions.php
```diff
@@ -0,0 +1,201 @@
+<?php
+/**
+ * Code related to spam detection and prevention.
+ */
+// Constants. Define these in wp-config.php to override.
+if ( ! defined( 'LLMS_SPAM_ACTION_NUM_LIMIT' ) ) {
```
brianhogg, line 10: We're using `llms_maybe_define_constant` in other spots but I'm fine to just do this standard pattern, especially if it needs to be defined before the function might be loaded.

includes/llms.spam.functions.php
```diff
@@ -0,0 +1,201 @@
...
+	$ip = preg_replace( '/[^0-9a-fA-F:., ]/', '', $ip );
+	$transient_key = 'llms_spam_activity_' . $ip;
+	$activity = get_transient( $transient_key );
+	if ( empty( $activity ) || ! is_array( $activity ) ) {
+		$activity = [];
+	}
+
+	// Remove old items.
+	$new_activity = [];
+	$now = current_time( 'timestamp', true ); // UTC
```
brianhogg: `current_time( 'timestamp' )` isn't recommended since WP 5.3 https://make.wordpress.org/core/2019/09/23/date-time-improvements-wp-5-3/

includes/llms.spam.functions.php
```diff
@@ -0,0 +1,201 @@
...
+	foreach( $activity as $item ) {
+		// Determine whether this item is recent enough to include.
+		if ( $item > $now-( LLMS_SPAM_ACTION_TIME_LIMIT ) ) {
```
brianhogg: Wrap in `absint()` in case the constant is defined incorrectly

includes/llms.spam.functions.php
```diff
@@ -0,0 +1,201 @@
...
+
+	$activity = llms_get_spam_activity( $ip );
+	$now = current_time( 'timestamp', true ); // UTC
+	array_unshift( $activity, $now );
+
+	// If we have more than the limit, don't bother storing them.
+	if ( count( $activity ) > LLMS_SPAM_ACTION_NUM_LIMIT ) {
+		rsort( $activity );
+		$activity = array_slice( $activity, 0, LLMS_SPAM_ACTION_NUM_LIMIT );
+	}
```
brianhogg: Add `absint()` around the constants

includes/llms.spam.functions.php
```diff
@@ -0,0 +1,201 @@
...
+
+	// If we can't determine the IP, let's bail.
+	if ( empty( $ip ) ) {
+		return false;
+	}
+
+	$ip = preg_replace( '/[^0-9a-fA-F:., ]/', '', $ip );
+	$transient_key = 'llms_spam_activity_' . $ip;
+	$activity = get_transient( $transient_key );
```
brianhogg: Not sure if it's an issue using transients when they can be deleted at any time? It does allow us to use the WP system of tracking the temporary data without making our own table etc.

includes/llms.spam.functions.php
brianhogg: Leaving as-is, it's relative anyway.

includes/llms.spam.functions.php
brianhogg: It's a 15 min window so transients should do the trick most of the time.

PR #2679, opened 2024-06-18
Elementor support for Courses

brianhogg:
## Description

- Adds native Elementor support for courses. 
- Adds several basic Elementor widgets, which are added into the editor when first editing a course with Elementor after this update is in place.
- Adds new `[lifterlms_course_instructors]` shortcode and associated Elementor widget.
- Adds link in the native Elementor menu to Launch Course Builder.

## How has this been tested?
Manually

## Screenshots 
Captura de pantalla 2024-06-24 a las 2 49 05 p  m
Captura de pantalla 2024-06-24 a las 2 49 20 p  m

## Checklist:


kimcoleman:
Adding notes here that I have locally tested this and the DB upgrade worked fine.

includes/class-llms-elementor-migrate.php
```diff
@@ -0,0 +1,273 @@
...
+			'id'       => uniqid(),
+			'elType'   => 'container',
+			'settings' => array(),
+			'elements' => array(
+				array(
+					'id'         => uniqid(),
+					'elType'     => 'widget',
+					'settings'   => array(
+						'content_width' => 'full',
+						'html'          => '<h3>' . esc_attr__( 'Course Information', 'lifterlms' ) . '</h3>',
```
ideadude, line 50: There is a heading widget (included in Elementor free) we could use.

includes/class-llms-elementor-migrate.php
```diff
@@ -0,0 +1,273 @@
...
+			return;
+		}
+
+		$content = json_decode( $content, true );
+
+		if ( ! is_array( $content ) ) {
+			return;
+		}
+
+		$content = array_merge( $content, $this->get_elementor_data_template() );
```
ideadude: As far as I can tell we throw away the default (core WP) post_content.

There were some reasons for doing this. I believe one is that for the LifterLMS blocks in particular, you might get content like this, which includes the WP block editor HTML comment and some preview HTML that is kind of frozen in time and not kept up to date.

````
<!-- wp:llms/course-continue-button -->
<div class="wp-block-llms-course-continue-button" style="text-align: center;">[lifterlms_course_continue_button]</div>
<!-- /wp:llms/course-continue-button -->
````

I'm worried about 2 things in particular.

1. We also remove any content the user added to the course description (the_content) in addition to what LifterLMS puts in there, including some blocks/etc (like paragraph, headings, shortcode, video, and image blocks) that do convert decently to Elementor this way.

2. Users of Elementor will expect some amount of the old content to transfer over when they click Edit with Elementor. It works this way for some other course CPTs (PMPro Courses e.g.) and I feel users will see this as a deficiency in LifterLMS vs the subtle issue it is. And ultimately, users who have already edited course pages will have to go back and forth between the WP and Elementor versions of the post to copy/paste things, if they even know that they can do that.

So I think we need to get "clever" here and try our best to (a) preserve the_content that is there while also (b) removing or addressing the bits that are going to be confusing because they don't work as intended.

A simple solution would be to grab the_content and use REGEX or HTML parsing to remove anything between HTML comments like `<!-- wp:llms/.*<!-- /wp:llms/.*-->` (need the actual REGEX)

Thoughts?

File: includes/llms.template.functions.php [hunk 3]
```diff
@@ -261,19 +261,6 @@ function lifterlms_template_single_course_tracks() {
 	}
 }
 
-/**
```
ideadude, line 264: Why was this code removed? If it was an accident, add it back. If not, let's include an explanation in the PR description and something about it in the changelog.

ideadude (commented):
Small suggestion to use the heading widget for Elementor.

Question RE some extra code that was removed.

Bigger question RE trying to preserve the_content when first editing with Elementor.

includes/class-llms-elementor-migrate.php
brianhogg: I believe until you save the post in Elementor, the_content is preserved. I’d be curious to know what block content from PMPro CPTs is preserved when starting to edit with elementor.

We could look into adding a widget with the content from the_content as a widget at the top if there is no content in `_elementor_data`, which I believe is all that Elementor does when you first edit.

includes/class-llms-elementor-migrate.php
brianhogg: I did that at first, but the font, color and sizing is all different than the headers added by the lifter course information widgets on the page. By using an HTML widget it inherits from the theme instead. I couldn't find an "inherit" option in the Heading widget but maybe there is one...

includes/llms.template.functions.php
brianhogg: Ah, the file was edited at one point, and the CS fixer removed the extra line. Since there's no other tangible changes I'll just revert the change to the file to avoid confusion.

brianhogg: The function is in the same file twice (it's also on line 245 above). Thanks to the `function_exists()` check there was no error, but didn't need to be defined a second time. It still exists so no changelog needed :)

includes/class-llms-elementor-migrate.php
brianhogg: I'm not seeing a way to ignore color and font when using the Elementor Heading widget, so I think the HTML widget is our best bet to have the styles inherited.

ideadude: Thanks. I figured there was a reason you did it this way. Let's ship it this way for now.

includes/class-llms-elementor-migrate.php
brianhogg: I've changed it to use `post_content` as the default for `_elementor_data` rather than an empty array. It also uncovered a bug where it needed to use `wp_slash` before the call to `update_post_meta` to avoid the json becoming unparsable with text editor content such as this.

ideadude: Awesome. The latest version now preserves the_content into the top of the post when first editing in Elementor. This feels about the best we can do.

We could try to strip out some things, but the block editor HTML comments are just comments and don't show up. If folks see dupe content (from shortcodes) they should notice and can remove the sections they don't want.

We may get feedback from folks on this and can consider tweaks in the future. In the majority of cases, folks will be editing a brand new blank page in the editor of their choice and sticking with it.

PR #2689, opened 2024-07-15
Fix/plugin checks 3

ideadude:
More Plugin Check Plugin (PCP) security fixes.

Some of these might have been missed earlier or broken during the merger of other PRs. Some had to have phpcs:ignore type comments tweaked to please the scanners.

For example, sometimes having a `-- explanation` at the end of the comment seemed to break the scanner. Code in the PCP itself seems to use this format to explain the ignores.

```
// PHPCS ignore reason: Content is escaped above.
// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```

### DIFF
--- a/includes/admin/class.llms.admin.assets.php
+++ b/includes/admin/class.llms.admin.assets.php
@@ -412,9 +412,9 @@ class LLMS_Admin_Assets {
 		';
 
 		echo '<script type="text/javascript">window.LLMS = window.LLMS || {};</script>';
-		
+
 		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- get_js_strings returns json_encoded strings.
-		echo '<script type="text/javascript">window.LLMS.l10n = window.LLMS.l10n || {}; window.LLMS.l10n.strings = ' . LLMS_L10n::get_js_strings( true ) . ';</script>';
+		echo '<script type="text/javascript">window.LLMS.l10n = window.LLMS.l10n || {}; window.LLMS.l10n.strings = ' . wp_json_encode( LLMS_L10n::get_js_strings( false ) ) . ';</script>';
 
 		$forms = LLMS_Forms::instance()->get_post_type();
 
--- a/includes/admin/class.llms.admin.assets.php
+++ b/includes/admin/class.llms.admin.assets.php
@@ -412,6 +412,8 @@ class LLMS_Admin_Assets {
 		';
 
 		echo '<script type="text/javascript">window.LLMS = window.LLMS || {};</script>';
+		
+		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- get_js_strings returns json_encoded strings.
 		echo '<script type="text/javascript">window.LLMS.l10n = window.LLMS.l10n || {}; window.LLMS.l10n.strings = ' . LLMS_L10n::get_js_strings( true ) . ';</script>';
 
 		$forms = LLMS_Forms::instance()->get_post_type();
--- a/includes/admin/class.llms.admin.builder.php
+++ b/includes/admin/class.llms.admin.builder.php
@@ -608,7 +608,7 @@ class LLMS_Admin_Builder {
 					echo self::get_template(
 						$template,
 						array(
-							'course_id' => $course_id,
+							'course_id' => $course_id, // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 						)
 					);
 				}
--- a/includes/admin/class.llms.admin.builder.php
+++ b/includes/admin/class.llms.admin.builder.php
@@ -604,13 +604,14 @@ class LLMS_Admin_Builder {
 				);
 
 				foreach ( $templates as $template ) {
-					// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- Output is escaped in the template file.
+					// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
 					echo self::get_template(
 						$template,
 						array(
-							'course_id' => $course_id, // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+							'course_id' => $course_id,
 						)
 					);
+					// phpcs:enable
 				}
 
 				?>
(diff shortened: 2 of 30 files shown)


includes/admin/class.llms.admin.assets.php
```diff
@@ -412,6 +412,8 @@ public function admin_print_scripts() {
 		';
 
 		echo '<script type="text/javascript">window.LLMS = window.LLMS || {};</script>';
+		
+		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- get_js_strings returns json_encoded strings.
 		echo '<script type="text/javascript">window.LLMS.l10n = window.LLMS.l10n || {}; window.LLMS.l10n.strings = ' . LLMS_L10n::get_js_strings( true ) . ';</script>';
```
brianhogg: Could we not use `esc_js` here?

```suggestion
		echo '<script type="text/javascript">window.LLMS.l10n = window.LLMS.l10n || {}; window.LLMS.l10n.strings = \'' . esc_js( LLMS_L10n::get_js_strings( true ) ) . '\';</script>';

```

includes/admin/class.llms.admin.builder.php
```diff
@@ -608,7 +608,7 @@ public static function output() {
 					echo self::get_template(
 						$template,
 						array(
-							'course_id' => $course_id,
+							'course_id' => $course_id, // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 						)
 					);
 				}
```
brianhogg: I think this comment would need to be above the echo

templates/course/lesson-preview.php
```diff
@@ -22,8 +22,8 @@
 $data_msg     = $restrictions['is_restricted'] ? ' data-tooltip-msg="' . esc_html( strip_tags( llms_get_restriction_message( $restrictions ) ) ) . '"' : '';
 ?>
 
-<div class="llms-lesson-preview<?php echo $lesson->get_preview_classes(); ?>">
-	<a class="llms-lesson-link<?php echo $restrictions['is_restricted'] ? ' llms-lesson-link-locked' : ''; ?>" href="<?php echo ( ! $restrictions['is_restricted'] ) ? get_permalink( $lesson->get( 'id' ) ) : '#llms-lesson-locked'; ?>"<?php echo $restrictions['is_restricted'] ? ' data-tooltip-msg="' . esc_html( strip_tags( llms_get_restriction_message( $restrictions ) ) ) . '"' : ''; ?>>
+<div class="llms-lesson-preview<?php echo esc_attr( $lesson->get_preview_classes() ); ?>">
+	<a class="llms-lesson-link<?php echo $restrictions['is_restricted'] ? ' llms-lesson-link-locked' : ''; ?>" href="<?php echo ( ! $restrictions['is_restricted'] ) ? esc_url( get_permalink( $lesson->get( 'id' ) ) ) : '#llms-lesson-locked'; ?>"<?php echo $restrictions['is_restricted'] ? ' data-tooltip-msg="' . esc_html( strip_tags( llms_get_restriction_message( $restrictions ) ) ) . '"' : ''; ?>>
```
brianhogg: I think it would be `esc_attr` here since it's an attribute to the tag. The `strip_tags` might be redundant.

```suggestion
	<a class="llms-lesson-link<?php echo $restrictions['is_restricted'] ? ' llms-lesson-link-locked' : ''; ?>" href="<?php echo ( ! $restrictions['is_restricted'] ) ? esc_url( get_permalink( $lesson->get( 'id' ) ) ) : '#llms-lesson-locked'; ?>"<?php echo $restrictions['is_restricted'] ? ' data-tooltip-msg="' . esc_attr( strip_tags( llms_get_restriction_message( $restrictions ) ) ) . '"' : ''; ?>>
```

templates/myaccount/form-redeem-voucher.php
```diff
@@ -25,7 +25,7 @@
 		<input id="llms-voucher-code" type="text" placeholder="<?php esc_html_e( 'Voucher Code', 'lifterlms' ); ?>" name="llms_voucher_code" required="required">
 	</p>
 
-	<button id="llms-redeem-voucher-submit" type="submit"><?php esc_html( _ex( 'Submit', 'Voucher Code', 'lifterlms' ) ); ?></button>
+	<button id="llms-redeem-voucher-submit" type="submit"><?php esc_html_ex( 'Submit', 'Voucher Code', 'lifterlms' ); ?></button>
```
brianhogg: `esc_html_ex()` doesn't exist

brianhogg (changes requested):

includes/admin/class.llms.admin.assets.php
ideadude: LLMS_L10n::get_js_strings returns an array of strings that are already escaped.

includes/admin/class.llms.admin.builder.php
ideadude: I agree. It reads correctly that way. For some reason the PCP was complaining unless the comment was on that line. I think we should refactor slightly so the code reads better and appeases the scanner.

templates/myaccount/form-redeem-voucher.php
ideadude:
```suggestion
	<button id="llms-redeem-voucher-submit" type="submit"><?php echo esc_html_x( 'Submit', 'Voucher Code', 'lifterlms' ); ?></button>
```

includes/admin/class.llms.admin.assets.php
brianhogg: It's not already escaped in this case and we're using `json_encode` vs `wp_json_encode` so it could have output we don't want.

brianhogg: So you can't use `esc_js()` on either the result of `json_encode()` or `wp_json_encode()` as it'll turn the double quotes into `&quot;`. The docs aren't too clear on that.

What we can do instead is remove the `true` to return json, and do the `wp_json_encode` here instead. That should play nice with the scanner.

brianhogg: Verified that `wp_json_encode`-ing here works and scanner doesn't complain.

includes/admin/class.llms.admin.builder.php
brianhogg: Added multi-line ignore to make the scanner happy. We could have probably one-lined the method as well I believe.

brianhogg (approved):

PR #2750, opened 2024-08-27
Ensure audio/video urls with specials chars can be embedded

### LINKED ISSUE #2749: Valid video URL for lesson featured video does not display
### Reproduction Steps

1. Install and activate LifterLMS
2. Upload any MP4 video to the media library. Make sure that that video is named a①b.mp4 before the upload.
3. Create a brand-new course.
4. Publish two brand-new lessons named Lesson 1 and Lesson 2 on that course.
5. In the media library, copy the link to a①b.mp4 and then paste that to the Video Embed URL field of Lesson 1. You should have .../a①b.mp4 in the Video Embed URL field of Lesson 1. Save the changes.
6. In the media library, copy the link to a①b.mp4, and then paste that to the URL bar of your browser. Copy the link in the URL bar and the paste that to the Video Embed URL field of Lesson 2. You should have .../a%E2%91%A0b.mp4 in the Video Embed URL field of Lesson 2. Save the changes.
7. View Lesson 1 and Lesson 2 on the front end.

### Expected Behavior
Both videos in Lesson 1 and Lesson 2 should display on the front-end.

### Actual Behavior
The video in Lesson 1 does no display on the front end. But the video on Lesson 2 displays on the front-end.

brianhogg:
## Description

Fixes #2749

## How has this been tested?
Manually

## Checklist:


ideadude:
The changes here remove the sanitizing of the url. Is parse_url an acceptable way to sanitize the URL?

The new code wouldn't trigger a warning in the Plugin Check Plugin because it's using a get method to get the url instead of $_GET directly.

brianhogg:
@ideadude When I was looking into why `filter_var` was failing for valid URLs like those with special chars, it appears `filter_var` is using an outdated method of validation:

https://www.php.net/manual/en/function.filter-var.php#128462

I added the `parse_url` just to catch "seriously malformed urls" and instead relying on `wp_oembed_get` which is validating via `wp_safe_remote_get` when the embed is discovered and created. Adding an `wp_http_validate_url` was adding an extra network call plus failing for certain urls which were valid and should return the url wrapped in an iframe.

I did find a method to encode the url and then validate which could provide more validation before handing it off to `wp_oembed_get` if we'd like use it:

```
    $path = parse_url( $url, PHP_URL_PATH );
    $encoded_path = array_map( 'urlencode', explode('/', $path ) );
    $url = str_replace( $path, implode( '/', $encoded_path ), $url );

    return filter_var( $url, FILTER_VALIDATE_URL ) ? true : false;
```

However given `FILTER_VALIDATE_URL` is not using the up to date spec and the validation is occuring in `wp_oembed_get` I think we're good as-is.

ideadude:
I think even a sanitize_text_field would protect from the really bad stuff without breaking valid URLs.

I also think we're good as-is, but the plugin repo guidelines are pretty specific around sanitizing as early as possible and escaping as late as possible to prevent issues from future updates.

Issue #2807, opened 2024-11-18
Warning loading translations in WP 6.7


brianhogg:
### Reproduction Steps

1. With LifterLMS and WP 6.7 installed, switch the site language to something other than English (ie. Español)

### Expected Behavior

Translations are loaded and there are no warnings.

### Actual Behavior

A warning is displayed and the translations do not load.

![CleanShot 2024-11-18 at 9  10 48@2x](https://github.com/user-attachments/assets/dbd9f6da-8538-415e-940d-ed6d146c0f4b)

### Error Messages / Logs

```
Notice: Function _load_textdomain_just_in_time was called incorrectly. Translation loading for the lifterlms domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the init action or later. Please see [Debugging in WordPress](https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/) for more information. (This message was added in version 6.7.0.) in /Users/brianhogg/laravel/lifter2/wp-includes/functions.php on line 6114

```

This issue has be recreated:
+ [ ] Locally
+ [ ] On a staging site
+ [ ] On a production website
+ [ ] With only LifterLMS and a default theme

PR #2840, opened 2024-12-20
Fix translation unloading

brianhogg:
## Description
It was mentioned on a related trac ticket that we should not be unloading the textdomain on `init`. This is causing related changes in WP 6.7 to not show the translations if anything uses the `lifterlms` textdomain early.

I believe the `unload_textdomain` was added to fix a conflict with Loco Translate and saving in a certain location. I'm not able to replicate this, so it may no longer be needed. Restricting to just `lifterlms` core for now but we could just remove it for all add-ons too.

Refs #2807

## How has this been tested?
Manually

## Checklist:

### DIFF
--- a/includes/functions/llms-functions-l10n.php
+++ b/includes/functions/llms-functions-l10n.php
@@ -33,7 +33,6 @@ function llms_get_locale( $domain = 'lifterlms' ) {
 	 * @param string $domain The textdomain.
 	 */
 	return apply_filters( 'plugin_locale', $locale, $domain );
-
 }
 
 function llms_l10n_get_safe_directory() {
@@ -52,7 +51,6 @@ function llms_l10n_get_safe_directory() {
 	 * @param string $path Full server path to the safe directory.
 	 */
 	return apply_filters( 'llms_l10n_safe_directory', WP_LANG_DIR . '/lifterlms' );
-
 }
 
 /**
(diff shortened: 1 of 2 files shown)


brianhogg:
Setting a different language for the User still works.

There's more information on translations with Loco translate and custom file locations. This should be fine without unloading the textdomain. https://localise.biz/wordpress/plugin/faqs/custom-folder

includes/functions/llms-functions-l10n.php
```diff
@@ -91,7 +89,9 @@ function llms_load_textdomain( $domain, $plugin_dir = null, $language_dir = null
 	$plugin_dir   = $plugin_dir ? $plugin_dir : LLMS_PLUGIN_DIR;
 	$language_dir = $language_dir ? $language_dir : 'languages';
 
-	unload_textdomain( $domain );
+	if ( 'lifterlms' !== $domain ) {
```
brianhogg: We may want to just remove this completely even for add-ons?

ideadude: Yes. Let's remove it completely. Then test and if we don't have the issue in the latest versions of WP, Loco, and Lifter, merge it.

brianhogg:
Remove completely and do a test with an add-on + loco translate

PR #2859, opened 2025-01-23
Sales reporting fixes and improvements

### LINKED ISSUE #2858: Sales report does not include events after 23:23:59 on the last day of a date range
### Reproduction Steps

1. Place an order at 23:30 (or anytime after 23:23:59)
2. Under _LifterLMS > Reporting_, "Sales" tab, enter a from date in the past, and to date of today

### Expected Behavior

Things like "Net Sales" and "# of sales" includes the transaction

### Actual Behavior

The transaction is not included.

### Tech notes

We're setting the date to "23:23:59" instead of "23:59:59" in our query https://github.com/gocodebox/lifterlms/blob/7dc84b53fb2ef15185c42ff3ffc3123fc91561f5/includes/abstracts/abstract.llms.analytics.widget.php#L162-L180

We likely want to query for less than `00:00:00` on the day after [as per this thread](https://stackoverflow.com/questions/14104304/mysql-select-where-datetime-matches-day-and-not-necessarily-time).

brianhogg:
## Description

- Fix for date range not including orders/transactions after 23:23:59 on the last day of the date range
- Updating the date range code on other widgets to use the recommended/safe method of fetching data between two dates
- Adds new "# of Transactions" widget
- "Net Revenue" (renamed from "Net Sales") is now the calculation of the transaction amount minus any amount refunded, regardless of the status of the associated order
- "# of new sales" (renamed from "# of sales") now includes Orders with status “On Hold” “Expired” “Pending Cancellation” and “Cancelled”
- "Amount Refunded" now includes refunded amounts regardless of the status of the order, and is based on the last modified date of a transaction.

### Tech Notes

When a transaction is refunded in full or partially, it updates the original transaction with new postmeta _llms_refund_amount. The chart for net sales is using post_modified rather than post_date so, if a transaction is edited or partially refunded, that “net sale” (which is currently just the original transaction amount) shows on the date of the last refund. As such, Net Sales now uses the original post_date of the date the sale transaction happened. Unless we take the serialized data from _llms_refund_data and create new refund transactions with them, this is the best we can do in a query. Previously, "net sales" was just gross sales without refunds subtracted.

"Amount refunded" still uses the modified date, and will be the sum of all partial refunds on a transaction as of the date the last refund happened.

Fixes #2858, #2860, #2861

## How has this been tested?
Manually

CleanShot 2025-04-02 at 11  03 46@2x

## Checklist:


includes/admin/reporting/widgets/class.llms.analytics.widget.sold.php
```diff
@@ -112,14 +107,14 @@ public function set_query() {
 						{$txn_meta_join}
 						JOIN {$wpdb->postmeta} AS sales ON sales.post_id = txns.ID
 						WHERE
-						        ( txns.post_status = 'llms-txn-succeeded' )
+						        ( txns.post_status = 'llms-txn-succeeded' OR txns.post_status = 'llms-txn-refunded' )
```
brianhogg, line 110: This will incorrectly include the full amount of the original transaction in the Net Sales amount, rather than the difference between the original amount and refunded.

brianhogg: We've changed the SELECT portion to get the sum of the amount minus any refunded amounts.

includes/admin/reporting/tabs/class.llms.admin.reporting.tab.sales.php
```diff
@@ -68,7 +66,7 @@ public function get_widget_data() {
 						'title'   => __( '# of Sales', 'lifterlms' ),
 						'cols'    => '1-4',
 						'content' => __( 'loading...', 'lifterlms' ),
-						'info'    => __( 'Number of new active or completed orders placed within this period', 'lifterlms' ),
+						'info'    => __( 'Number of new non-refunded orders placed within this period', 'lifterlms' ),
```
brianhogg, line 69: Possibly sanitize translation string

brianhogg:
Investigate using JOIN instead of a subquery for the SUM portion, and maybe using another JOIN or a sub-query for the "IN" of what order IDs are included.

includes/admin/reporting/tabs/class.llms.admin.reporting.tab.sales.php
brianhogg: Escaped when it's output in admin/reporting/tabs/widgets.php

includes/admin/reporting/widgets/class.llms.analytics.widget.sold.php
```diff
@@ -102,41 +97,49 @@ public function set_query() {
 			$this->format_date( $dates['end'], 'end' ),
 		);
 
-		$this->query_function = 'get_results';
+		$this->query_function = 'get_var';
 		$this->output_type    = OBJECT;
 
-		$this->query = "SELECT
-							  txns.post_modified AS date
```
brianhogg, line 109: We need to group this by date otherwise we don't get a chart.

brianhogg:
When a transaction is refunded in full or partially, it updates the original transaction with new postmeta `_llms_refund_amount`. The chart for net sales is using post_modified rather than post_date so, if a transaction is edited or partially refunded, that “net sale” (which is currently just the original transaction amount) shows on the date of the last refund. As such, Net Sales now uses the original `post_date` of the date the sale transaction happened. Unless we take the serialized data from `_llms_refund_data` and create new refund transactions with them, this is the best we can do in a query. Previously, "net sales" was just gross sales without refunds subtracted.

"Amount refunded" still uses the modified date, and will be the sum of all partial refunds on a transaction as of the date the last refund happened.

PR #2864, opened 2025-01-27
fix order of quiz prev and next question buttons

kamzil:
## Description
Fix order of quiz prev/next buttons. After this fix, "next" button will properly be placed after the "prev" button.

## How has this been tested?
By installing the plugin and opening a quiz

## Checklist:

### DIFF
--- a/assets/js/llms-quiz.js
+++ b/assets/js/llms-quiz.js
@@ -774,9 +774,9 @@
 				$header = $( '<header class="llms-quiz-header" id="llms-quiz-header" />' )
 				$footer = $( '<footer class="llms-quiz-nav" id="llms-quiz-nav" />' );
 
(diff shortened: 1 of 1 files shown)


brianhogg:
@kamzil Thanks for the PR! We wouldn't merge this in as-is for everyone, as many sites expect it to be in the current order. We have a [CSS-based solution for swapping the buttons](https://lifterlms.com/docs/how-to-switch-the-previous-and-next-question-quiz-buttons/) and we're investigating an option to allow the order to be swapped if someone wants to.

kamzil:
Thanks for the link, but for some reason the suggested CSS doesn't have an effect on our site with Sky Pilot theme, at least on Firefox on Mac or Edge on Windows. I hope we can find another solution soon.

brianhogg:
Closing as we would not be switching the order of the buttons for all users. We would look into adding an option to do so if the CSS solution doesn't work.

PR #2884, opened 2025-02-25
Change display format of datepickers to match the date format of the site

### LINKED ISSUE #2447: Respect the date format saved in WordPress Dashboard > Settings > General  >  Date Format
### Reproduction Steps
1. Set the **date format** to `d/m/Y` in `WordPress Dashboard > Settings > General  >  Date Format`.
2. Publish a new LifterLMS course.
3. On the back-end of that course, turn on `Course Options > Restrictions > Enable Enrollment Period`.
4. Then, set the date for `Course Options > Restrictions > Enrollment Start Date`.
5. You'll notice that it still follows that `m/d/Y` date format.

### Expected Behavior
It must follow that format you set in `WordPress Dashboard > Settings > General  >  Date Format`.

### Actual Behavior
It does not follow the date format you set in `WordPress Dashboard > Settings > General  >  Date Format`. It sticks to the `m/d/Y` date format.

+ [x] Locally
+ [ ] On a staging site
+ [ ] On a production website

### Affected Tickets
HS-214507

brianhogg:
## Description

- [ ] Need to handle when the alt field is cleared, to also clear the hidden input value

POC for having a datepicker that shows the format of the site, but saves using a hidden input in the previous format.

Fixes #2447

## How has this been tested?
Manually

## Checklist:

### DIFF
--- a/.changelogs/datepicker-display-format-site-format.yml
+++ b/.changelogs/datepicker-display-format-site-format.yml
@@ -1,0 +1,6 @@
+significance: patch
+type: changed
+links:
+  - "#2447"
+entry: Display date pickers in the site date format, but save the data in the
+  previous format.
--- a/assets/js/private/llms-metaboxes.js
+++ b/assets/js/private/llms-metaboxes.js
@@ -279,6 +279,15 @@
 				minDate: minDate,
 				altField: altField,
 				altFormat: altFormat
+			} ).on( "keyup", function() {
+				var date;
+				try {
+					date = $.datepicker.parseDate( $.datepicker._defaults.dateFormat, this.value );
+				} catch ( e ) { }
+
+				if ( !date && altField ) {
+					$( altField ).val( "" );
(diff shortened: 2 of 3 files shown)


brianhogg:
Want to handle clearing the date and still saving the blank value.

ideadude:
An option to consider is using the datetime-local input type. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local

This would localize the display of the datetime based on the user's browser's locale and preferences. I'm not sure if it's better to use the WP site's setting vs the browser's setting. There doesn't seem to be a way to override the locale without using JavaScript and in that case it might be better to use a jQuery like datepicker.

brianhogg:
Ended up keeping the jQuery datepicker "alt field" functionality, but handling the case where the value is cleared. https://jsfiddle.net/tj_vantoll/au9v5/

PR #3068, opened 2025-12-12
Divi

brianhogg:
## Description
Initial core Divi 4 support with LifterLMS modules

`yarn build` in the `includes/divi` folder rebuilds the styling and JS files.

For todos see the post mortem. It would be good if the 'disable default lifterlms content' showed the correct toggle value after refresh, and only appeared on the Course Settings (or LLMS post types) rather than everywhere. It should be possible to restrict since there are filters for Woocommerce post types to rename things like the Excerpt to Short Description.

## How has this been tested?
Manually

## Screenshots 
CleanShot 2025-12-12 at 14 35 45@2x
CleanShot 2025-12-12 at 14 36 10@2x

## Checklist:


includes/divi/extension/includes/fields/Input/Input.jsx
```diff
@@ -0,0 +1,35 @@
+// External Dependencies
```
brianhogg, line 1: Can likely delete this

PR #3109, opened 2026-03-24
Feature: "Any" option for engagement trigger posts

flintfromthebasement:
## Summary

- Engagements can now fire for **any** course, membership, lesson, quiz, section, access plan, or course track — not just a specific one
- Addresses the feature request where N engagements were needed for N courses when the same email/certificate/achievement should fire for all of them
- Works for all trigger types: enrollment, purchase, completion, quiz pass/fail, etc.
- Covers courses **and** memberships (and all other post types) in one change

## How It Works

**Admin UI:** The trigger post select fields (e.g., "Select a Course") can now be cleared. When cleared, the placeholder reads "Any Course" (or "Any Membership", etc.), and a help text says "Leave blank to apply to all."

**Storage:** An empty selection saves the literal string `'any'` in `_llms_engagement_trigger_post`. This is an explicit opt-in — we intentionally avoided treating empty/zero values as wildcards so that existing engagements with accidentally blank trigger posts don't suddenly fire on everything.

**Query:** The engagement matching SQL adds `OR relation_meta.meta_value = 'any'` alongside the specific post ID match, so both targeted and wildcard engagements are returned.

**List table:** The engagements admin list shows "Any" in italics for wildcard engagements.

## Design Decision: Explicit `'any'` vs Empty-Means-Any

We considered treating empty or `0` values as "any" but rejected it because:
- Existing engagements with accidentally empty trigger posts would silently start firing on everything
- An explicit sentinel value (`'any'`) makes the intent clear in the database
- It's safer for sites upgrading — no surprise behavior changes

## Files Changed

| File | Change |
|------|--------|
| `class.llms.engagements.php` | SQL WHERE clause matches both specific post IDs and `'any'` |
| `class.llms.meta.box.engagement.php` | Placeholder text, value loading, save logic (empty → `'any'`), description text |
| `class.llms.admin.post.table.engagements.php` | Shows "Any" label for wildcard engagements |

## Duplicate/Spam Safety

- **Achievements & certificates:** The existing `dupcheck()` uses the *actual* related post ID (the course the student enrolled in), not the engagement's trigger post config. So an "any course" achievement is awarded once per course — not re-awarded for the same course.
- **Emails:** No dupcheck (by design) — they fire on every matching event, which is correct for welcome/notification emails.

## Test Plan

- [ ] Create an engagement with trigger "Student enrolls in a course", leave course blank (shows "Any Course")
- [ ] Enroll a student in Course A → engagement fires
- [ ] Enroll same student in Course B → engagement fires again
- [ ] Create a second engagement for a *specific* Course A
- [ ] Enroll in Course A → both engagements fire
- [ ] Enroll in Course B → only the "any" engagement fires
- [ ] Verify "Any Membership" works the same way for membership triggers
- [ ] Verify the engagements list table shows "Any" for wildcard engagements
- [ ] Verify existing specific engagements are unaffected
- [ ] Verify achievements with "any" trigger only award once per course (dupcheck)

### DIFF
--- a/.changelogs/feature_any-engagement-trigger.yml
+++ b/.changelogs/feature_any-engagement-trigger.yml
@@ -1,0 +1,7 @@
+significance: minor
+type: added
+links:
+  - "#3109"
+entry: Added an explicit 'Any' trigger option for engagements so one engagement
+  can apply to all matching courses, memberships, lessons, quizzes, sections,
+  access plans, or tracks.
--- a/includes/admin/post-types/meta-boxes/class.llms.meta.box.engagement.php
+++ b/includes/admin/post-types/meta-boxes/class.llms.meta.box.engagement.php
@@ -71,12 +71,14 @@ class LLMS_Meta_Box_Engagement extends LLMS_Admin_Metabox {
 				),
 				'id'               => '_faux_engagement_trigger_post_course',
 				'label'            => __( 'Select a Course', 'lifterlms' ),
+				'placeholder'      => __( 'Any Course', 'lifterlms' ),
 			),
 
 			'lesson'           => array(
 				'controller_value' => array( 'lesson_completed' ),
 				'id'               => '_faux_engagement_trigger_post_lesson',
 				'label'            => __( 'Select a Lesson', 'lifterlms' ),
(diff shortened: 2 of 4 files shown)


brianhogg:
@cursor review

PR #3168, opened 2026-06-03
Fix: wrap add_query_arg() with esc_url() on admin notice href

thisismyurl:
Hey folks,

The checkout settings link in the no-payment-gateways admin notice builds its URL with add_query_arg() but concatenates the result directly into '<a href="' . add_query_arg(...) . '">' without esc_url(). URLs placed in HTML attribute context must go through esc_url() — it validates the scheme and encodes unsafe characters.

One-line fix in class.llms.admin.notices.core.php.

(full disclosure, I used AI help with the testing and tweaks - Christopher)


brianhogg:
Duplicate

PR #3182, opened 2026-06-15
fix(media): use correct media type labels for block protection toolbar

### LINKED ISSUE #3165: Protecting a file says "Protect Image" incorrectly
### Reproduction Steps

1. Create a new WP page, and use the File block to upload a document
2. Click the lock icon to protect

### Expected Behavior

Will prompt to protect the file

### Actual Behavior

Says "protect image" in the modal and also when hovering over the lock icon

Image

Image

faisalahammad:
## Summary

The media protection toolbar button and modal always said "Protect Image" even when the selected block was a file, video, or audio block. This change makes the label match the actual media type.

Fixes #3165

### `src/js/admin-media-protection-block-protect.js`

**Before:**
```js
<ToolbarButton
    icon="lock"
    label="Protect Image"
    onClick={ () => setModalOpen( true ) }
/>
```

**After:**
```js
const mediaTypeLabel = getMediaTypeLabel( props.name );

<ToolbarButton
    icon="lock"
    label={ LLMS.l10n.replace( 'Protect %s', { '%s': mediaTypeLabel } ) }
    onClick={ () => setModalOpen( true ) }
/>
```

**Why:** `core/image`, `core/video`, `core/audio`, and `core/file` are all supported, but the labels were hardcoded for images. A new `getMediaTypeLabel()` helper returns the correct noun for each block type.

### `includes/admin/class.llms.admin.assets.php`

**Before:**
Core did not provide the translatable strings used by the new dynamic labels, so the toolbar would fall back to the literal English strings.

**After:**
Core now adds the needed strings via the existing `lifterlms_js_l10n_admin` filter:

```php
public function add_media_protection_js_l10n( $strings ) {
    return array_merge(
        $strings,
        array(
            'Protect %s' => __( 'Protect %s', 'lifterlms' ),
            'Select a Course or Membership to protect this %s:' => __( 'Select a Course or Membership to protect this %s:', 'lifterlms' ),
            'image' => __( 'image', 'lifterlms' ),
            'audio' => __( 'audio', 'lifterlms' ),
            'video' => __( 'video', 'lifterlms' ),
            'file' => __( 'file', 'lifterlms' ),
        )
    );
}
```

**Why:** This keeps the change inside the existing `LLMS.l10n` system instead of introducing a new `wp-i18n` dependency for one script.

## Testing

**Test 1: File block label**
1. Create a page and add a File block with an uploaded file.
2. Click the lock icon in the toolbar.

Result: toolbar tooltip and modal button say "Protect File", and the select label says "Select a Course or Membership to protect this file:".

**Test 2: Video and Audio blocks**
1. Add a Video block and an Audio block.
2. Click the lock icon for each.

Result: labels say "Protect Video" and "Protect Audio" respectively.

**Test 3: Image block remains unchanged**
1. Add an Image block.
2. Click the lock icon.

Result: labels still say "Protect Image".

## Types of changes

Bug fix (non-breaking change which fixes an issue).

## Checklist


brianhogg:
@faisalahammad Needs an update with dev. Ideally giving me write permissions would be ideal so I can update it myself if needed before merge. Thanks!

includes/admin/class.llms.admin.assets.php
```diff
@@ -490,6 +491,28 @@ protected function get_analytics_options() {
 		return apply_filters( 'llms_get_analytics_js_options', compact( 'currency_format' ) );
 	}
 
+	/**
+	 * Add translatable strings used by the media protection block editor toolbar.
+	 *
+	 * @since 10.0.6
```
brianhogg: Incorrect hard-coded value. [version] instead.

brianhogg (changes requested):

includes/admin/class.llms.admin.assets.php
faisalahammad: Changed @since 10.0.6 to @since [version]. Also merged dev into branch.

faisalahammad:
Addressed both review items:

- Fixed @since 10.0.6 -> @since [version] in includes/admin/class.llms.admin.assets.php
- Merged dev branch into this branch (no conflicts)

Ready for re-review. Please re-request review when you get a chance. Thanks!

includes/admin/class.llms.admin.assets.php
```diff
@@ -490,6 +491,28 @@ protected function get_analytics_options() {
...
 
+	/**
+	 * Add translatable strings used by the media protection block editor toolbar.
+	 *
+	 * @since [version]
+	 *
+	 * @param array $strings Array of JS l10n strings.
+	 * @return array
+	 */
+	public function add_media_protection_js_l10n( $strings ) {
```
brianhogg: We don't need to separately add the strings for the js when using LLMS.l10n.translate or .replace, they are picked up in the build.

includes/admin/class.llms.admin.assets.php
```diff
@@ -35,6 +35,7 @@ public function __construct() {
 		add_action( 'admin_print_footer_scripts', array( $this, 'admin_print_footer_scripts' ) );
 		add_action( 'enqueue_block_editor_assets', array( $this, 'block_editor_assets' ) );
 		add_action( 'elementor/editor/before_enqueue_scripts', array( $this, 'elementor_editor_assets' ) );
+		add_filter( 'lifterlms_js_l10n_admin', array( $this, 'add_media_protection_js_l10n' ) );
```
brianhogg: This needs to be removed as it is not needed.

brianhogg (changes requested):

includes/admin/class.llms.admin.assets.php
faisalahammad: Fixed. PHP l10n filter and add_media_protection_js_l10n method both removed. Strings auto-extracted by gulp pot-js, same pattern as lesson-timer.js.

includes/admin/class.llms.admin.assets.php
faisalahammad: Removed. The add_media_protection_js_l10n method and its filter registration are both deleted from LLMS_Admin_Assets.

brianhogg:
Merged in to test with build of 10.2

PR #3184, opened 2026-06-15
fix(integrations): redesign integrations list table

### LINKED ISSUE #3150: Remove "Integration ID" column in the Integrations list
This column isn't needed.

Make it a "Description" column with the one-sentence description from the product page, and a "Learn More" link afterwards to the getting started doc.

Rename "Installed" to "Configured" as it will only be checked if the associated plugin is installed, or the required API key(s) has/have been added.

faisalahammad:
## Summary

Replaces the `Integration ID` column in the LifterLMS settings Integrations table with a `Description` column (short blurb + `Learn More` link) and adds a new `Configured` column that indicates when an integration's required credentials are present.

Fixes #3150

### `includes/abstracts/llms.abstract.integration.php`

Added three new public extension points so add-ons can declare their own configuration state and docs link without changing signatures.

```php
public $learn_more_url = '';
public $configured_help = '';
public function is_configured() { return true; }
```

**Why:** add-ons need a way to tell the table whether keys have been provided, and where to send users for setup docs. Defaults preserve backward compatibility for any existing integration extending the abstract.

### `includes/integrations/class.llms.integration.bbpress.php` and `class.llms.integration.buddypress.php`

Split the existing inline `Learn More` HTML out of `$description` into `$learn_more_url`, and added an `is_configured()` override returning `true` (no API keys required).

```php
$this->description    = __( 'Restrict forums and topics to memberships and add forums to courses.', 'lifterlms' );
$this->learn_more_url = 'https://lifterlms.com/docs/lifterlms-and-bbpress/';

public function is_configured() { return true; }
```

**Why:** the previous `$description` was a single HTML blob with the docs link baked in. Splitting it lets the table render the blurb and the link independently and keeps the description clean for the per-integration settings page title.

### `includes/admin/settings/class.llms.settings.integrations.php`

Replaced the `Integration ID` `<th>`/`<td>` with a `Description` cell and added a new `Configured` column between `Installed` and `Enabled`.

**Before:**
```php
<th><?php esc_html_e( 'Integration ID', 'lifterlms' ); ?></th>
...
<td><?php echo esc_html( $integration->id ); ?></td>
```

**After:**
```php
<th><?php esc_html_e( 'Description', 'lifterlms' ); ?></th>
...
<td>
    <?php echo wp_kses_post( $integration->description ); ?>
    <?php if ( ! empty( $integration->learn_more_url ) ) : ?>
        <a href="<?php echo esc_url( $integration->learn_more_url ); ?>" target="_blank" rel="noopener noreferrer"><?php esc_html_e( 'Learn More', 'lifterlms' ); ?></a>
    <?php endif; ?>
</td>
```

The `Configured` column follows the same `tip--bottom-right` / `screen-reader-text` / `fa fa-check-circle` pattern as `Installed` and `Enabled`. When `is_configured()` returns false, the cell shows a dash and uses `$configured_help` as the tooltip text (or `Not configured` if no help was set).

## Testing

**Test 1: bbPress row**
1. Activate bbPress alongside LifterLMS.
2. Go to LifterLMS -> Settings -> Integrations.

Result: bbPress row shows the short blurb, a `Learn More` link opening `https://lifterlms.com/docs/lifterlms-and-bbpress/` in a new tab, and checks in both `Installed` and `Configured` columns.

**Test 2: BuddyPress row**
1. Activate BuddyPress.
2. Reload the Integrations page.

Result: BuddyPress row shows its blurb, `Learn More` link, and both `Installed` and `Configured` checks.

**Test 3: Inactive integration**
1. With neither bbPress nor BuddyPress active, reload the page.

Result: rows render with the `Integration ID` column gone, `Description` populated, and the `Installed`/`Configured` columns showing dashes with tooltips.

**Test 4: Coding standards**
- `vendor/bin/phpcs` and `vendor/bin/phpcs --error-severity=1 --warning-severity=6` both clean on the four modified files.
- `php -l` clean on all four.

### DIFF
--- a/.changelogs/issue-3150.yml
+++ b/.changelogs/issue-3150.yml
@@ -1,0 +1,3 @@
+significance: minor
+type: changed
+entry: "Restructured the Integrations list table: removed the Integration ID column, added a Description column with a Learn More link, and added a Configured column to indicate when an integration's required credentials are present."
--- a/includes/abstracts/llms.abstract.integration.php
+++ b/includes/abstracts/llms.abstract.integration.php
@@ -38,11 +38,27 @@ abstract class LLMS_Abstract_Integration extends LLMS_Abstract_Options_Data {
 	/**
 	 * Integration Description
 	 *
+	 * A short, one-sentence description of the integration shown in the
+	 * Integrations list table.
+	 *
 	 * Should be defined by extending class in configure() function (so it can be translated).
 	 *
 	 * @var string
 	 */
 	public $description = '';
+
+	/**
+	 * Integration "Learn More" URL
+	 *
+	 * Optional URL to the integration's getting started / documentation page.
+	 *
+	 * When defined, the Integrations list table renders a "Learn More" link
+	 * alongside the integration's description.
+	 *
+	 * @since 10.0.6
+	 * @var string
+	 */
+	public $learn_more_url = '';
 
 	/**
 	 * Integration Missing Dependencies Description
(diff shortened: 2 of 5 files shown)


brianhogg:
@faisalahammad Thanks but for this feature request issue, "Installed" is basically already "Configured" for addons, so we don't need or want a new "is_configured" method that would need to be overridden. I'll close this for now as this issue will potentially need some coordination between the add-ons. We also have an internal task to decide the wording for each add-on.

As a general note we can't hard-code version numbers, it needs to be `[version]`

PR #3188, opened 2026-06-16
feat(notifications): add order failed email notification

### LINKED ISSUE #3049: Failed order notification to customer and admin
If there's a failed order during checkout, there's no notification of this. Following up and assisting on failures could increase sales.

Failed order notification to customer and admin. We have failed recurring payment notifications, but nothing for initial orders. WooCommerce has a good model to follow for this in their emails.

Having the gateway handle this, since some failing order notifications are not legitimate.

faisalahammad:
## Summary

Adds a new `order_failed` email notification that fires when an order is marked as `llms-failed`. Covers both the initial checkout failure and the terminal recurring payment failure (after all retry rules are exhausted). Customer is on by default, product author and custom recipients are opt-in.

Fixes #3049

### New: `includes/notifications/controllers/class.llms.notification.controller.order.failed.php`

New controller `LLMS_Notification_Controller_Order_Failed` that subscribes to two existing action hooks:

- `lifterlms_order_status_failed` — fires on every transition into `llms-failed`. Covers initial checkout failure.
- `llms_automatic_payment_maximum_retries_reached` — fires from `LLMS_Order::maybe_schedule_retry()` when retry rules are exhausted. Covers terminal recurring failure.

A per-request static cache (`self::$sent`) keyed by order ID prevents a double-send when both hooks fire for the same order in one request (e.g. when the terminal retry failure cascades into the `llms-failed` status transition).

**Why:** keeps the existing source files in `LLMS_Controller_Orders` and `LLMS_Order` untouched, reuses every existing abstract, and gives the trigger full admin UI (subject, body, merge codes, per-recipient opt-in) for free.

### New: `includes/notifications/views/class.llms.notification.view.order.failed.php`

New view `LLMS_Notification_View_Order_Failed` with email body, subject, title, and a `{{RETRY_NOTICE}}` merge code that switches between "We have exhausted all automatic retry attempts..." for recurring orders and "No further payments will be attempted..." for one-time orders. Deleted-product guard on `{{PRODUCT_TITLE_LINK}}` falls back to the plain title.

### Modified: `includes/notifications/class.llms.notifications.php`

Added `'order_failed'` to the `$triggers` array in `load()`. Autoloader already maps the directories.

### New: `tests/phpunit/unit-tests/notifications/class-llms-test-notification-order-failed.php`

Test class extending `LLMS_NotificationTestCase`. Covers `test_is_registered`, `test_set_merge_data` (six merge codes plus `{{FAKE_CODE}}` pass-through), and `test_get_title`.

### New: `.changelogs/order-failed-notification.yml`

Standard changelog entry, `significance: minor`, `type: added`.

## Testing

**PHPUnit (run locally):**

```
vendor/bin/phpunit --filter LLMS_Test_Notification_Order_Failed
vendor/bin/phpunit --group notifications
```

Result: 3 new tests pass, 14 assertions. The full notifications group (29 tests) is green except for one pre-existing incomplete test (`LLMS_Test_Notification_Achievement_Earned::test_set_merge_data` — unrelated to this change).

**PHPCS:**

```
vendor/bin/phpcs \
  includes/notifications/controllers/class.llms.notification.controller.order.failed.php \
  includes/notifications/views/class.llms.notification.view.order.failed.php \
  includes/notifications/class.llms.notifications.php \
  tests/phpunit/unit-tests/notifications/class-llms-test-notification-order-failed.php
```

Result: clean.

**Manual end-to-end test** (in `TESTING_INSTRUCTIONS.md`):

1. Move any order to `llms-failed` from the admin order edit screen.
2. The customer's email queue receives the new "Order Failed" notification.
3. For recurring orders, when the last retry rule is exhausted, the customer gets one final "Order Failed" email with the retry-exhausted message.
4. Toggling the Author / Custom subscribers in the settings screen works as expected.
5. Disabling the notification in admin settings stops further sends.

## Build

`lifterlms-fix-3049.zip` available for manual install:

```
/Users/faisal/Coding/GitHub/lifterlms-fix-3049.zip
```

WP Admin → Plugins → Upload Plugin → Install Now → Activate.

## Coexistence with `payment_retry`

When a recurring payment finally fails after retries are exhausted, the site can send two related emails if both are enabled:

1. The last `payment_retry` email from the retry sequence.
2. The new `order_failed` email at terminal failure.

The terminal email is the one that matters and cannot be derived from the retry email (different merge codes, different message), so they are kept separate. To avoid duplicates, disable the email channel on `payment_retry` under LifterLMS → Settings → Notifications.

### DIFF
--- a/includes/notifications/controllers/class.llms.notification.controller.order.failed.php
+++ b/includes/notifications/controllers/class.llms.notification.controller.order.failed.php
@@ -159,7 +159,7 @@ class LLMS_Notification_Controller_Order_Failed extends LLMS_Abstract_Notificati
 
 			case 'email':
 				$options[] = $this->get_subscriber_option_array( 'author', 'no' );
-				$options[] = $this->get_subscriber_option_array( 'student', 'yes' );
+				$options[] = $this->get_subscriber_option_array( 'student', 'no' );
 				$options[] = $this->get_subscriber_option_array( 'custom', 'no' );
 				break;
 
@@ -248,12 +248,12 @@ class LLMS_Notification_Controller_Order_Failed extends LLMS_Abstract_Notificati
 	public function send_test( $type, $data = array() ) {
 
 		if ( empty( $data['order_id'] ) ) {
-			return;
+			return false;
 		}
 
 		$order         = llms_get_post( $data['order_id'] );
 		if ( ! is_a( $order, 'LLMS_Order' ) ) {
-			return;
+			return false;
 		}
 		$this->user_id = $order->get( 'user_id' );
 		$this->post_id = $order->get( 'id' );
--- a/includes/notifications/controllers/class.llms.notification.controller.order.failed.php
+++ b/includes/notifications/controllers/class.llms.notification.controller.order.failed.php
@@ -1,0 +1,267 @@
+<?php
+/**
+ * Notification Controller: Order Failed
+ *
+ * @package LifterLMS/Notifications/Controllers/Classes
+ *
+ * @since [version]
+ * @version [version]
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Notification Controller: Order Failed
+ *
+ * Sends a notification when an order is marked as failed. Covers both the
+ * initial checkout failure and the terminal recurring payment failure that
+ * occurs after all retry rules are exhausted.
+ *
+ * @since [version]
+ */
+class LLMS_Notification_Controller_Order_Failed extends LLMS_Abstract_Notification_Controller {
+
+	/**
+	 * Trigger Identifier
+	 *
+	 * @var string
+	 */
+	public $id = 'order_failed';
+
+	/**
+	 * Number of accepted arguments passed to the callback function
+	 *
+	 * @var integer
+	 */
+	protected $action_accepted_args = 1;
+
+	/**
+	 * Action hooks used to trigger sending of the notification
+	 *
+	 * @var array
+	 */
+	protected $action_hooks = array(
+		'lifterlms_order_status_failed',
+		'llms_automatic_payment_maximum_retries_reached',
+	);
+
+	/**
+	 * Determines if test notifications can be sent
+	 *
+	 * @var bool
+	 */
+	protected $testable = array(
+		'basic' => false,
+		'email' => true,
+	);
+
+	/**
+	 * Per-request cache of order IDs for which the notification has already
+	 * been dispatched. Prevents duplicate sends when both action hooks fire
+	 * for the same order in a single request (e.g. when the terminal retry
+	 * failure cascades into the `llms-failed` status transition).
+	 *
+	 * @var int[]
+	 */
+	protected static $sent = array();
+
+	/**
+	 * Callback function called when an order is marked as failed.
+	 *
+	 * @since [version]
+	 *
+	 * @param LLMS_Order $order Instance of an LLMS_Order.
+	 * @return void
+	 */
+	public function action_callback( $order = null ) {
+
+		if ( ! is_a( $order, 'LLMS_Order' ) ) {
+			return;
+		}
+
+		$order_id = $order->get( 'id' );
+		if ( ! $order_id || isset( self::$sent[ $order_id ] ) ) {
+			return;
+		}
+
+		self::$sent[ $order_id ] = true;
+
+		$this->user_id = $order->get( 'user_id' );
+		$this->post_id = $order_id;
+
+		$this->send();
+
+	}
+
+	/**
+	 * Takes a subscriber type (student, author, etc) and retrieves a User ID
+	 *
+	 * @since [version]
+	 *
+	 * @param string $subscriber Subscriber type string.
+	 * @return int|false
+	 */
+	protected function get_subscriber( $subscriber ) {
+
+		switch ( $subscriber ) {
+
+			case 'author':
+				$order = llms_get_post( $this->post_id );
+				if ( ! is_a( $order, 'LLMS_Order' ) ) {
+					return false;
+				}
+				$product = $order->get_product();
+				if ( ! $product || is_a( $product, 'WP_Post' ) ) {
+					return false;
+				}
+				$uid = $product->get( 'author' );
+				break;
+
+			case 'student':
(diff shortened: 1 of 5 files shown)


includes/notifications/controllers/class.llms.notification.controller.order.failed.php
```diff
@@ -0,0 +1,267 @@
...
+	 */
+	protected function set_subscriber_options( $type ) {
+
+		$options = array();
+
+		switch ( $type ) {
+
+			case 'email':
+				$options[] = $this->get_subscriber_option_array( 'author', 'no' );
+				$options[] = $this->get_subscriber_option_array( 'student', 'yes' );
```
brianhogg: Needs to default to 'no', admins needs to enable this.

includes/notifications/views/class.llms.notification.view.order.failed.php
```diff
@@ -0,0 +1,217 @@
...
+	/**
+	 * Setup notification title for output.
+	 *
+	 * @since [version]
+	 *
+	 * @return string
+	 */
+	protected function set_title() {
+		// Translators: %d = The order ID.
+		return sprintf( __( 'Order #%d payment failed', 'lifterlms' ), '{{ORDER_ID}}' );
```
brianhogg: This is showing as "Order #0 payment failed" since %d is not an appropriate pattern match for the `{{ORDER_ID}}` string

includes/notifications/views/class.llms.notification.view.order.failed.php
```diff
@@ -0,0 +1,217 @@
...
+
+		$rows = array(
+			'PRODUCT_TITLE_LINK' => '{{PRODUCT_TYPE}}',
+			'PLAN_TITLE'         => __( 'Plan', 'lifterlms' ),
+			'PAYMENT_AMOUNT'     => __( 'Amount', 'lifterlms' ),
+		);
+
+		ob_start();
+		?><p><?php printf( esc_html__( 'Hello %s,', 'lifterlms' ), '{{CUSTOMER_NAME}}' ); ?></p>
+		<p><?php printf( esc_html__( 'We were unable to process the payment for your %1$s on order #%2$d. {{RETRY_NOTICE}}', 'lifterlms' ), '{{PRODUCT_TITLE}}', '{{ORDER_ID}}' ); ?></p>
```
brianhogg: Incorrect placeholder %d for `{{ORDER_ID}}` shows as `#0`

brianhogg: Also `{{RETRY_NOTICE}}` is incorrect here. That's only for recurring orders and one-time orders will show a confusing message. I think we should remove it from the tempate.

includes/notifications/controllers/class.llms.notification.controller.order.failed.php
```diff
@@ -0,0 +1,267 @@
...
+	 */
+	public function send_test( $type, $data = array() ) {
+
+		if ( empty( $data['order_id'] ) ) {
+			return;
+		}
+
+		$order         = llms_get_post( $data['order_id'] );
+		if ( ! is_a( $order, 'LLMS_Order' ) ) {
+			return;
```
brianhogg: Needs to return false to match the `@return`

includes/notifications/controllers/class.llms.notification.controller.order.failed.php
```diff
@@ -0,0 +1,267 @@
...
+		}
+
+		$order_id = $order->get( 'id' );
+		if ( ! $order_id || isset( self::$sent[ $order_id ] ) ) {
+			return;
+		}
+
+		self::$sent[ $order_id ] = true;
```
brianhogg, line 87: If we're going to add this, a test would be good for when both action hooks fire in the same request making sure only one notification is sent.

brianhogg (changes requested):

includes/notifications/controllers/class.llms.notification.controller.order.failed.php
faisalahammad: Fixed. All three subscribers now default to 'no'. Admin must opt-in via LifterLMS → Settings → Notifications. The existing setup_args() and new test_double_send_prevention now enable the student subscriber explicitly via set_option() so tests still fire under the new defaults.

includes/notifications/views/class.llms.notification.view.order.failed.php
faisalahammad: Fixed. Changed %d to %s in set_title() so {{ORDER_ID}} renders as the order id.

includes/notifications/views/class.llms.notification.view.order.failed.php
faisalahammad: Fixed. Changed %2 to %2 in the default body.

faisalahammad: Removed {{RETRY_NOTICE}} from the default body. The merge code and its set_merge_data() resolver remain registered so admins can still use it in custom templates.

includes/notifications/controllers/class.llms.notification.controller.order.failed.php
faisalahammad: Fixed. Both early returns in send_test() now return false.

includes/notifications/controllers/class.llms.notification.controller.order.failed.php
faisalahammad: Added test_double_send_prevention. It calls action_callback() twice with the same order and asserts only one notification exists. Test passes (4/4 in the class, 15 assertions). The setup_args() and the new test both enable the student subscriber via set_option() under the new opt-in defaults.

faisalahammad (commented):
All 6 review items addressed:
- Subscriber defaults set to 'no' (admin opt-in)
- %d → %s in set_title() and %2 → %2 in body
- {{RETRY_NOTICE}} removed from default body template
- send_test() returns false instead of null
- test_double_send_prevention covers the static $sent dedup cache
- Tests: PHPUnit 4/4 pass, 15 assertions. PHPCS clean. CodeRabbit clean.
Ready for re-review.

PR #3189, opened 2026-06-16
fix(media-protection): allow undoing a protected file

### LINKED ISSUE #3018: Ability to "undo" a protected file
Currently, if you select a course/membership to protect a file, there's no way to undo that.

If the course/membership value is cleared for protected file, "unprotect" the file. We'd need to have a note about where the file was originally located.

We may want to not support this since:

a) the file might have been re-uploaded already and still exist
b) Links within existing documents wouldn't be automatically updated, or we'd need to do some kind of search/replace

faisalahammad:
## Summary

Adds ability to undo a protected media file by clearing the selected course or membership. The file moves back to its original public location, the protection metadata is removed, and the user sees a read-only note of the original location and a warning before unprotecting.

Fixes #3018

### `includes/class-llms-media-protector.php`

**Added:**
- `ORIGINAL_FILE_META_KEY` constant for storing the original public location.
- `remove_authorization_meta_from_media_post()` to clear auth meta and per-user cache.
- `get_original_attached_file()` to read the stored original path.

**Why:** Need a recorded original path before protection so we can restore it, and need a clean way to clear the auth meta and the cached `is_authorized_to_view` result for the current user.

### `includes/admin/class-llms-admin-media-protection-attachment-settings.php`

**Added:** `move_attachment_to_public_dir()`, `store_original_attached_file()`, and `files_are_identical()` helper.

**Changed:**
- `move_attachment_to_protected_dir()` now records the original `_wp_attached_file` before moving.
- `attachment_fields_to_save()` routes an empty product value through to the unprotect method.
- `attachment_fields_to_edit()` enables `allow_clear` on the select2, shows the original location note, and renders the unprotect warning container.

**Why:** Need a symmetric move-back method. The original path is recorded once so the unprotect method has a target. Conflict detection handles the case where a different file already exists at the original public location (returns `WP_Error`).

### `includes/class-llms-rest-fields.php`

**Changed:** `_llms_media_protection_product_id` `update_callback` treats an empty value as a request to unprotect and returns `WP_Error` on failure.

**Added:** Read-only `_llms_media_protection_original_file` REST field exposing the original location.

**Why:** The block editor needs to know the original location to render the note, and the REST path needs to honor the same "clear to unprotect" semantics as the classic save path. Backward compatibility preserved by keeping the existing field as a scalar integer.

### `src/js/admin-media-protection-block-protect.js`

**Changed:** `data-allow_clear='true'`, send `0` when the select2 is cleared, display the original location and the unprotect warning, and wrap the change event listener with a cleanup so it does not leak across modal opens.

### `assets/js/llms-admin-media-protection-attachment-settings.js`

**Changed:** Toggle the unprotect warning visibility when the classic media modal select2 is cleared.

## Testing

10 unit tests pass (19 assertions). PHPCS clean on all modified PHP files. CodeRabbit review addressed.

**Test 1: Classic media library**
1. Edit an attachment, select a course or membership, save.
2. Reopen the attachment. Original location note and unprotect warning container are shown. Select2 allows clearing.
3. Clear the selection, save. File moves back to original public path, authorization meta is removed.

**Test 2: Block editor**
1. Add a media block, click the lock toolbar button, pick a course, save.
2. Reopen the modal. Currently protected by and original location labels render.
3. Clear the selection, save. Block URL updates to the public URL.

**Test 3: Conflict**
1. Re-upload the same file to the original public path before unprotecting. Unprotect removes the protected duplicate.
2. Place a different file at the original public path. Unprotect aborts and returns `WP_Error` so the UI can show the conflict message.

### DIFF
--- a/includes/admin/class-llms-admin-media-protection-attachment-settings.php
+++ b/includes/admin/class-llms-admin-media-protection-attachment-settings.php
@@ -227,8 +227,8 @@ class LLMS_Admin_Media_Protection_Attachment_Settings {
 		}
 
 		// The original public location must have been recorded when the file was protected.
-		$original_relative = $protector->get_original_attached_file( $attachment_id );
-		if ( ! $original_relative ) {
+		$original_file_path = $protector->get_original_attached_file( $attachment_id );
+		if ( ! $original_file_path ) {
 			error_log( 'Unable to unprotect attachment ' . $attachment_id . ': original location is not recorded.' );
 			return false;
 		}
@@ -236,8 +236,7 @@ class LLMS_Admin_Media_Protection_Attachment_Settings {
 		$metadata    = wp_get_attachment_metadata( $attachment_id );
 		$current_file = get_attached_file( $attachment_id );
 
-		$uploads     = wp_upload_dir();
-		$target_file = $uploads['basedir'] . DIRECTORY_SEPARATOR . $original_relative;
+		$target_file = $original_file_path;
 		$target_dir  = dirname( $target_file );
 
 		global $wp_filesystem;
@@ -373,20 +372,11 @@ class LLMS_Admin_Media_Protection_Attachment_Settings {
 	 * @return void
 	 */
 	protected function store_original_attached_file( $attachment_id, $metadata, $file ) {
-		$relative = '';
-
-		// Prefer the relative path stored in `_wp_attached_file` so it stays accurate even
-		// if the upload base directory has been customized.
-		$attached = get_post_meta( $attachment_id, '_wp_attached_file', true );
-		if ( $attached ) {
-			$relative = ltrim( $attached, '/' );
-		}
-
-		if ( ! $relative ) {
+		if ( ! $file ) {
 			return;
 		}
 
-		update_post_meta( $attachment_id, LLMS_Media_Protector::ORIGINAL_FILE_META_KEY, $relative );
+		update_post_meta( $attachment_id, LLMS_Media_Protector::ORIGINAL_FILE_META_KEY, $file );
 	}
 
 	/**
--- a/includes/admin/class-llms-admin-media-protection-attachment-settings.php
+++ b/includes/admin/class-llms-admin-media-protection-attachment-settings.php
@@ -227,8 +227,8 @@ class LLMS_Admin_Media_Protection_Attachment_Settings {
 		}
 
 		// The original public location must have been recorded when the file was protected.
-		$original_file_path = $protector->get_original_attached_file( $attachment_id );
-		if ( ! $original_file_path ) {
+		$original_relative = $protector->get_original_attached_file( $attachment_id );
+		if ( ! $original_relative ) {
 			error_log( 'Unable to unprotect attachment ' . $attachment_id . ': original location is not recorded.' );
 			return false;
 		}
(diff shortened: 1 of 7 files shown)


faisalahammad:
The PHP 8.3 job hit a transient network error (`ECONNRESET` at `npm ci`) during dependency download. All other PHP versions (7.4, 8.0, 8.1, 8.2) passed cleanly with zero test failures. Code is solid.

Maintainer can re-trigger the failed job when ready. I don't have the perms to do it from the fork.

includes/admin/class-llms-admin-media-protection-attachment-settings.php
```diff
@@ -168,6 +201,215 @@ function move_attachment_to_protected_dir( $attachment_id ) {
...
+	 * @return void
+	 */
+	protected function store_original_attached_file( $attachment_id, $metadata, $file ) {
+		$relative = '';
+
+		// Prefer the relative path stored in `_wp_attached_file` so it stays accurate even
+		// if the upload base directory has been customized.
+		$attached = get_post_meta( $attachment_id, '_wp_attached_file', true );
+		if ( $attached ) {
+			$relative = ltrim( $attached, '/' );
```
brianhogg: Are we sure this will always be correct, depending on the server config?

brianhogg:
@faisalahammad Thanks for the PR! This is something I've wanted to add for a while.

The UX is a little confusing, but might be close to the best we can do with the limitations on the media dialog. Clicking the "x" for the course immediately removes the restriction before the message is shown, as demo'ed in this video:


https://github.com/user-attachments/assets/4aa3e44a-7591-4b56-8a7c-f0857c36dcc9

If we show the warning/note about clearing will restore the file publicly, rather than showing the note about the original location, I think this would mostly solve things. Alternatively, a button to explicitly un-restrict the file (if possible) could work too. Thoughts?

includes/admin/class-llms-admin-media-protection-attachment-settings.php
faisalahammad: You are right — stripping just the leading slash leaves the full `_wp_attached_file` value intact, so on many servers the restored path becomes `$uploads['basedir'] + wp-content/uploads/...`, which is wrong.

I changed the approach: `store_original_attached_file()` now writes the exact absolute public path (not a derived relative string). The unprotect method uses it directly, so no path math guesses, no double nesting, no reliance on how the server reports `_wp_attached_file`.

brianhogg: @faisalahammad I actually think having it as a relative path is better. Otherwise, if they switch hosting providers, the absolute path has changed and all restores would fail. It was just a thought on whether trimming on just the slash was enough, and it likely is.

If we are storing absolute paths now I think it would be best to revert that. Thanks!

faisalahammad: Reverted in 21b0c4604. `store_original_attached_file()` now stores the relative path from `_wp_attached_file` again (with `ltrim('/', ...)`). `move_attachment_to_public_dir()` uses `$uploads['basedir']` + stored relative path to reconstruct the target. Absolute paths would break on host migration — thanks for catching that.

PR #3191, opened 2026-06-16
feat(elementor): add enrollment visibility controls to widgets

### LINKED ISSUE #3099: Add "Enrollment Visibility" to Elementor Widgets
There's no way to hide/show an elementor widget based on the enrollment status of the current visitor

Add an option similar to the Enrollment Visibility option on our standard WP editor blocks.

faisalahammad:
## Summary

Adds an Enrollment Visibility section to every LifterLMS Elementor widget. Course authors can now show or hide a widget based on the current visitor's enrollment status, login state, or enrollment in specific courses or memberships, mirroring the visibility option that already exists on standard Gutenberg blocks.

The decision is evaluated server-side in the widget base class. Inside the Elementor editor the widget always renders so authors can keep editing it regardless of their own enrollment status.

Fixes #3099

### `includes/elementor/class-llms-elementor-widget-base.php`

Adds `add_visibility_controls()` which registers a new "Enrollment Visibility" section under the Advanced tab with two selects ("Display to" and "Enrolled In"). The `render()` method now evaluates those settings before delegating to a new abstract `render_widget()` method. Adds protected `is_visible()` and private `check_enrollment()` helpers that implement the same decision tree used by the blocks library. A new `llms_elementor_widget_is_visible` filter allows third-party customization of the visibility decision.

**Before:** every widget rendered unconditionally.

**After:** a widget renders only when the current visitor passes the configured visibility check.

**Why:** keeps the decision logic in one place inside the widget layer so the change does not depend on edits to the vendored `libraries/lifterlms-blocks` package.

### All six Elementor widget classes

Each widget's `render()` is renamed to `render_widget()` and now just contains the `do_shortcode()` call. Each widget's `_register_controls()` ends with `$this->add_visibility_controls()` so the new section is added automatically.

**Before:** `protected function render() { echo do_shortcode( '[...]' ); }`

**After:** `protected function render_widget() { echo do_shortcode( '[...]' ); }` plus a call to `add_visibility_controls()` at the end of `_register_controls()`.

**Why:** the base class now controls the visibility gate, subclasses only define what to render.

### `.changelogs/3099-elementor-enrollment-visibility.yml`

New hand-written changelog entry under the `added` type with `minor` significance.

## Testing

1. Install the plugin on a WordPress site with Elementor active.
2. Open a course in Elementor, add a LifterLMS widget (Pricing Table, Course Progress, etc.), open the Advanced tab.
3. Confirm the new "Enrollment Visibility" section shows "Display to" and "Enrolled In" controls.
4. For each "Display to" value, save the page and verify the front end:
   - `everyone` — visible to all visitors.
   - `enrolled users` + `in this course` — visible only when the visitor is enrolled in the current course.
   - `enrolled users` + `in any course or membership` — visible when the visitor has any enrollment.
   - `non-enrolled users or visitors` — hidden from enrolled users.
   - `logged in users` / `logged out users` — visible only to the matching audience.
5. In the Elementor editor, change the widget to "enrolled users" and confirm it still appears in the editor preview regardless of your own enrollment status.

Result: works as expected. PHPCS clean, PHPUnit suite passes aside from two pre-existing failures on dev unrelated to this change.

## Screenshots
image

## Types of changes

## Checklist:

### DIFF
--- a/includes/elementor/class-llms-elementor-widget-course-meta-info.php
+++ b/includes/elementor/class-llms-elementor-widget-course-meta-info.php
@@ -1,16 +1,54 @@
 <?php
+/**
+ * Course Meta Information Elementor widget.
+ *
+ * @package LifterLMS/Classes/Elementor
+ *
+ * @since 7.7.0
+ */
 
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * LLMS_Elementor_Widget_Course_Meta_Info class.
+ *
+ * @since 7.7.0
+ * @since [version] Added enrollment visibility controls.
+ */
 class LLMS_Elementor_Widget_Course_Meta_Info extends LLMS_Elementor_Widget_Base {
 
+	/**
+	 * Get widget name.
+	 *
+	 * @since 7.7.0
+	 *
+	 * @return string
+	 */
 	public function get_name() {
 		return 'llms_course_meta_information_widget';
 	}
 
+	/**
+	 * Get widget title.
+	 *
+	 * @since 7.7.0
+	 *
+	 * @return string
+	 */
 	public function get_title() {
 		return __( 'Course Meta Information', 'lifterlms' );
 	}
 
+	/**
+	 * Register widget controls.
+	 *
+	 * @since 7.7.0
+	 * @since [version] Added enrollment visibility section.
+	 *
+	 * @return void
+	 */
 	protected function _register_controls() {
+
 		$this->start_controls_section(
 			'content_section',
 			array(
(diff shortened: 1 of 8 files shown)


includes/elementor/class-llms-elementor-widget-course-meta-info.php
```diff
@@ -1,16 +1,54 @@
 <?php
+/**
```
brianhogg, line 2: FYI I have not been adding extra comments in addition to the comment on the class itself. We can leave them for this PR though.

brianhogg:
@faisalahammad Thanks for the PR! If we're going to add this in, we'd want it to match the current enrollment visibility functionality, specifically for when "enrolled users" or "non-enrolled users or visitors" is chosen and you can pick specific courses. Unless there's a limitation in Elementor in doing that?

CleanShot 2026-07-21 at 09 11 06@2x
CleanShot 2026-07-21 at 09 13 51@2x

The first option is also "in this course".

includes/elementor/class-llms-elementor-widget-course-meta-info.php
faisalahammad: Understood, noted. Keeping the comments as-is for this PR.

PR #3194, opened 2026-06-18
fix(media-protection): invalidate authorization cache when media is protected

### LINKED ISSUE #3167: Protecting media with Redis (Object Cache pro) fails
### Reproduction Steps

1. Have Object Cache Pro enabled
2. In the media library, protect a file

### Expected Behavior

URL of the media is rewritten to a llms_media_id=... url

### Actual Behavior

File is moved to the protected folder, but the URL remains the same (the original location of the file)

HS-275360

faisalahammad:
## Summary

Media protection URL rewriting was broken when a persistent object cache (like Object Cache Pro) is active. The authorization result for an attachment is cached per `(media_id, user_id)` with `wp_cache_add()`, so once an unprotected attachment is viewed its `'null'` sentinel stays in cache and never gets overwritten when the file is later protected. This fix switches the writes to `wp_cache_set()` and explicitly invalidates the current user's cache whenever a media file is moved into the protected directory.

Fixes #3167

### `includes/class-llms-media-protector.php`

**Before:**
```php
wp_cache_add( $cache_key, 'null', 'llms_media_authorization', $cache_expiration );
...
wp_cache_add( $cache_key, $authorization, 'llms_media_authorization', $cache_expiration );
```

**After:**
```php
wp_cache_set( $cache_key, 'null', 'llms_media_authorization', $cache_expiration );
...
wp_cache_set( $cache_key, $authorization, 'llms_media_authorization', $cache_expiration );
```

**Why:** `wp_cache_add()` only writes when the key is missing. With a persistent object cache the `'null'` sentinel from the first unprotected view survives across requests, so `is_authorized_to_view()` keeps returning the old value after the file is protected. `wp_cache_set()` overwrites the cache with the current state.

Also added `invalidate_authorization_cache()` and hooked it into `add_authorization_meta_to_media_post()` so the cache is cleared when the protection meta is added.

### `includes/admin/class-llms-admin-media-protection-attachment-settings.php`

**After:** after a successful move to the protected directory, call `$protector->invalidate_authorization_cache( $attachment_id )`.

**Why:** clears the current user's cached authorization immediately so the next `wp_get_attachment_url()` call returns the protected URL.

### `includes/class-llms-rest-fields.php`

**After:** in the REST `update_callback` for `_llms_media_protection_product_id`, after a successful move call `$protector->invalidate_authorization_cache( $object->ID )`.

**Why:** the block-editor protection modal uses the REST endpoint, so the classic-admin path alone did not cover it.

### `tests/phpunit/unit-tests/class-llms-test-media-protector.php`

New tests covering:
- `invalidate_authorization_cache()` deletes the cache for the supplied user and defaults to the current user.
- `add_authorization_meta_to_media_post()` clears the cached authorization.
- `is_authorized_to_view()` overwrites a stale `'null'` cache entry.

### `.changelogs/3167.yml`

Patch-level changelog entry for the fix.

## Testing

**Test 1: Protect an unprotected image with Object Cache Pro active**
1. Upload an image to the media library.
2. View the image URL. It should be the normal `/wp-content/uploads/...` URL.
3. Open the attachment, select a course/membership in "LifterLMS Media Protection", and save.
4. Refresh the attachment. URL is now `https://example.com/?llms_media_id={ID}`.

Result: Works as expected.

**Test 2: Re-protect an already protected media file**
1. With Object Cache Pro active, protect an attachment.
2. Re-open it and change the selected course/membership.
3. The new product is honored on the next load.

Result: Works as expected.

**Test 3: Without object cache**
1. Repeat Test 1 after dropping the `object-cache.php` drop-in.

Result: Same behavior, no regression.

**Automated tests**
```bash
composer tests-run -- --filter LLMS_Test_Media_Protector
```
Result: OK (4 tests, 10 assertions).

### DIFF
--- a/includes/class-llms-media-protector.php
+++ b/includes/class-llms-media-protector.php
@@ -526,7 +526,7 @@ class LLMS_Media_Protector {
 	/**
 	 * Build the cache key used for the authorization result of a media/user pair.
 	 *
-	 * @since 10.0.6
+	 * @since [version]
 	 *
 	 * @param int $media_id The post ID of the media file.
 	 * @param int $user_id  The ID of the user wanting to view the media file.
@@ -542,7 +542,7 @@ class LLMS_Media_Protector {
 	 * When the protection state of a media item changes the cached authorization result must be
 	 * removed so the next request does not reuse a stale value from a persistent object cache.
 	 *
-	 * @since 10.0.6
+	 * @since [version]
 	 *
 	 * @param int      $media_id The post ID of the media file.
 	 * @param int|null $user_id  Optional. The ID of the user whose cached result should be removed.
@@ -1078,7 +1078,7 @@ class LLMS_Media_Protector {
 	 * stale values being retained by persistent object caches such as Object Cache Pro.
 	 *
 	 * @since 7.7.0
-	 * @since 10.0.6 Invalidates the cached authorization result for the current user.
+	 * @since [version] Invalidates the cached authorization result for the current user.
 	 *
 	 * @param $post_id
 	 * @param string $hook_name The name of the filter that will be applied by {@see LLMS_Media_Protector::is_authorized_to_view()}.
--- a/includes/class-llms-media-protector.php
+++ b/includes/class-llms-media-protector.php
@@ -432,7 +432,7 @@ class LLMS_Media_Protector {
 			return null;
 		}
 
-		$cache_key     = 'llms-media-authorization-' . $media_id . '-' . $user_id;
+		$cache_key     = $this->get_authorization_cache_key( $media_id, $user_id );
 		$authorization = wp_cache_get( $cache_key, 'llms_media_authorization', false, $found );
 		if ( $found ) {
 			return ( ( $authorization === 'null' ) ? null : $authorization );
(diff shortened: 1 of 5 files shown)


includes/class-llms-media-protector.php
```diff
@@ -1029,6 +1073,13 @@ public function upload_dir( $uploads ) {
 	/**
 	 * Add authorization meta to the post.
 	 *
+	 * Clears any cached authorization result for the current user so a subsequent read
+	 * recomputes the authorization against the new protection state. This protects against
+	 * stale values being retained by persistent object caches such as Object Cache Pro.
+	 *
+	 * @since 7.7.0
+	 * @since 10.0.6 Invalidates the cached authorization result for the current user.
```
brianhogg: Needs [version] not hard-coded value.

Technically from the documentation standards we're not adding `@since` comments to existing public methods but I'm okay with this one.

tests/phpunit/unit-tests/class-llms-test-media-protector.php
```diff
@@ -0,0 +1,193 @@
+<?php
+/**
+ * Tests for LLMS_Media_Protector
+ *
+ * @package LifterLMS/Tests
+ *
+ * @group media_protection
+ *
+ * @since 10.0.6
```
brianhogg: [version]

tests/phpunit/unit-tests/class-llms-test-media-protector.php
```diff
@@ -0,0 +1,193 @@
...
+ */
+class LLMS_Test_Media_Protector extends LLMS_UnitTestCase {
+
+	/**
+	 * Cache group used by the media protector.
+	 *
+	 * @since 10.0.6
```
brianhogg: Remove all these `@since` comments on methods/props.

brianhogg (changes requested):

includes/class-llms-media-protector.php
faisalahammad: Switched hard-coded 10.0.6 to [version] placeholder on the new methods and the existing method docblock.

tests/phpunit/unit-tests/class-llms-test-media-protector.php
faisalahammad: Updated file header to [version] placeholder.

tests/phpunit/unit-tests/class-llms-test-media-protector.php
faisalahammad: Removed all @since lines from test methods and properties. Only file header retains @since [version].

faisalahammad (commented):
Addressed all 3 review comments:

- Switched @since 10.0.6 to @since [version] on new methods
- Switched test file header to @since [version]
- Removed @since lines from all test method/property docblocks

Ready for re-review.

PR #3195, opened 2026-06-18
fix(i18n): add and correct translators comment placement

faisalahammad:
## Summary

Adds missing `// translators:` comments above translation calls with placeholders, and corrects structural issues from the bulk i18n commit. All changes are comment-only — no logic or output changes.

Fixes WordPress.WP.I18n.MissingTranslatorsComment PHPCS warnings.

## Changes

**Standalone `<?php // translators: ?>` blocks** — moved into the same PHP block, directly above the `__()`/`_n()`/`_x()` call. The sniff requires the comment and translation call to be in the same PHP block. Standalone comment-only blocks scattered across HTML were unrecognized.

**Orphaned comments above `sprintf()`/`error_log()`** — moved directly above the `__()`/`_n()` call inside the `sprintf()` args. A comment above `sprintf()` is not recognized by the sniff even when the gettext call is nested inside.

**Capitalized `// Translators:`** — normalized to lowercase `// translators:` (6 instances across admin tools, bulk-enroll, instructors table). WordPress coding standard uses lowercase.

**Duplicate comments** — removed where a pre-existing translators comment already existed next to the new one.

**Block comment removed** — one `/* Translators: %s - name of class. */` block comment replaced with standard inline format.

## Testing

**PHPCS i18n sniff:**
```bash
# Custom ruleset at /tmp/phpcs-i18n.xml
vendor/bin/phpcs --standard=/tmp/phpcs-i18n.xml includes templates
```
Result: 0 MissingTranslatorsComment warnings. 2 residual warnings are pre-existing (abstract.llms.update.php false positive, parent-course.php NoHtmlWrappedStrings).

**PHP syntax:**
```bash
git diff HEAD~1 --name-only | xargs -I{} php -l {} 2>&1 | grep -v "No syntax errors"
```
Result: all files pass.

**Full lint:**
```bash
composer check-cs
```
Result: exits 0, no new violations.


brianhogg:
Would rather not flag all these files in a wp.org update. Many of the comments don't really clarify (ie. `// translators: %s: value`) or are obvious without. Closing for now but can re-open if needed.

PR #3197, opened 2026-06-19
fix(filesystem): replace raw filesystem operations with WP_Filesystem APIs

faisalahammad:
## Summary

Replace raw PHP filesystem operations flagged by Plugin Check with WP_Filesystem APIs and wp_delete_file()/wp_tempnam() helpers across 5 files. All 18 file_system_operations_* errors resolved.

### includes/class.llms.data.php — get_file_version()

**Before:**
```php
$fp = fopen( $file, 'r' );
$file_data = fread( $fp, 8192 );
fclose( $fp );
```

**After:**
```php
global $wp_filesystem;
require_once ABSPATH . 'wp-admin/includes/file.php';
if ( ! WP_Filesystem() ) {
    return '';
}
$file_data = $wp_filesystem->get_contents( $file );
$file_data = substr( $file_data, 0, 8192 );
```

**Why:** Plugin Check flags fopen/fread/fclose as raw filesystem calls. WP_Filesystem get_contents() reads in one call and we slice the first 8KB via substr.

### includes/class.llms.install.php — create_files()

**Before:**
```php
$file_handle = @fopen( trailingslashit( $file['base'] ) . $file['file'], 'w' );
if ( $file_handle ) {
    fwrite( $file_handle, $file['content'] );
    fclose( $file_handle );
}
```

**After:**
```php
global $wp_filesystem;
require_once ABSPATH . 'wp-admin/includes/file.php';
WP_Filesystem();
// ...
if ( wp_mkdir_p( $file['base'] ) && ! $wp_filesystem->exists( $file_path ) ) {
    $wp_filesystem->put_contents( $file_path, $file['content'], 0644 );
}
```

**Why:** Drop phpcs:ignore comments and use WP_Filesystem put_contents(). Removes 3 phpcs:ignore annotations on the previous lines.

### includes/admin/class-llms-admin-export-download.php

**Before:** `unlink( $path );`
**After:** `wp_delete_file( $path );`

**Why:** wp_delete_file() is the WordPress wrapper for unlink() and is the recommended API per WordPress coding standards.

### includes/admin/class.llms.admin.page.status.php — remove_log_file()

**Before:**
```php
if ( $log && is_file( $log ) && is_writable( $log ) ) {
    unlink( $log );
```

**After:**
```php
global $wp_filesystem;
require_once ABSPATH . 'wp-admin/includes/file.php';
WP_Filesystem();

if ( $log && $wp_filesystem->exists( $log ) && $wp_filesystem->is_writable( $log ) ) {
    wp_delete_file( $log );
```

**Why:** Plugin Check flags is_writable and unlink. Use the WP_Filesystem equivalents plus wp_delete_file.

#### array_to_csv()

**Before:** Used `fopen('php://temp')` + `fputcsv` + `fread` + `fclose` to build an in-memory CSV string.

**After:** Builds CSV directly via a new `csv_row()` helper that produces RFC 4180 compliant output without touching a stream.

**Why:** fopen/fread/fclose were flagged. CSV format identical to fputcsv for flat fields (always-enclosed, doubled enclosure for embedded quotes). Signature and return type unchanged, existing callers unaffected.

#### send_email()

**Before:** `tempnam` + `fopen` + `fwrite` + `rename` + `fclose` + `unlink`.

**After:** `wp_tempnam()` + `WP_Filesystem::put_contents()` + `wp_delete_file()`.

**Why:** Use the WordPress temp file helper and WP_Filesystem write instead of low-level stream and filesystem ops.

## Skipped

- `includes/class-llms-media-protector.php:606` `readfile()`: already has `phpcs:ignore`. Intentional. Streams large protected media files to the browser. `WP_Filesystem::get_contents()` would load the entire file into memory which is worse for large media.

## Testing

1. `php -l` on all 5 modified files: no syntax errors
2. `vendor/bin/phpcs --sniffs=WordPress.WP.AlternativeFunctions` on modified files: zero errors
3. `composer check-cs` full project run: zero errors
4. Manual review of each change confirms behavior preserved (file read returns same 8KB slice, file write produces identical content, log deletion logic equivalent, CSV output identical for flat voucher data, email attachment still sent with .csv extension)

## Build

Two commits on `fix/filesystem-api-replacement`:
- `468d6f9fb` fix: replace raw filesystem operations with WP_Filesystem APIs
- `bc2c4c9a9` chore: add changelog entry for filesystem API fix

### DIFF
--- a/.changelogs/3197.yml
+++ b/.changelogs/3197.yml
@@ -1,0 +1,3 @@
+significance: patch
+type: fixed
+entry: Replaced raw PHP filesystem operations (fopen, fread, fwrite, fclose, unlink, rename, is_writable) with WordPress WP_Filesystem APIs and wp_delete_file() helpers across 5 files to resolve Plugin Check file_system_operations warnings.
--- a/includes/admin/class-llms-admin-export-download.php
+++ b/includes/admin/class-llms-admin-export-download.php
@@ -68,7 +68,7 @@ class LLMS_Admin_Export_Download {
 		header( 'Content-Disposition: attachment; filename="' . $export . '"' );
 
 		$file = file_get_contents( $path );
-		unlink( $path );
+		wp_delete_file( $path );
 		// phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 		echo $file;
 		exit;
--- a/includes/admin/class.llms.admin.page.status.php
+++ b/includes/admin/class.llms.admin.page.status.php
@@ -275,8 +275,13 @@ class LLMS_Admin_Page_Status {
 			$handle = sanitize_title( wp_unslash( $_REQUEST['llms_delete_log'] ) );
 			$log    = isset( $logs[ $handle ] ) ? $logs[ $handle ] : false;
 
-			if ( $log && is_file( $log ) && is_writable( $log ) ) {
-				unlink( $log );
+			global $wp_filesystem;
+			/** @var WP_Filesystem_Base $wp_filesystem */
+			require_once ABSPATH . 'wp-admin/includes/file.php';
+			WP_Filesystem();
+
+			if ( $log && $wp_filesystem->exists( $log ) && $wp_filesystem->is_writable( $log ) ) {
+				wp_delete_file( $log );
 				llms_redirect_and_exit( esc_url_raw( self::get_url( 'logs' ) ) );
 			}
 		}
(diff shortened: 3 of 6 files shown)


brianhogg:
Again we're not looking to fix Plugin Checker warnings unless there's a real reason to. Things like a hand-rolled `csv_row` method would need some testing and could affect exports across the board. `$wp_filesystem->get_contents` vs `fread( $fp, 8192 )` is a change in behaviour.

Closing for now, but will keep in mind if we are forced to switch to only WP methods for file system access.

PR #3205, opened 2026-06-19
fix(naming): resolve naming-convention warnings from Plugin Check

faisalahammad:
## Summary

Fix naming-convention warnings flagged by the WordPress Plugin Check report. Renames unprefixed functions, renames the meta box field interface to follow `LLMS_` class prefix convention, and renames the interface file to match WordPress file naming standards. Backward compatibility preserved via deprecated wrappers.

### Function Renames (includes/functions/llms.functions.course.php)

**Before:**
```php
function get_course( $course_id ) { ... }
function get_lesson( $lesson_id ) { ... }
```

**After:**
```php
function llms_get_course( $course_id ) { ... }
function llms_get_lesson( $lesson_id ) { ... }
```

**Why:** Unprefixed functions risk collisions with other plugins. Added `llms_` prefix per coding standards. Old functions kept as deprecated wrappers calling the new ones.

### Template Function Renames (includes/llms.template.functions.php)

**Before:**
```php
function get_course_data( $course_id ) { ... }
function get_lesson_data( $lesson_id ) { ... }
function get_payment_options( $order_id ) { ... }
```

**After:**
```php
function llms_get_course_data( $course_id ) { ... }
function llms_get_lesson_data( $lesson_id ) { ... }
function llms_get_payment_options( $order_id ) { ... }
```

**Why:** Same prefix issue. Deprecated wrappers maintain backward compatibility.

### Interface Rename (Meta Box Fields)

**Before:**
```php
interface Meta_Box_Field_Interface { ... }
```

**After:**
```php
interface LLMS_Meta_Box_Field_Interface { ... }
```

**Why:** Core classes must use `LLMS_` prefix per coding standards. Updated all 19 implementing classes and the class loader.

### File Rename

**Before:** `includes/admin/meta-boxes/fields/interface.meta.box.field.php`

**After:** `includes/admin/meta-boxes/fields/class-llms-metabox-field-interface.php`

**Why:** WordPress file naming standard requires `class-llms-<name>.php` format.

## Testing

**Test 1: Backward compatibility**
1. Call `get_course( $id )` in a theme or plugin
2. Verify it returns same result as `llms_get_course( $id )`
3. Check deprecated function triggers notice in debug mode

Result: Old functions work, new functions work, deprecation notices fire correctly.

**Test 2: Meta box fields**
1. Open course/lesson/membership edit screen
2. Verify all meta box fields render correctly
3. Save and confirm data persists

Result: All field types (text, textarea, editor, checkbox, select, etc.) work as expected.

### DIFF
--- a/.changelogs/naming-conventions.yml
+++ b/.changelogs/naming-conventions.yml
@@ -1,0 +1,3 @@
+significance: patch
+type: fixed
+entry: "Fixed naming-convention warnings reported by WordPress.org Plugin Check. Renamed several internal helper functions (llms_get_course, llms_get_lesson, llms_get_section_data, llms_get_lesson_data, llms_get_product_query_var, llms_get_available_payment_options) and the internal admin meta-box field interface (LLMS_Metabox_Field_Interface). Public API functions, template variables, WordPress core hook names, and vendored libraries were not changed. Deprecated wrappers added for backward compatibility."
--- a/includes/admin/post-types/meta-boxes/fields/class-llms-metabox-field-interface.php
+++ b/includes/admin/post-types/meta-boxes/fields/class-llms-metabox-field-interface.php
@@ -11,11 +11,12 @@
 defined( 'ABSPATH' ) || exit;
 
 /**
- * Meta_Box_Field_Interface interface
+ * LLMS_Metabox_Field_Interface interface
  *
  * @since Unknown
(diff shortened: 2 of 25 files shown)


brianhogg:
We would need to wait a long time before actually removing these deprecated public functions. If someone is extending the Meta_Box_Field_Interface class this would break their site. If we find a conflict down the road, we'll address it separately then. We have had no reports of collisions to date and can re-evaluate at that point.

PR #3209, opened 2026-06-19
fix: resolve PHPCS Plugin Check false positives and wp_redirect security warnings

faisalahammad:
## Summary

Replace `wp_redirect()` with `wp_safe_redirect()` in the template loader and lesson progression controller to mitigate open redirect risks flagged by the WordPress Plugin Check tool.

### includes/class.llms.template.loader.php

**Before:** `wp_redirect()` used for content restriction redirects.

**After:** `wp_safe_redirect()`.

**Why:** Restricts redirect destinations to local URLs, prevents open redirect.

### includes/controllers/class.llms.controller.lesson.progression.php

**Before:** `wp_redirect()` used for lesson auto-advance after completion.

**After:** `wp_safe_redirect()`.

**Why:** Same open redirect mitigation.

## Testing

**Test: Lesson auto-advance**

1. Enroll student in course with multiple lessons
2. Complete lesson
3. Verify redirect goes to next lesson

Result: Works as expected

**Test: Content restriction redirect**

1. Set course to members only
2. Access course page as logged-out user
3. Verify redirect to configured page

### DIFF
--- a/includes/admin/post-types/meta-boxes/fields/llms.interface.meta.box.field.php
+++ b/includes/admin/post-types/meta-boxes/fields/llms.interface.meta.box.field.php
@@ -15,7 +15,6 @@ defined( 'ABSPATH' ) || exit;
  *
  * @since Unknown
  */
-// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedClassFound -- Legacy interface, backward compatibility.
 interface Meta_Box_Field_Interface {
 
 	public function output();
--- a/includes/admin/post-types/meta-boxes/fields/llms.interface.meta.box.field.php
+++ b/includes/admin/post-types/meta-boxes/fields/llms.interface.meta.box.field.php
@@ -15,6 +15,7 @@ defined( 'ABSPATH' ) || exit;
  *
  * @since Unknown
  */
+// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedClassFound -- Legacy interface, backward compatibility.
 interface Meta_Box_Field_Interface {
 
 	public function output();
--- a/templates/course/parent-course.php
+++ b/templates/course/parent-course.php
@@ -15,8 +15,4 @@ global $post;
 
 $lesson = new LLMS_Lesson( $post );
 
-echo '<p class="llms-parent-course-link">' . sprintf(
-	// translators: %s: parent course title.
-	__( 'Back to: %s', 'lifterlms' ),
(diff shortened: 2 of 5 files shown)


brianhogg:
@faisalahammad The "fixes" number is pointing to a pull request. Not sure if there's an issue this is fixing or if that should be removed from the description?

templates/course/parent-course.php
```diff
@@ -15,4 +15,8 @@
 
 $lesson = new LLMS_Lesson( $post );
 
-echo wp_kses_post( sprintf( __( '<p class="llms-parent-course-link">Back to: <a class="llms-lesson-link" href="%1$s">%2$s</a></p>', 'lifterlms' ), get_permalink( $lesson->get( 'parent_course' ) ), get_the_title( $lesson->get( 'parent_course' ) ) ) );
+echo '<p class="llms-parent-course-link">' . sprintf(
```
brianhogg: I don't want to change the existing translation string because of a plugin check false positive. We can revert this.

includes/admin/post-types/meta-boxes/fields/llms.interface.meta.box.field.php
```diff
@@ -15,6 +15,7 @@
  *
  * @since Unknown
  */
+// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedClassFound -- Legacy interface, backward compatibility.
```
brianhogg: Likewise I don't want to add an ignore just for the plugin check tool. We can remove this and add or possibly rename if needed down the road.

brianhogg (changes requested):

faisalahammad:
Removed the `Fixes #3194` line. That PR is unrelated, you were right. Also reverted the parent-course.php translation change and the meta-box field phpcs:ignore per your review. PR now only contains the two wp_safe_redirect() swaps and the changelog entry.

brianhogg:
@faisalahammad I see the description change but could you confirm the changes have been pushed? I'm still seeing the change to includes/admin/post-types/meta-boxes/fields/llms.interface.meta.box.field.php and templates/course/parent-course.php in the diff https://github.com/gocodebox/lifterlms/pull/3209/changes. Thanks!

faisalahammad:
@brianhogg I had to step away from my desk halfway through because of Jumma prayer, so the changes weren’t fully pushed yet. When you get a chance, could you please take another look and re-review it? Really appreciate your help! Thanks a lot.

PR #3229, opened 2026-07-01
Feature/help scout beacon

alexjpanagis:
Adds an in-dashboard HelpScout help beacon so users can reach documentation and support (or pre-sales) without leaving wp-admin.

The entire feature lives in a new self-instantiating admin class, `LLMS_Admin_Help_Beacon` (following the `LLMS_Admin_Review` pattern), plus a one-line `require_once` in the loader and a changelog entry. Three files:

- **`includes/admin/class-llms-admin-help-beacon.php`** (new)
  - **Screen scoping** — only renders on the core LifterLMS management screens (`toplevel_page_lifterlms`, `lifterlms_page_llms-{dashboard,settings,reporting,import,status,resources,add-ons}`). Intentionally excluded from the post editors and Course Builder so it never clutters the editing UI. Filter: `llms_help_beacon_screens`.
  - **License-based beacon selection** — shows the support beacon on licensed sites and the pre-sales beacon otherwise. `is_licensed()` checks the bundled helper (`llms_helper_options()->get_license_keys()`, guarded by `function_exists`). Filters: `llms_help_beacon_is_licensed`, `llms_help_beacon_id`. Beacon IDs are class constants.
  - **Consent gate** — renders a floating launcher (styled to match `.llms-button-primary`); the HelpScout script is injected only after the user confirms a `confirm()` prompt, so no third-party script or cookie loads until opt-in.
  - **Form prefill** — passes the current user's name/email to `Beacon('prefill', …)`.
  - **System report as agent-only context (support beacon only)** — `get_session_data()` reuses `LLMS_Data::get_data('system_report')` and maps each top-level section to a HelpScout `session-data` attribute, bounded to the API's 20-key / 10,000-char limits. This appears in the agent's conversation view, not the customer's message box. Filter: `llms_help_beacon_session_data` (return `[]` to disable).
  - Inline JS/CSS output via the existing `llms()->assets->enqueue_inline()` system; dynamic data encoded with `wp_json_encode( …, JSON_HEX_TAG | JSON_HEX_AMP )`.
- **`includes/class-llms-loader.php`** — `require_once` the new class in `includes_admin()` and an `@since [version]` docblock note.
- **`.changelogs/help-scout-beacon.yml`** — changelog entry (`minor` / `added`).

All new hooks use the `llms_` prefix, all new symbols use `@since [version]`, and every change is additive and filterable — no public signatures changed and no build step (pure PHP inline output).

## How has this been tested?

Manually tested in a local environment (Local / WP 7.0 / PHP 8.2.29) with LifterLMS running from this branch, unlicensed and (pre-sales) install + adding a license to compare (video shared in Slack): 

- Verified the launcher appears on the LifterLMS management screens (Dashboard, Settings, Status) and does **not** appear on the course/lesson editors or Course Builder.
- Verified the consent `confirm()` gate: no HelpScout script or cookie loads until the user clicks OK.
- Verified the pre-sales beacon loads on the unlicensed site and that the current user's name/email are pre-filled in the form.
- Verified the support beacon loads when LifterLMS is licensed and that the current user's name/email are pre-filled in the form.
- Note: HelpScout's message send requires a secure context (`crypto.randomUUID`), so the admin must be served over HTTPS.
- The support-beacon path (session data) is gated to licensed sites; it can be exercised locally with `add_filter( 'llms_help_beacon_is_licensed', '__return_true' )`, which switches to the support beacon and fires `Beacon('session-data', …)` with the system report. Agent-side visibility was not verified locally (requires the support HelpScout account).
- `php -l` is clean on all three changed files.

The PHPUnit suite and PHPCS were **not** run in this environment — please rely on CI for those (see checklist below).

## Types of changes

New feature (non-breaking change which adds functionality).

## Checklist:
- [x] This PR requires and contains at least one changelog fil yml file: `npm run dev changelog add -- -i` and follow theprompt. See also: https://github.com/gocodebox/lifterlms/blob/trunk/packages/dev/README.md#changelog-add -->
- [ ] My code follows the LifterLMS Coding & Documentation Stay; PHPCS not run locally in this environment. Check code:`composer run-script check-cs-errors` -->


includes/admin/class-llms-admin-help-beacon.php
```diff
@@ -0,0 +1,393 @@
...
+		}
+
+		/**
+		 * Filters whether the site is considered licensed when determining which help beacon to display.
+		 *
+		 * @since [version]
+		 *
+		 * @param boolean $licensed Whether a valid license key was found.
+		 */
+		return (bool) apply_filters( 'llms_help_beacon_is_licensed', $licensed );
```
brianhogg: Remove this filter

includes/admin/class-llms-admin-help-beacon.php
```diff
@@ -0,0 +1,393 @@
...
+
+		/**
+		 * Filters the HelpScout beacon ID to display.
+		 *
+		 * @since [version]
+		 *
+		 * @param string  $beacon_id   The HelpScout beacon ID.
+		 * @param boolean $is_licensed Whether the site has a valid license.
+		 */
+		return apply_filters( 'llms_help_beacon_id', $beacon_id, $is_licensed );
```
brianhogg: Remove this filter

includes/admin/class-llms-admin-help-beacon.php
```diff
@@ -0,0 +1,393 @@
...
+	 *
+	 * @return boolean
+	 */
+	public function is_licensed() {
+
+		$licensed = false;
+
+		if ( function_exists( 'llms_helper_options' ) ) {
+			foreach ( llms_helper_options()->get_license_keys() as $key ) {
+				if ( isset( $key['status'] ) && 1 == $key['status'] ) { // phpcs:ignore Universal.Operators.StrictComparisons.LooseEqual -- Status is stored as a numeric string.
```
brianhogg: Wrap $key['status'] in absint() or something to remove this phpcs:ignore and loose comparison.

brianhogg (changes requested):

.changelogs/help-scout-beacon.yml
```diff
@@ -0,0 +1,3 @@
+significance: minor
+type: added
+entry: "Added a HelpScout help beacon to the LifterLMS admin screens and to the screens under the Courses, Memberships, Engagements, and Orders menus. The support beacon is shown to sites with a valid license and the pre-sales beacon otherwise, the current user's name and email are pre-filled, and the LifterLMS system report is attached to conversations as agent-only session data. The beacon only loads after the user opts in via a consent prompt."
```
brianhogg: This is very long, should be a small one-sentence summary of the change. We're not attaching session data either now I don't think?

brianhogg (changes requested):

.changelogs/help-scout-beacon.yml
alexjpanagis: Good call, thank you. I forgot to update the changelog to match the latest changes made to the PR. This is now done to shorten/clarify. Let me know if this is all set now. 🙌

brianhogg:
@alexjpanagis Could you update the PR with the base branch? It doesn't look like I'm able to do so, and it's needed to test the changes with the consolidated libraries included in core now. Thanks!

alexjpanagis:
@brianhogg Done, let me know if that is what you meant :)

brianhogg:
@alexjpanagis I'd have to test Yoast again to see how they do it, but it looks like they need to hit Okay to the consent every time they open the beacon and every page load. I think we'd want that consent to persist? Not a deal breaker and requiring consent each time is more privacy forward.

alexjpanagis:
@brianhogg Correct, yes. I wanted to play it on the safe side, so the implementation takes the same approach as Yoast (installed from the WP plugin repo, not their paid plugin). 👍 

The one difference (to my knowledge) is that Yoast has a bypass: ask_consent = ! options->get( 'tracking' ), so if the site already opted into Yoast's usage tracking, in my testing, they skip the prompt entirely. We don't have an equivalent global opt-in (that I am aware of), so we always ask, which I feel makes more sense (i.e., is in line with being more privacy forward, as you mentioned).

PR #3238, opened 2026-07-13
Add multi-select for engagement trigger associated items

### LINKED ISSUE #3233: Add ability to choose multiple associated items (ie. courses) for the engagement
We now have the ability to select "all" for an engagement trigger, meaning any course/membership would match and trigger it. You can also select a single course. But if you want to select some, or select all but one, you can't.

Add ability to choose multiple associated items (ie. courses) for the engagement, or items(s) to exclude if “All” is selected.

faisalahammad:
## Description

Currently engagement triggers can target either one specific post (course, lesson, membership) or any post of that type. This adds the ability to select multiple items, limiting triggers to a specific set of courses, lessons, memberships, etc.

Fixes #3233

### Changes

- **Multi-select UI:** Single-select Select2 replaced with a multi-select for each trigger post type (course, lesson, membership, access plan, quiz, section). Multiple IDs can be picked for a single engagement.
- **Storage:** Each selected ID is stored in its own `_llms_engagement_trigger_post` postmeta row (one row per ID). An empty selection stores a single `any` row. This keeps the trigger query as a regular equality match instead of `FIND_IN_SET`.
- **Query:** `get_engagements()` uses `meta_value = %d OR meta_value = 'any'`, matching the per-row storage. Backward compatible with existing single-ID engagements.
- **Admin list table:** The trigger column loops the stored IDs and renders each as a linked item.

## How has this been tested?

Physically tested in a local WordPress install against the built plugin zip.

**Test 1: Multi-select UI**
1. Go to Engagements > Add New
2. Select a trigger type (e.g. "Course Enrolled")
3. Select two or more courses from the course selector
4. Fill in the engagement type and engagement fields
5. Click Publish
Result: Multiple courses are saved (one postmeta row per ID) and listed in the admin columns.

**Test 2: Backward compatibility**
1. Create an engagement with a single course selected
2. Verify it still works correctly (stored as a single ID, matched properly)
Result: Single-ID engagements unaffected.

**Test 3: "Any" behavior**
1. Create an engagement with no courses selected (leave blank)
2. Save and reload
Result: Trigger applies to all courses (stored as "any").

**Test 4: Trigger fires for each selected item**
1. With two courses selected, enroll a student in course A and confirm the engagement fires
2. Enroll in course B and confirm it fires again
Result: Engagement triggers for both selected courses.

Automated tests: engagement and notification PHPUnit suites pass (`class-llms-test-engagements`, `class-llms-test-engagement-handler`, notification earned tests). PHPCS (`check-cs-errors`) passes clean.

## Screenshots

image

## Types of changes

New feature (non-breaking change which adds functionality).

## Checklist:


includes/admin/post-types/meta-boxes/class.llms.meta.box.engagement.php
```diff
@@ -351,11 +353,17 @@ public function save( $post_id ) {
...
 
-			// An empty trigger post means "any" — store explicitly so the intent is clear.
-			if ( empty( $val ) ) {
-				$val = 'any';
+			if ( is_array( $posted ) ) {
+				$ids = array_map( 'absint', $posted );
+				$ids = array_filter( $ids );
+				$val = $ids ? implode( ',', $ids ) : 'any';
+			} else {
+				$val = llms_filter_input_sanitize_string( INPUT_POST, '_faux_engagement_trigger_post_' . $var );
```
brianhogg: I'd like to move away from using the `llms_filter_input_sanitize_string` and other similar wrappers. This is in line with our removal of `llms_verify_nonce` that things like the plugin checker didn't recognize as a way to validate/sanitize.

includes/admin/post-types/post-tables/class.llms.admin.post.table.engagements.php
```diff
@@ -79,16 +79,25 @@ public function manage_columns( $column, $post_id ) {
...
+							$title = get_the_title( $id );
+							$url   = get_edit_post_link( $id );
+						}
+
+						$links[] = sprintf( '<a href="%s">%s (ID# %d)</a>', esc_url( $url ), esc_html( $title ), esc_html( $id ) );
+
 					}
 
-					printf( '<a href="%s">%s (ID# %d)</a>', esc_url( $link ), esc_html( $title ), esc_html( $tid ) );
+					echo wp_kses_post( implode( ', ', $links ) );
```
brianhogg: I'd rather we echo as built vs creating a new array of links. It keeps the esc_* methods happening at the point of echo rather than beforehand.

includes/admin/post-types/meta-boxes/class.llms.meta.box.engagement.php
```diff
@@ -70,24 +70,24 @@ public function get_fields() {
 					'course_purchased',
 				),
 				'id'               => '_faux_engagement_trigger_post_course',
-				'label'            => __( 'Select a Course', 'lifterlms' ),
-				'placeholder'      => __( 'Any Course', 'lifterlms' ),
+				'label'            => __( 'Select Course(s)', 'lifterlms' ),
+				'placeholder'      => __( 'Add courses to limit this trigger', 'lifterlms' ),
```
brianhogg: The placeholders should be kept as "Any ___" to clearly indicate that by default, the engagement will apply to any lesson/course/etc if nothing is selected.

includes/class.llms.engagements.php
```diff
@@ -152,7 +152,7 @@ private function get_engagements( $trigger_type, $related_post_id = '' ) {
 			$related_select = ', relation_meta.meta_value AS related_post_id';
 			$related_join   = "LEFT JOIN $wpdb->postmeta AS relation_meta ON triggers.ID = relation_meta.post_id";
 			$related_where  = $wpdb->prepare(
-				"AND relation_meta.meta_key = '_llms_engagement_trigger_post' AND ( relation_meta.meta_value = %d OR relation_meta.meta_value = 'any' )",
+				"AND relation_meta.meta_key = '_llms_engagement_trigger_post' AND ( FIND_IN_SET( %d, relation_meta.meta_value ) OR relation_meta.meta_value = 'any' )",
```
brianhogg: It looks like using FIND_IN_SET can be dramatically less efficient of a query. We can store the IDs in multiple meta rows with the same meta key instead and do a regular `IN`.

brianhogg (changes requested):
Thanks for the submission! I've added some comments.

To confirm for this and for the other PRs, have they been physically tested or just tested via AI? While using AI to create the code is fine, we did update the [contributing guidelines](https://github.com/gocodebox/lifterlms/blob/trunk/.github/CONTRIBUTING.md) to make clear that the PR needs to use the template that appears when you physically open the PR in github, which includes the checkbox stating that the PR has been tested.

includes/admin/post-types/meta-boxes/class.llms.meta.box.engagement.php
faisalahammad: Done. Removed the wrapper and read $_POST directly with wp_unslash and absint. Nonce is checked at the top of save().

includes/admin/post-types/post-tables/class.llms.admin.post.table.engagements.php
faisalahammad: Done. Removed the links array. Each link is echoed inside the loop with esc_url and esc_html at the point of echo.

includes/admin/post-types/meta-boxes/class.llms.meta.box.engagement.php
faisalahammad: Done. Placeholders are back to Any Course, Any Lesson, Any Access Plan, Any Membership, Any Quiz and Any Section.

includes/class.llms.engagements.php
faisalahammad: Done. FIND_IN_SET is gone. Each selected ID is stored in its own _llms_engagement_trigger_post meta row and the query uses a regular equality match (meta_value = %d OR meta_value = 'any'). Empty selection stores a single any row.

faisalahammad (commented):
Thanks for the review. All four comments are addressed in 0070aa6.

- Removed the llms_filter_input_sanitize_string wrapper and read $_POST directly with wp_unslash and absint (nonce checked at the top of save).
- Echo the trigger links inside the loop so esc_url and esc_html run at the point of echo.
- Restored the Any ___ placeholders for all trigger post types.
- Dropped FIND_IN_SET. Each selected ID is stored in its own _llms_engagement_trigger_post meta row and the query uses a regular equality match.

Also updated this PR description to use the default template with the testing checklist filled in. Ready for another look.

.changelogs/3233-multi-select-engagement-triggers.yml
```diff
@@ -0,0 +1,3 @@
+significance: minor
+type: added
+entry: "Added ability to select multiple associated items (courses, lessons, memberships, etc.) for engagement triggers."
```
brianhogg, line 4: It looks like we're missing the "items(s) to exclude if “All” is selected" piece of the feature request, but I'm okay with doing "all or some" first.

brianhogg:
@faisalahammad Please also confirm the comment above https://github.com/gocodebox/lifterlms/pull/3238#pullrequestreview-4702861427

includes/admin/post-types/meta-boxes/class.llms.meta.box.engagement.php
```diff
@@ -351,19 +353,28 @@ public function save( $post_id ) {
 
 		if ( $var ) {
 
-			$val = llms_filter_input_sanitize_string( INPUT_POST, '_faux_engagement_trigger_post_' . $var );
+			// The nonce is verified at the top of save(). Read the posted IDs directly.
+			// phpcs:ignore WordPress.Security.NonceVerification.Missing
```
brianhogg: @faisalahammad Why do we need this phpcs:ignore if we're doing the nonce verification in the same function.

faisalahammad: Good catch. Removed the phpcs:ignore in f65c81a. Nonce is verified at the top of save() so the sniff passes without it.

.changelogs/3233-multi-select-engagement-triggers.yml
faisalahammad: Agreed, "all or some" is in this PR. The exclude-when-All part is deferred to a follow-up issue.

faisalahammad:
Addressed the open review comments in f65c81a23.

- Metabox phpcs:ignore (line 357): removed it. The nonce is verified at the top of save() so check-cs-errors passes without the ignore.
- Changelog: agreed to keep "all or some" in this PR. The exclude-when-All part is deferred to a follow-up.

Both threads resolved. Ready for another look when you have time.

faisalahammad:
To confirm both points above:

- This PR was physically tested in a local WordPress install, not AI only. I installed the built plugin zip and went through each test case in the description by hand (multi-select save, single-ID backward compat, blank "any" behavior, and the engagement firing for each selected course). The trigger fired as expected in each case.
- The PR description follows the pull request template. All four checklist items are checked, including "My code has been tested."

@brianhogg let me know if you would like more detail on the manual test steps.

brianhogg:
@faisalahammad Will test but looks like we're pretty much there. Feel free to update the changelog yml to add your `@faisalahammad` attribution if you'd like that to appear. Thanks!

brianhogg (approved):

faisalahammad:
Re: brianhogg attribution — added in commit 453c41c. Changelog yml now carries attributions: [@faisalahammad] on the 3233 entry, which renders as Thanks @faisalahammad! when the release tooling runs. Schema validates clean.

brianhogg:
@faisalahammad Could the height of the box to select item(s) be the standard height by default? I'm not sure why it's double height in the screenshot (possibly WP 7.0 related formatting), as it should expand if many courses are picked.

faisalahammad:
@brianhogg Hi, I’ve tried adjusting the height of the box to the standard size, but I wasn’t able to reduce it. Could you please help me look into this? Thanks!

brianhogg:
@faisalahammad Fixed the styling to match multi-select boxes like Enroll New Students (edit course), but when testing the engagements persist to `postmeta`, however do not load back after saving or reloading. Is this different than what you're seeing?

https://github.com/user-attachments/assets/106ddaf7-0444-4f3f-82e8-c4e3bb29ed47

brianhogg (changes requested):
Does not work when testing. Selected trigger post(s) persist, but do not load back into the editor after save/reload.

faisalahammad:
Hi @brianhogg,
Sorry for the trouble. Should I resolve the conflicts and push? I'm confused because you've already merged into the 'dev' branch. Thanks for your help!

faisalahammad:
Fixed. Saved trigger post IDs now load back into the multi-select on editor reload.

The faux field (`_faux_engagement_trigger_post_*`) stores no meta, so the Select renderer's default `selected` value was empty — no `<option>` got marked `selected`, and ajax-Select2 hydrates only from existing `<option selected>` markup. The fix passes the saved IDs as the field `selected` key, which the renderer already honors (same pattern as the coupon/voucher multi-selects).

Demo: https://screendrop-worker.faisalahammad24.workers.dev/5c1d7b4c

@brianhogg could you take another look?

PR #3338, opened 2026-08-26
Add the user to the order status change note

### LINKED ISSUE #3039: Add the user who makes an order or student ernollment change to the log
When there are multiple admins, it's not possible to know who made an order or student enrollment status change when on the Order edit screen.

Add a note of the user (username and ID, in case they change their name) who made the change in the order log.

robindevitt:
## Description
Add the user to the order status change note.

Fixes #3039 

## How has this been tested?
Manually and reviewed via the order edit screen.

## Screenshots
Screenshot 2026-08-26 at 10 13 07

## Types of changes
The function `add_note` has a second parameter `added_by_user` which by default is `false`. So this change is setting the `added_by_user` parameter to true so the user details are shown in the notes.

By default it renders the users `display_name`. Referenced the screenshot for these below notes.
**Note 1**: Currently the note author was set to show as `LifterLMS`
**Note 2**: Showing the status update made by me with my display name being shown.
**Note 3**: I changed my display name and then changed the order status, to confirm it shows my updated display name.

## Checklist:

### DIFF
--- a/.changelogs/issue_3039.yml
+++ b/.changelogs/issue_3039.yml
@@ -1,0 +1,8 @@
+significance: patch
+type: changed
+comment: Add the user to the status change note.
+links:
+  - "#3039"
+attributions:
+  - "@robindevitt"
+entry: Add the user to the status change note.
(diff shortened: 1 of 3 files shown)


brianhogg:
@robindevitt Thanks! This sorts out the part about recording what user made a status change, but to finish the feature request we can add the user ID (if there is one) to the order notes output as well (ie. "by Jane Admin (#5)"). The note showing who did a student enrollment status change is already in there. If you're able to make that addition great, otherwise we can merge this in and I can do a follow up!

robindevitt:
@brianhogg thank you.

I have pushed an update now that include the information to the order note test. I'm using the details from the note itself rather that writing that additional information to the note text.

Note where the user_id are "0" which essentially will be "Lifter LMS" I have not appended the extra text for these to the notes. Most of these comments are likely to be automated and not made by a user.

Screenshot 2026-09-04 at 09 23 49

